PR #39, opened 2022-11-18
Adds Serial panel

### LINKED ISSUE #31: Add panel for configuring serial ports
Depends on #17.

The ports should be configurable via the GUI, as is done in the old software.

dalonsoa:
Only the widgets with some placeholder callbacks are included. No tests are added, as there's no point on testing the callbacks. These won't be implemented until there's some mock devices to work with.

Close #31

### DIFF
--- a/FINESSE_GUI/serial/serial_view.py
+++ b/FINESSE_GUI/serial/serial_view.py
@@ -13,7 +13,7 @@ class SerialPortControl(QWidget):
         self,
         devices: Dict[str, Dict[str, str]],
         avail_ports: Sequence[str],
-        avail_baud_rates: Sequence[int],
+        avail_baud_rates: Sequence[str],
     ) -> None:
         super().__init__()
 
@@ -23,9 +23,9 @@ class SerialPortControl(QWidget):
         self.setLayout(layout)
 
     def _create_controls(
-        self, avail_ports: Sequence[str], avail_baud_rates: Sequence[int]
+        self, avail_ports: Sequence[str], avail_baud_rates: Sequence[str]
     ) -> QGridLayout:
-        """Creates the controls for the devices.
+        """Creates the controls for the ports of the devices.
 
         Args:
             avail_ports: List of available ports.
@@ -65,7 +65,7 @@ class SerialPortControl(QWidget):
                     self.on_open_close_clicked, device_name=name, button=_open_close_btn
                 )
             )
-            _open_close_btn.setChecked(self._is_open[name])
+            _open_close_btn.setChecked(False)
             layout.addWidget(_open_close_btn, i, 3)
 
         return layout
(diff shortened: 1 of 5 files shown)


FINESSE_GUI/serial/serial_view.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+        self._devices[device_name]["port"] = new_baud_rate
+
+    def on_open_close_clicked(self, device_name: str, button: QPushButton) -> None:
+        """Open/Close the connection of the chosen device when the button is pushed.
+
+        This method tries to open/close the device and, if successful, change the label
+        on the button. Otherwise, it reverts back to the previous state.
+
+        TODO: Do we need this? Not doing anything for now, obviously.
```
alexdewar, line 105: I think this is ok as a placeholder for the time being.

pyproject.toml
```diff
@@ -8,7 +8,8 @@ authors = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.9"
+python = ">=3.9,<3.12"
+PySide6 = "^6.4.0"
```
alexdewar, line 12: FYI When I ran `poetry install` it complained that the lock file was out of sync, so I think you may need to run `poetry lock --no-update`

alexdewar (changes requested):
I spotted one typo and it seems that the `poetry` lock file needs regenerating.

But other than that it looks nice!

FINESSE_GUI/serial/serial_view.py
```diff
@@ -0,0 +1,145 @@
...
+
+class SerialPortControl(QWidget):
+    """Widgets to control the communication with a single serial port."""
+
+    def __init__(
+        self,
+        devices: Dict[str, Dict[str, str]],
+        avail_ports: Sequence[str],
+        avail_baud_rates: Sequence[str],
+    ) -> None:
```
alexdewar, line 17: Would you mind adding a docstring here?

I'd like to enable warnings for missing docstrings (#42) so that we can have total coverage before tackling the MkDocs stuff.

alexdewar (changes requested):

FINESSE_GUI/serial/serial_view.py
```diff
@@ -0,0 +1,145 @@
...
+            _brate.currentTextChanged.connect(  # type: ignore
+                partial(self.on_baud_rate_changed, device_name=name)
+            )
+            layout.addWidget(_brate, i, 2)
+
+            # And a manual way to open/close it
+            _open_close_btn = QPushButton("Open")
+            _open_close_btn.setCheckable(True)
+            _open_close_btn.released.connect(  # type: ignore
+                partial(
```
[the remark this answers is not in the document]
dalonsoa: It's the neat verison of lambda functions.

alexdewar (approved):
LGTM!

PR #50, opened 2022-11-22
Add main window

alexdewar:
This adds a main window to FINESSE, which at the moment only contains the `SerialPortControl`. It also adds the main entry point to the program so users can run `python -m finesse`.

This depends on #49 so we should merge that first, but I thought I'd create a draft PR for this now as I think this is a more sensible base for any GUI-related work @dc2917 does rather than the current `main` branch.


dalonsoa (approved):
Looks great!

Issue #12, opened 2022-10-25
Set up continuous deployment to auto-generate Windows binaries


alexdewar:
While we will make this project cross platform as far as possible, the target platform is Windows and so it would be useful to automatically generate Windows binaries. Apart from anything, it may be that the deployment machine won't have suitable development tools installed, so being able to just download the binaries would be a big help.

Previously I have used `pyinstaller` with some success.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-14

PR #51, opened 2022-11-22
Make FINESSE buildable with `pyinstaller`

alexdewar:
This allows for building FINESSE by running `pyinstaller FINESSE.spec` (the output is put in `dist/`). Tested on Windows and Linux and it runs without a hitch!

I decided to enable the option to build everything into a single (compressed) executable, so that when we get to publishing these files on GitHub somewhere (#12), they're easier to work with (i.e. you don't have to extract a zip file anything).

I ended up having to use a stub entry point again :cry:. @dalonsoa Did you figure that out with iPANACEA at some point?


dalonsoa:
That was solved in this PR, but it was never merged... https://github.com/ImperialCollegeLondon/iPANACEA

In summary, you can directly used `__main__.py`

dalonsoa (changes requested):
Use `__main__.py`, not a stub.

alexdewar:
I tried using `finesse/__main__.py` as an entry point instead of the stub, but I get the following error:

```
ImportError: attempted relative import with no known parent package
```

I think the problem is that in `finesse/__main__.py` there are some relative imports (cf. iPANACEA's `__main__.py`), so Python is throwing a wobbly:

```py
from . import config
from .gui.main_window import MainWindow
```

Any ideas?

stub.py
```diff
@@ -0,0 +1,5 @@
+"""Stub script for pyinstaller."""
+from finesse.__main__ import main
```
dalonsoa, line 2: I think that the options we have are either use absolute imports in `__main__.py` or, indeed, use a stub. Nothing against it.

I would re-organise a bit some things, though, as importing atom `__main__` just does not look correct:

1. Move the function that configures the logger to a `logger.py` file
2. Move the `main` function to `__init__.py`, importing there from logger, main window, etc.
3. Make `__main__.py` and `stub.py` identical, just one doing `from . import main` and the other `from finesse import main`.

Does this make sense?

dalonsoa (changes requested):

stub.py
alexdewar: Yup, sounds good!

dalonsoa (approved):
Keeping aside an issue with `poetry.lock` (as usual), all looks good!

PR #53, opened 2022-11-22
Add panel for controlling stepper motor

### LINKED ISSUE #32: Add panel for stepper motor controller
The panel will both show the status of the stepper motor controller as well as allowing the user to set the position of the view mirror attached to the stepper motor.

See diagram in #24.

From the proposal:

> Upper left red box: Shows the status of the stepper motor controller,
the boxes marked zenith, nadir, home... allow the user to point and click the required position of the view mirror attached to the stepper motor. The blue highlight (in this case over cold_BB) indicates the current position of the mirror.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-21

alexdewar:
Closes #32.

This PR adds the panel for controlling the stepper motor (which is in the top left of the GUI for the old program). Users can click on one of the buttons to move to a preset angle specified by the name on the button or can go to a specific angle/step by choosing a number and clicking "GOTO". The background of the last-clicked button changes to blue, as was done in the old program.

In the process, I created a `DummyStepperMotor` class to support the functionality of the GUI. I know we said we wouldn't venture into making dummy devices in this sprint, but this seemed like the cleanest way to implement the GUI functionality. (Its implementation will no doubt change when we realise the actual stepper motor behaves a bit differently.) I put it in a separate `hardware` module and it communicates with the GUI using PyPubSub messages. This should keep the concerns neatly separated.

### DIFF
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -11,11 +11,11 @@ class StepperMotorControl(QGroupBox):
         super().__init__("Target control")
 
         layout = QGridLayout()
-        zenith = QPushButton("ZENITH")
-        nadir = QPushButton("NADIR")
-        hot_bb = QPushButton("HOT_BB")
-        cold_bb = QPushButton("COLD_BB")
-        home = QPushButton("HOME")
+        zenith = self._create_stepper_button("ZENITH")
+        nadir = self._create_stepper_button("NADIR")
+        hot_bb = self._create_stepper_button("HOT_BB")
+        cold_bb = self._create_stepper_button("COLD_BB")
+        home = self._create_stepper_button("HOME")
         park = QPushButton("PARK")
         self.angle = QSpinBox()
         self.angle.setMaximum(359)
@@ -33,6 +33,14 @@ class StepperMotorControl(QGroupBox):
 
         self.setLayout(layout)
 
+    def _create_stepper_button(self, name: str) -> QPushButton:
+        btn = QPushButton(name)
+        btn.clicked.connect(lambda: self._preset_clicked(btn))  # type: ignore
+        return btn
+
+    def _preset_clicked(self, btn: QPushButton) -> None:
+        pub.sendMessage("stepper.move", step=btn.text().lower())
+
     def _goto_clicked(self) -> None:
         """Move stepper motor to specified position."""
-        pub.sendMessage("stepper.move", step_number=self.angle.value())
+        pub.sendMessage("stepper.move", step=self.angle.value())
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -36,21 +36,32 @@ class StepperMotorControl(QGroupBox):
         self.setLayout(layout)
 
     def _create_stepper_button(self, name: str) -> QPushButton:
+        """Create a button to move the motor to a preset position.
+
+        Args:
+            name: The name of the preset (will be converted to lowercase)
+        """
         btn = QPushButton(name)
         btn.clicked.connect(lambda: self._preset_clicked(btn))  # type: ignore
         return btn
 
-    def _set_last_clicked(self, btn: QPushButton) -> None:
+    def _set_selected_button(self, btn: QPushButton) -> None:
+        """Change the currently selected button, indicated by colour.
+
+        Args:
+            btn: The just-clicked button
+        """
         if self.last_clicked:
             self.last_clicked.setStyleSheet("")
         btn.setStyleSheet("background-color: blue")
         self.last_clicked = btn
 
     def _preset_clicked(self, btn: QPushButton) -> None:
-        self._set_last_clicked(btn)
+        """Move the stepper motor to preset position."""
+        self._set_selected_button(btn)
         pub.sendMessage("stepper.move", step=btn.text().lower())
 
     def _goto_clicked(self) -> None:
         """Move stepper motor to specified position."""
-        self._set_last_clicked(self.goto)
+        self._set_selected_button(self.goto)
         pub.sendMessage("stepper.move", step=self.angle.value())
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -1,4 +1,5 @@
 """Code for controlling the stepper motor which moves the mirror."""
+from pubsub import pub
 from PySide6.QtWidgets import QGridLayout, QGroupBox, QPushButton, QSpinBox
 
 
@@ -16,9 +17,10 @@ class StepperMotorControl(QGroupBox):
         cold_bb = QPushButton("COLD_BB")
         home = QPushButton("HOME")
         park = QPushButton("PARK")
-        angle = QSpinBox()
-        angle.setMaximum(359)
+        self.angle = QSpinBox()
+        self.angle.setMaximum(359)
         goto = QPushButton("GOTO")
+        goto.clicked.connect(self._goto_clicked)  # type: ignore
 
         layout.addWidget(zenith, 0, 0)
         layout.addWidget(nadir, 0, 1)
@@ -26,7 +28,11 @@ class StepperMotorControl(QGroupBox):
         layout.addWidget(cold_bb, 0, 3)
         layout.addWidget(home, 1, 0)
         layout.addWidget(park, 1, 1)
-        layout.addWidget(angle, 1, 2)
+        layout.addWidget(self.angle, 1, 2)
         layout.addWidget(goto, 1, 3)
 
         self.setLayout(layout)
+
+    def _goto_clicked(self) -> None:
+        """Move stepper motor to specified position."""
+        pub.sendMessage("stepper.move", step_number=self.angle.value())
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -16,7 +16,6 @@ class StepperMotorControl(QGroupBox):
         hot_bb = self._create_stepper_button("HOT_BB")
         cold_bb = self._create_stepper_button("COLD_BB")
         home = self._create_stepper_button("HOME")
-        park = QPushButton("PARK")
         self.angle = QSpinBox()
         self.angle.setMaximum(359)
         goto = QPushButton("GOTO")
@@ -27,7 +26,6 @@ class StepperMotorControl(QGroupBox):
         layout.addWidget(hot_bb, 0, 2)
         layout.addWidget(cold_bb, 0, 3)
         layout.addWidget(home, 1, 0)
-        layout.addWidget(park, 1, 1)
         layout.addWidget(self.angle, 1, 2)
         layout.addWidget(goto, 1, 3)
 
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -1,4 +1,6 @@
 """Code for controlling the stepper motor which moves the mirror."""
+from typing import Optional
+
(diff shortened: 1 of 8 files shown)


finesse/gui/stepper_motor_view.py
```diff
@@ -0,0 +1,67 @@
...
+        self.setLayout(layout)
+
+    def _create_stepper_button(self, name: str) -> QPushButton:
+        """Create a button to move the motor to a preset position.
+
+        Args:
+            name: The name of the preset (will be converted to lowercase)
+        """
+        btn = QPushButton(name)
+        btn.clicked.connect(lambda: self._preset_clicked(btn))  # type: ignore
```
dalonsoa, line 45: Use `partial` from `functools`, as I did. `mypy` seems to be happy with it.

finesse/gui/stepper_motor_view.py
```diff
@@ -0,0 +1,67 @@
...
+        """Change the currently selected button, indicated by colour.
+
+        Args:
+            btn: The just-clicked button
+        """
+        if self.last_clicked:
+            self.last_clicked.setStyleSheet("")
+        btn.setStyleSheet("background-color: blue")
+        self.last_clicked = btn
+
```
dalonsoa, line 58: I'm not totally sure, but the way you're using this makes me think on button groups. Have a look at this in case that is what you actually want to use: https://doc.qt.io/qtforpython/PySide6/QtWidgets/QButtonGroup.html

finesse/gui/stepper_motor_view.py
```diff
@@ -0,0 +1,67 @@
...
+    def _preset_clicked(self, btn: QPushButton) -> None:
+        """Move the stepper motor to preset position."""
+        self._set_selected_button(btn)
+        pub.sendMessage("stepper.move", target=btn.text().lower())
+
+    def _goto_clicked(self) -> None:
```
dalonsoa, line 64: GOTO makes me think on the old Fortran 77...

finesse/hardware/dummy_stepper_motor.py
```diff
@@ -0,0 +1,52 @@
...
+
+        Args:
+            target: The target angle (in degrees) or the name of a preset
+        """
+        if isinstance(target, str):
+            self.current_step = DummyStepperMotor.get_preset_step(target)
+        else:
+            step = round(self.steps_per_rotation * target / 360.0)
+            if step < 0 or step >= self.steps_per_rotation:
+                raise ValueError("step ")
```
dalonsoa, line 49: So we need something more informative here?

tests/test_stepper_motor.py
```diff
@@ -0,0 +1,49 @@
+"""Tests for DummyStepperMotor."""
+import pytest
+
+from finesse.hardware.dummy_stepper_motor import DummyStepperMotor
+
+PRESETS = ("zenith", "nadir", "hot_bb", "cold_bb", "home")
```
dalonsoa, line 6: You can import this from the stepper motor module.

tests/test_stepper_motor.py
```diff
@@ -0,0 +1,49 @@
...
+def test_constructor() -> None:
+    """Check that the constructor only accepts values in permissible range."""
+    with pytest.raises(ValueError):
+        DummyStepperMotor(0)
+
+    with pytest.raises(ValueError):
+        DummyStepperMotor(-1)
+
+    # Should work
+    DummyStepperMotor(len(PRESETS))
```
dalonsoa, line 18:
```suggestion
    for num in (-1, 0):
        with pytest.raises(ValueError):
            DummyStepperMotor(num)

    for num in (len(PRESETS), len(PRESETS)+10):
        DummyStepperMotor(num)
```

tests/test_stepper_motor.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_move_to_number() -> None:
```
dalonsoa, line 21: I think that these tests that check that some values work well but others are not valid call for `pytest.parametrize` or even `hypothesis` to be really thorough a check edge cases.

finesse/hardware/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""This module contains code for interfacing with different hardware devices."""
+from .dummy_stepper_motor import DummyStepperMotor
+
+# TODO: Replace with a real stepper motor device
+stepper = DummyStepperMotor(3600)
```
dalonsoa, line 5: I don't think we want to create a global stepper here. We need to think about this, but considering how many devices we have, I think we should consider to setup a sort of `DeviceManager` living at the same level as the MainWindow in charge of starting them, dealing with their Errors, etc. For now, it is ok, but something to bear in mind.

dalonsoa (changes requested):
There are a couple of suggestions both in the font and back end, but otherwise looks ok. I bet the real device will be more challenging...

finesse/gui/stepper_motor_view.py
alexdewar: Ahh that's neat! I managed to rewrite my "make selected buttons blue" logic using a `QButtonGroup`, which is much cleaner.

tests/test_stepper_motor.py
alexdewar: Good idea. I've moved it into `config.py` now so that we can also use it from the GUI side and not have to hardcode these values everywhere.

tests/test_stepper_motor.py
alexdewar: Ah I've never used either of those, but it seems a sensible approach. I've ported over the tests to use `pytest.parametrize` now.

finesse/hardware/__init__.py
alexdewar: Yep, I agree. Once the rest of the code takes shape it'll become a bit more obvious how to approach this.

alexdewar:
I think it's ready for re-review now for when you're back.

In addition to the changes you suggested:

- I also rebased the branch onto `main`
- I swapped the `QGridLayout` in `MainWindow` for two separate columns of widgets, as `StepperMotorControl` was getting vertically stretched

finesse/hardware/dummy_stepper_motor.py
alexdewar: Oops! Yes we do

alexdewar:
Ping @dalonsoa. Not urgent but if you have a chance to re-review this that'd be great.

It'd also be good to get your input @dc2917 while you're working on FINESSE.

dc2917 (approved):
Looks good to me, I can't see any issues here

finesse/gui/stepper_motor_view.py
```diff
@@ -0,0 +1,59 @@
...
+        layout.addWidget(self.angle, 1, 2)
+        layout.addWidget(self.goto, 1, 3)
+
+        self.setLayout(layout)
+
+    def _add_checkable_button(self, name: str) -> QPushButton:
+        """Add a selectable button to button_group."""
+        btn = QPushButton(name)
+        btn.setCheckable(True)
+        btn.setStyleSheet("QPushButton::checked{background: blue; color: white;}")
```
dalonsoa, line 50: Unless you fancy specifically this style, buttons in a group change colour when they are pushed - at least in Mac; I would expect the same in others. See video. So I think this line can be removed.


https://user-images.githubusercontent.com/6095790/205260718-a79fd9c1-008e-48f1-a0b0-7e7ace8a1f28.mov

finesse/hardware/dummy_stepper_motor.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        if isinstance(target, str):
+            self.current_step = DummyStepperMotor.get_preset_step(target)
+        else:
+            step = round(self.steps_per_rotation * target / 360.0)
+            if step < 0 or step >= self.steps_per_rotation:
+                raise ValueError("step number is out of range")
+            self.current_step = step
+
+        logging.info(f"Moving stepper motor to step {self.current_step}")
```
dalonsoa, line 52: I was very confused by the log results, when I was selecting 1, clicking GOTO and having 10 printed in the terminal. I think that in the GUI there should be a clear indication that the number you are putting there is the angle in degrees and having the log showing:

```output
Moving stepper motor to step 10 (= 1º)
```

Or something like that.

finesse/gui/stepper_motor_view.py
```diff
@@ -0,0 +1,59 @@
...
+        for i, preset in enumerate(ANGLE_PRESETS):
+            btn = self._add_checkable_button(preset.upper())
+            self.button_group.addButton(btn)
+
+            row, col = divmod(i, BUTTONS_PER_ROW)
+            layout.addWidget(btn, row, col)
+
+        # We also have a way for users to move the mirror to an angle of their choice
+        self.angle = QSpinBox()
+        self.angle.setMaximum(359)
```
dalonsoa, line 38: This makes the angular resolution equal to 1º. Is that good enough for the application?

tests/test_stepper_motor.py
```diff
@@ -0,0 +1,53 @@
...
+def test_constructor_good(steps: int) -> None:
+    """Check valid arguments to constructor."""
+    DummyStepperMotor(steps)
+
+
+@pytest.mark.parametrize("steps", (-1000000, -1000, *range(-1, len(ANGLE_PRESETS))))
+def test_constructor_bad(steps: int) -> None:
+    """Check invalid arguments to constructor."""
+    with pytest.raises(ValueError):
+        DummyStepperMotor(steps)
```
dalonsoa, line 20: The normal use when using parametric tests is that you provide as input to the test the input parameters and the expected outputs/outcomes. So, these two functions could be written in a more concise manner as (see [VirtualRainforest](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/89407edba9c707f33281dd9051fecbee7ec716dd/tests/test_models.py#L43) for a more complex example):

```python
from contextlib import nullcontext as does_not_raise


@pytest.mark.parametrize(
    "steps,raises",
    [
        [
            num,
            pytest.raises(ValueError) if num < len(ANGLE_PRESETS) else does_not_raise(),
        ]
        for num in range(-5, len(ANGLE_PRESETS) + 5)
    ],
)
def test_constructor(steps: int, raises: Any) -> None:
    """Check invalid arguments to constructor."""
    with raises:
        DummyStepperMotor(steps)
```

Another option is to put the logic to decide on the outcome inside the test:

```python
from contextlib import nullcontext as does_not_raise


@pytest.mark.parametrize(
    "steps",
    list(range(-5, len(ANGLE_PRESETS) + 5)),
)
def test_constructor(steps: int) -> None:
    """Check invalid arguments to constructor."""
    raises = (
        pytest.raises(ValueError) if steps < len(ANGLE_PRESETS) else does_not_raise()
    )
    with raises:
        DummyStepperMotor(steps)
```

Assuming `mypy` does not complain much, I think this approach is more concise than splitting into good and bad tests.

dalonsoa (approved):
I have a few suggestions, but otherwise it looks good, so I'll approve it.

finesse/gui/stepper_motor_view.py
alexdewar: Good point. I went with blue because that was the colour used in the old GUI, but I think the default style is probably fine.

For me, the buttons just appear as pressed in by default, not blue, but I guess that's fine too.

finesse/gui/stepper_motor_view.py
alexdewar: It should be. I think that's what the old program has.

tests/test_stepper_motor.py
alexdewar: Oh that's really cool. Thanks!

PR #54, opened 2022-11-23
OPUS view

### LINKED ISSUE #34: Add panel showing interferometer connection and scan status
The interferometer is connected over HTTP.

The panel should display connection status, including any errors that have occurred (see diagram in #24) as well as allowing the user to start and stop the device etc.

I *think* the current software has a button for launching an external program (Opus) which provides more detailed information from the interferometer.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-22

dalonsoa:
Implements the view to control and monitor the connection to the interferometer. It follows the same layout than the original application, which does not mean it is the best option.

At the moment, there's no much in terms of backend relationship and - potentially - some of this information (the connection URLs) should be moved to the backend model - when there's a backend module. But it should work, for now. 

I've modified the main window to accommodate this view.

image

Close #34

### DIFF
--- a/finesse/gui/main_window.py
+++ b/finesse/gui/main_window.py
@@ -1,6 +1,7 @@
 """Code for FINESSE's main GUI window."""
-from PySide6.QtWidgets import QMainWindow
+from PySide6.QtWidgets import QGridLayout, QGroupBox, QMainWindow, QWidget
 
+from .opus_view import OPUSControl
 from .serial_view import SerialPortControl
 
 
@@ -12,13 +13,22 @@ class MainWindow(QMainWindow):
         super().__init__()
         self.setWindowTitle("FINESSE")
 
+        layout = QGridLayout()
+
         devices = {
             "ST10": {"port": "COM5", "baud_rate": "9600"},
             "DP9800": {"port": "COM1", "baud_rate": "9600"},
         }
-        serial_port = SerialPortControl(
+        serial_port: QGroupBox = SerialPortControl(
             devices,
             ("COM1", "COM5", "COM7"),
             ("600", "9600", "115200"),
         )
-        self.setCentralWidget(serial_port)
+        opus: QGroupBox = OPUSControl()
+
+        layout.addWidget(serial_port, 3, 0)
+        layout.addWidget(opus, 0, 1)
+
+        central = QWidget()
+        central.setLayout(layout)
+        self.setCentralWidget(central)
--- a/finesse/gui/main_window.py
+++ b/finesse/gui/main_window.py
@@ -24,7 +24,7 @@ class MainWindow(QMainWindow):
             ("COM1", "COM5", "COM7"),
             ("600", "9600", "115200"),
         )
-        opus: QGroupBox = OPUSControl()
+        opus: QGroupBox = OPUSControl("127.0.0.1")
 
         layout.addWidget(serial_port, 3, 0)
         layout.addWidget(opus, 0, 1)
--- a/finesse/gui/opus_view.py
+++ b/finesse/gui/opus_view.py
@@ -2,6 +2,8 @@
 import logging
 from functools import partial
 
+from PySide6.QtCore import QUrl
+from PySide6.QtWebEngineWidgets import QWebEngineView
 from PySide6.QtWidgets import QGroupBox, QHBoxLayout, QPushButton, QVBoxLayout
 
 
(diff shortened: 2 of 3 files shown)


dalonsoa:
I'll add tests for those things that are not placeholders after a first review, @alexdewar

finesse/gui/opus_view.py
```diff
@@ -0,0 +1,226 @@
...
+
+        logging.getLogger("OPUS").addHandler(ch)
+
+    def __init__(self, log_area: weakref.ref):
+        """Constructor of the Handler.
+
+        Args:
+            log_area: A weak reference to the log area.
+        """
+        super(OPUSLogHandler, self).__init__()
```
alexdewar, line 197: With Python 3 you can just do:

```suggestion
        super().__init__()
```

finesse/gui/opus_view.py
```diff
@@ -0,0 +1,226 @@
...
+        If the log area has been destroyed before the logger, it will raise an
+        AttributeError. This can only happen during tests and can be safely ignored.
+        """
+        try:
+            self.log_area().append(self.format(record))
+            cursor = self.log_area().textCursor()
+            cursor.movePosition(QTextCursor.End)
+            self.log_area().setTextCursor(cursor)
+        except AttributeError:
+            pass
```
alexdewar, line 212: I'd just convert `log_area` to a strong reference once at the start of the function, rather than doing it every time you want it:

```suggestion
        log_area = self.log_area()
        if not log_area:
            # log_area no longer exists
            return

        log_area.append(self.format(record))
        cursor = log_area.textCursor()
        cursor.movePosition(QTextCursor.End)
        log_area.setTextCursor(cursor)
```

alexdewar (changes requested):
This is cool! I like that there's placeholder code in there for the HTTP protocol business. I agree that we probably don't need the web view in the long term but we can deal with that another time.

I've got some minor suggestions, but otherwise all good!

finesse/gui/opus_view.py
```diff
@@ -0,0 +1,226 @@
...
+    "cancel": "opusrs/cmd.htm?opusrscancel",
+    "opus": "TBC",
+}
+
+
+class OPUSControl(QGroupBox):
+    """Class that monitors and control the OPUS interferometer."""
+
+    def __init__(self, ip: str, commands: Optional[Dict[str, str]] = None) -> None:
+        """Creates the widgets to monitor and control the OPUS interferometer.
```
alexdewar, line 32:
```suggestion
        """Create the widgets to monitor and control the OPUS interferometer.
```

finesse/gui/main_window.py
```diff
@@ -12,13 +13,22 @@ def __init__(self) -> None:
...
         self.setWindowTitle("FINESSE")
 
+        layout = QGridLayout()
+
         devices = {
             "ST10": {"port": "COM5", "baud_rate": "9600"},
             "DP9800": {"port": "COM1", "baud_rate": "9600"},
         }
-        serial_port = SerialPortControl(
+        serial_port: QGroupBox = SerialPortControl(
```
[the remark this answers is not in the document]
dalonsoa: Yep. For some reason `mypy` was not happy at all :(

finesse/gui/opus_view.py
```diff
@@ -0,0 +1,226 @@
...
+        logging.info("Getting OPUS status!")
+        self.status.load(QUrl(self.url("status")))
+        self.status.show()
+
+        logging.getLogger("OPUS").error("Oh, no! Something bad happened!")
+
+    def open_opus(self) -> None:
+        """Opens OPUS front end somewhere else.
+
+        TODO: No idea what this is supposed to do.
```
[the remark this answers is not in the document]
dalonsoa: Well, despite the `open` verb there, this is link to the `OPUS` button. What is that button supposed to do?

finesse/gui/opus_view.py
dalonsoa: I'm not sure if this will work as you think because what happens is - I think - the python wrapper to the underlying C++ object still exists (so self.log_area is not None) but the actual C++ object is done. If it works, you option is way cleaner. Let's use it and see how it goes.

finesse/gui/opus_view.py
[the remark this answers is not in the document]
alexdewar: Ohhh I see. That just opens the manufacturer's program (called OPUS) which shows spectrography info etc. I suspect it is using the same HTTP-based protocol that we will be, so it probably makes sense to reimplement at least a portion of what it does ourselves, but I guess for now we should just keep the old behaviour.

dalonsoa:
Tha's what happens when you edit things directly in GitHub...

PR #60, opened 2022-11-30
Add code for handling uncaught exceptions

### LINKED ISSUE #55: Add default handler for uncaught exceptions
When uncaught exceptions occur, we really want to:

- Notify the user with a dialog box
- Write relevant info to the program log

The info should include a full stack trace.

alexdewar:
This should make it easier for end-users to communicate with us about bugs. Closes #55.

The code is based on something similar I did for iPANACEA. It could potentially use an error icon, but I don't think that's a priority.

On my machine it looks like this:
image


dalonsoa (approved):
This looks great!

PR #61, opened 2022-12-01
Auto-deploy compiled Windows binaries to GitHub

### LINKED ISSUE #12: Set up continuous deployment to auto-generate Windows binaries

alexdewar:
This is based on @dalonsoa's work for iPANACEA.

At the moment, I've set the CI to deploy a Windows binary on every push, which allows us to test code on PRs before merging them, but in future we may want to restrict this job to just run on `main` or for new releases.

Closes #12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,15 @@ jobs:
       run: |
         poetry run mkdocs build -s
 
+    - name: Build package
+      if: matrix.os == 'windows-latest'
+      run: poetry run pyinstaller FINESSE.spec
+
+    - uses: actions/upload-artifact@v2
+      if: success() && matrix.os == 'windows-latest'
+      with:
+        name: FINESSE
+        path: dist/FINESSE.exe
 
   docs:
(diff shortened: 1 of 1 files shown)


alexdewar:
I really wanted to make a file `FINESSE.exe` that's directly downloadable as opposed to it being bundled in a zip file, but it seems that this is currently not possible: https://github.com/actions/upload-artifact/issues/14

It's a pity, because the executable is already compressed, so there's not really any benefit to zipping it. Oh well.

dalonsoa (approved):
It looks good to me. I would not worry with being Zipped - it is really a very minor issue that do not depend on us, anyway.

alexdewar:
Cool, thanks!

PR #62, opened 2022-12-01
Add panel for temperature control and monitoring

### LINKED ISSUE #35: Add panel for temperature control and monitoring
This should contain controls for setting the temperature to a particular value (#18).

Eventually, this should also include a plot of the current temperature (#20).

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-24

dc2917:
Closes #35. 

This PR adds the panel for controlling and monitoring the temperatures of the "hot" and "cold" blackbodies (which is in the middle-right of the GUI for the old program). Temperatures are displayed on a graph which can be updated as new measurements are obtained. Widget functionality is yet to be implemented.

Note that there seems to be a requirement of the ordering of imports, with PySide6 needing to be imported before setting the matplotlib backend, which `isort` overwrites - thoughts/suggestions on this would be appreciated.


finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+    def __init__(self) -> None:
+        """Creates the widgets to control and monitor TC4820 HOT.
+
+        Returns:
+            None
+        """
+        super().__init__("TC4820 HOT")
+
+        layout = self._create_controls()
+        self.setLayout(layout)
```
dalonsoa, line 196: As you mention, I think this redundant. I'll remove the cold version and modify this one as:

```suggestion
class TC4820(QGroupBox):
    """Widgets to view the TC4820 properties."""

    def __init__(self, location: str) -> None:
        """Creates the widgets to control and monitor TC4820.

        Args:
            location: The location this controller is monitoring/controlling, typically "hot" or "cold".

        Returns:
            None
        """
        super().__init__(f"TC4820 {location.upper()}")

        self.location = location
        
        layout = self._create_controls()
        self.setLayout(layout)
```

Then, where relevant, you will use this as `TC4820("hot")` or `TC4820("cold")`.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+        self._set_sbox = QSpinBox()
+        layout.addWidget(self._set_sbox, 2, 1)
+
+        self._update_pbtn = QPushButton("UPDATE")
+        layout.addWidget(self._update_pbtn, 2, 3)
+
+        return layout
+
+
+class TC4820_COLD(QGroupBox):
```
dalonsoa, line 257: Remove

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+
+        self._power_slider = QSlider()
+        self._power_slider.setOrientation(Qt.Orientation.Horizontal)
+        layout.addWidget(self._power_slider, 1, 1, 1, 3)
+        layout.addWidget(QLineEdit("40"), 1, 4)
+
+        self._poll_light = QLabel()
+        self._poll_light.setPixmap(QPixmap("./finesse/gui/images/poll_off.png"))
+        self._alarm_light = QLabel()
+        self._alarm_light.setPixmap(QPixmap("./finesse/gui/images/alarm_on.png"))
```
dalonsoa, line 244: Just use `off` as default.
```suggestion
        self._alarm_light.setPixmap(QPixmap("./finesse/gui/images/alarm_off.png"))
```

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+    app = QApplication(sys.argv)
+
+    window = QMainWindow()
+
+    layout = QGridLayout()
+
+    bb_monitor = BBMonitor()
+    dp9800 = DP9800()
+    tc4820_hot = TC4820_HOT()
+    tc4820_cold = TC4820_COLD()
```
dalonsoa, line 347:
```suggestion
    tc4820_hot = TC4820("hot")
    tc4820_cold = TC4820("cold")
```

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+
+        layout = self._create_controls()
+        self.setLayout(layout)
+
+    def _create_controls(self) -> QGridLayout:
+
+        layout = QGridLayout()
+
+        wid = QLabel("CONTROL")
+        wid.setAlignment(Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter)
```
dalonsoa, line 203: To keep it short, maybe create a shortcut for this:
```suggestion
        alignment = Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter
        
        wid = QLabel("CONTROL")
        wid.setAlignment(alignment)
```
and replace it everywhere else.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+
+        return layout
+
+    def _create_figure(self) -> None:
+        """Creates the matplotlib figure to be contained within the panel.
+
+        Returns:
+            None
+        """
+        self._figure, self._ax1 = plt.subplots()
```
dalonsoa, line 56: I would keep the axes in a dictionary, so they can be updated based on a signal coming from somewhere in an easier way:
```suggestion
        self._figure, ax = plt.subplots()
        self.ax = {"hot": ax}
        ...
        self.ax["cold"] = self.ax["hot"].twinx()
```

And a similar thing with the lines, putting them into `self.lines["hot"]` and `self.lines["cold"]`. I think it is neater in this case because then you can use loops to repeat the same actions over the two sets of data.

dalonsoa (changes requested):
I've made a few comments, but it looks good overall. The layout will need to be updated as we add more stuff - and because I think @alexdewar has changed the layout of the main window - and obviously we need to add some callbacks to respond to events, but this will be hard to do until we start coding the backend, so all ok on that front, for now.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+    QWidget,
+)
+
+# plt.style.use("./finesse_gui.style")
+plt.rcParams["figure.facecolor"] = "black"
+plt.rcParams["axes.facecolor"] = "black"
+plt.rcParams["axes.edgecolor"] = "white"
+plt.rcParams["text.color"] = "white"
+plt.rcParams["xtick.color"] = "white"
+plt.rcParams["ytick.color"] = "white"
```
alexdewar, line 24: I'm not sure about setting the plot parameters globally like this. It's fine for now, but it might pose a problem later down the line if we want to produce another plot somewhere else in the code that looks different.

I guess you're trying to follow the colour scheme from the screen shot of the old program here? That's good, but tbh I think the `matplotlib` defaults are fine here, so I'd just drop this chunk. Personally I'm not a fan of black backgrounds for plots.

If @jonemurray really wants to keep this colour scheme though (or do something else...) we can always accommodate that.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+
+        layout.addWidget(self._canvas, 0, 0)
+
+        return layout
+
+    def _create_figure(self) -> None:
+        """Creates the matplotlib figure to be contained within the panel.
+
+        Returns:
+            None
```
alexdewar, line 54: You don't need to specify this if the function's just returning `None`.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+        """
+        self._figure, self._ax1 = plt.subplots()
+        self._canvas = FigureCanvasQTAgg(self._figure)
+
+        t = [1040, 1050, 1060, 1070, 1080, 1090, 1100]
+        hot_bb_temp = [55, 57.5, 60, 62.5, 65, 67.5, 70]
+        cold_bb_temp = [1, 1.1, 1.15, 1.2, 1.3, 1.4, 1.5]
```
alexdewar, line 61: This is fine for now, as it's just as a placeholder, but usually I'd make things like this arguments (as @dalonsoa did with `SerialPortControl` for example).

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+
+        self._bb_hot_line = self._ax1.plot(t, hot_bb_temp, color=[0, 1, 0])
+        self._ax1.set_xlabel("")
```
alexdewar, line 64: Do you need to explicitly set the x label to nothing here?

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+        self._ax1.set_ylabel("HOT BB", color=[0, 1, 0])
+        self._ax1.set_xlim([1045, 1101.4])
+        self._ax1.set_ylim([20, 80])
+
+        self._ax2 = self._ax1.twinx()
+        self._bb_cold_line = self._ax2.plot(t, cold_bb_temp, color=[1, 1, 0])
+        self._ax2.set_ylabel("COLD BB", color=[1, 1, 0])
+        self._ax2.set_ylim([0, 10])
```
alexdewar, line 72: As with the comment above, we won't want the axis limits and labels to be hard-coded forever, but it's fine to leave them as is for now.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+        self._ch8 = QLineEdit()
+        self._ch8.setReadOnly(True)
+        layout.addWidget(self._ch8, 1, 8)
+
+        wid = QLabel("POLL")
+        wid.setAlignment(Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter)
+        layout.addWidget(wid, 0, 9, 2, 1)
+
+        self._poll_light = QLabel()
+        self._poll_light.setPixmap(QPixmap("./finesse/gui/images/poll_off.png"))
```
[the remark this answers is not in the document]
alexdewar: You also *might* want to consider creating all these `QPixmaps` in one place and making them e.g. global variables, so you don't have to keep loading the files. It's not important though, so don't worry if it's a faff.

alexdewar (changes requested):
The code looks good for the most part and it looks nice rendered in the GUI too. Good work!

I have made some minor suggestions here and there.

jonemurray:
Hi All,

Lots of code! Sorry I’ve not been involved, I’ve been up to my eyes with other work and we have just been collared by ESA to take FINESSE out to Andoya, Norway this coming February so more work now required to plan the logistics around that.

It would be good to see how the GUI looks, if you have time next week.

Thanks, Jon.

alexdewar:
Hi Jon,

(I've taken the liberty of trimming the message history off your comment above. For some reason GitHub's email handler doesn't seem clever enough to do it itself.)

No that's ok @jonemurray. No pressure!

I'd love to show you what we've done so far and get your feedback. Daniel (@dc2917) and I were hoping to come see you next week, if we can find a time that works for everyone.

We've basically got a rough initial draft of the GUI finished, so we may be in a position to start tinkering with actual hardware at some point next week, so if we could also arrange for a longer stint in the lab too, that'd be ace. We don't need the whole rig up and running to work on things though, so if there's a way we can try out the sensors without you having to go to the trouble of handling liquid nitrogen etc., that's totally fine.

Unfortunately I'm away on Tues and Weds of next week, but I'm keen to make progress with FINESSE, so I can make myself available on the other days. I'll be in on Monday anyway as it's our anchor day and free after lunch.

finesse/gui/temp_control.py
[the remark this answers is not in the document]
dc2917: Ah yes, I meant to ask about the best way to set the image file path in the PR actually - thanks!

alexdewar:
@dc2917 Something else I forgot to mention...

This isn't a biggie, but your commit messages seem to be condensed all into a single line. I'm guessing you're committing with `git commit -m`? For longer messages, it's better to not use the `-m` flag and to write it in an editor.

Normally with git commits, you want the first line to be a brief summary of what you've done (some people suggest <= 50 characters) and then you write a longer message a couple of lines below, e.g.:

```
Add more tests

Currently we have no tests, which is bad. Add tests for X, Y and Z... (etc. etc.)
```

Command-line git as well as most tools commonly use just the first line when they're giving a condensed view of the list of changes.

finesse/gui/main_window.py
```diff
@@ -26,8 +27,17 @@ def __init__(self) -> None:
         )
         opus: QGroupBox = OPUSControl("127.0.0.1")
 
+        bb_monitor: QGroupBox = BBMonitor()
+        dp9800: QGroupBox = DP9800()
+        tc4820_hot: QGroupBox = TC4820_HOT()
+        tc4820_cold: QGroupBox = TC4820_COLD()
+
         layout.addWidget(serial_port, 3, 0)
         layout.addWidget(opus, 0, 1)
```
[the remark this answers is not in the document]
alexdewar: Actually, I think once you've sorted the merge conflicts this might not matter anymore.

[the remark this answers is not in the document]
dc2917: Cool, thanks for the heads up. But yeah, I was mostly avoiding touching the positioning of the other components until we have them all added.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+        wid = QLabel("ALARM")
+        wid.setAlignment(Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter)
+        layout.addWidget(wid, 2, 4)
+
+        self._control_val = QLineEdit("70.5")
+        self._control_val.setReadOnly(True)
+        self._control_val.setAlignment(Qt.AlignmentFlag.AlignCenter)
+        layout.addWidget(self._control_val, 0, 1)
+
+        self._pt100_val = QLineEdit("70.34")  # CH_7?
```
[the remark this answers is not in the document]
dc2917: Ah, this was just because I noticed that the value here corresponded to the value in the CH_7 box, so wondered whether the two components need to be connected.

alexdewar:
@dc2917 Let me know if/when this is ready for re-review or if there's anything I can do to help :-)

finesse/gui/temp_control.py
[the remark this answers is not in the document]
dc2917: I had a go at doing this with `importlib` (which is apparently intended as a replacement for `pkg_resources`) but it was failing the GitHub `pytest` (although not my local `pytest`) so I've removed it for the time being until I can work out how to do it properly. I do therefore still have the hard-coded paths in there, but can try and sort that before merging if you'd prefer.

finesse/gui/temp_control.py
```diff
@@ -41,138 +34,147 @@ def _create_controls(self) -> QGridLayout:
...
+        colours = plt.rcParams["axes.prop_cycle"]
+        hot_colour = colours.by_key()["color"][0]
+        cold_colour = colours.by_key()["color"][1]
 
-        self._bb_hot_line = self._ax1.plot(t, hot_bb_temp, color=[0, 1, 0])
-        self._ax1.set_xlabel("")
-        self._ax1.set_ylabel("HOT BB", color=[0, 1, 0])
-        self._ax1.set_xlim([1045, 1101.4])
-        self._ax1.set_ylim([20, 80])
+        self._ax["hot"].plot(t, hot_bb_temp, color=hot_colour)
```
alexdewar, line 70: I'm not sure why you're plotting `hot_bb_temp` and `cold_bb_temp` when they've been set to all `None`s. Surely that means nothing will be plotted...? If so, you can just drop this call to `plot()` and not bother setting `hot_bb_temp` and `cold_bb_temp` above.

finesse/gui/temp_control.py
```diff
@@ -41,138 +34,147 @@ def _create_controls(self) -> QGridLayout:
...
+        self._ax["cold"] = self._ax["hot"].twinx()
+        self._ax["cold"].plot(t, cold_bb_temp, color=cold_colour)
+        self._ax["cold"].set_ylabel("COLD BB", color=cold_colour)
+        self._ax["cold"].set_ylim([0, 10])
 
         self._canvas.draw()
 
-    def _update_figure(self) -> None:  # , x, y) -> None:
-        """Updates the matplotlib figure to be contained within the panel.
+    def _btn_pressed(self, name) -> None:
```
alexdewar, line 82: This needs a type annotation:

```suggestion
    def _btn_pressed(self, name: str) -> None:
```

Not sure why `mypy` didn't flag it!

alexdewar: It's also good practice to name functions according to what they *do*, so in this case it could be `_toggle_axis_visibility` or something. (I'm actually not particularly consistent with this myself tbh :laughing:)

finesse/gui/temp_control.py
alexdewar: Ah, now I see you're actually for these `None`s in `_update_figure` below. The way I'd do this is to just plot zero points then in `_update_figure` you can just check this with `if xdata`.
 
```suggestion
        self._ax["hot"].plot(t, (), color=hot_colour)
```

finesse/gui/temp_control.py
```diff
@@ -41,138 +34,147 @@ def _create_controls(self) -> QGridLayout:
...
+        cold_colour = colours.by_key()["color"][1]
 
-        self._bb_hot_line = self._ax1.plot(t, hot_bb_temp, color=[0, 1, 0])
-        self._ax1.set_xlabel("")
-        self._ax1.set_ylabel("HOT BB", color=[0, 1, 0])
-        self._ax1.set_xlim([1045, 1101.4])
-        self._ax1.set_ylim([20, 80])
+        self._ax["hot"].plot(t, hot_bb_temp, color=hot_colour)
+        self._ax["hot"].set_ylabel("HOT BB", color=hot_colour)
+        self._ax["hot"].set_xlim([1045, 1101.4])
```
alexdewar, line 72: These are very specific values and obvs going forward we'll want the x limits to be set dynamically in some way.

finesse/gui/temp_control.py
```diff
@@ -41,138 +34,147 @@ def _create_controls(self) -> QGridLayout:
...
         xdata.pop(0)
         y1data.pop(0)
         y2data.pop(0)
 
-        x = xdata[-1] + 10
-
-        # Basic RNG for testing
-        y1 = (60 * y1data[-1] + 50) % 70
-        y2 = (5 * y2data[-1] + 1) % 8
+        if xdata[-1] is not None:
```
alexdewar, line 100: See comment above

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,359 @@
...
+
+        x = xdata[-1] + 10
+
+        # Basic RNG for testing
+        y1 = (60 * y1data[-1] + 50) % 70
+        y2 = (5 * y2data[-1] + 1) % 8
+
+        xdata.append(x)
+        y1data.append(y1)
+        y2data.append(y2)
```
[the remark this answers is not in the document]
alexdewar: This seems to have been forgotten

finesse/gui/temp_control.py
```diff
@@ -41,138 +34,147 @@ def _create_controls(self) -> QGridLayout:
...
-            None
-        """
-        xdata = list(self._bb_hot_line[0].get_xdata())
-        y1data = list(self._bb_hot_line[0].get_ydata())
-        y2data = list(self._bb_cold_line[0].get_ydata())
+    def _update_figure(self, event) -> None:
+        """Updates the matplotlib figure to be contained within the panel."""
+        xdata = list(self._ax["hot"].lines[0].get_xdata())
+        y1data = list(self._ax["hot"].lines[0].get_ydata())
+        y2data = list(self._ax["cold"].lines[0].get_ydata())
```
alexdewar, line 94: These variables could be given more descriptive names. How about `hot_data` and `cold_data`?

finesse/gui/temp_control.py
```diff
@@ -41,138 +34,147 @@ def _create_controls(self) -> QGridLayout:
...
-        self._ax2.set_ylim([self._ax2.get_ylim()[0], 2 * self._ax2.get_ylim()[1]])
+        self._ax["hot"].lines[0].set_xdata(xdata)
+        self._ax["hot"].lines[0].set_ydata(y1data)
+        self._ax["cold"].lines[0].set_xdata(xdata)
+        self._ax["cold"].lines[0].set_ydata(y2data)
+
+        self._ax["cold"].lines[0].set_linestyle(linestyle)
+        self._ax["hot"].lines[0].set_linestyle(linestyle)
+        self._ax["cold"].lines[0].set_marker(marker)
+        self._ax["hot"].lines[0].set_marker(marker)
```
alexdewar, line 127: Do you need to explicitly set these every time you update the plot? Assuming you only need to do it once, it should be done in the constructor.

finesse/gui/temp_control.py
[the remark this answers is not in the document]
alexdewar: This does need to be sorted before we can merge unfortunately because (I think) it'll lead to broken executables being generated by `pyinstaller`, which we need. (If you want to check whether it works locally, you can run `pyinstaller FINESSE.spec` then run whatever it puts in `dist/`.) I'd do all the other bits first though and then maybe we can tackle this together.

What were the errors that GitHub was reporting? Your penultimate commit on this branch was failing but when I looked into it it seemed to be a spurious test failure on Windows so I reran it and it worked the second time.

finesse/gui/temp_control.py
```diff
@@ -181,46 +183,52 @@ def _create_controls(self) -> QGridLayout:
...
+    def __init__(self, name: str) -> None:
+        """Creates the widgets to control and monitor a TC4820.
 
-        Returns:
-            None
+        Args:
+            name (str): Name of the blackbody the TC4820 is controlling
         """
-        super().__init__("TC4820 HOT")
+        super().__init__("TC4820 %s" % name.upper())
```
alexdewar, line 195: Nowadays in Python it's better to use f-strings:

```suggestion
        super().__init__(f"TC4820 {name.upper()}")
```

finesse/gui/temp_control.py
```diff
@@ -181,46 +183,52 @@ def _create_controls(self) -> QGridLayout:
...
+        """Creates the overall layout for the panel.
 
+        Returns:
+            QGridLayout: The layout containing the figure.
+        """
         layout = QGridLayout()
 
+        align = Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter
+
         wid = QLabel("CONTROL")
```
alexdewar, line 210: It's good that you've defined `align` above, but I don't really like that you're reusing `wid`. I'd either give the variables different names or make a separate function for doing this like so:

```py
def _add_label(name: str, row: int, col: int):
```

pyproject.toml
```diff
@@ -13,6 +13,8 @@ python = ">=3.9,<3.12"
 PySide6 = "^6.4.0"
 platformdirs = "^2.5.4"
 matplotlib = "^3.6.2"
+pypubsub = "^4.0.3"
+pyserial = "^3.5"
```
alexdewar, line 17: I'm a bit confused by this change. It looks like you added this explicitly in d4c9452f3640aa1de22d25059bc2c64af9a407c2 but the changelog doesn't say why you're adding dependencies (I know they've already been added on `main` though). Were you trying to prevent a merge conflict?

alexdewar (changes requested):
Getting there! I've made some more minor suggestions.

The main blocker I think is the whole image file path issue (I've added a comment about this) but let's deal with that last.

alexdewar:
Also the merge conflicts need to be resolved, obviously. The conflicts in `pyproject.toml`/`poetry.lock` might be a consequence of changes made in d4c9452f3640aa1de22d25059bc2c64af9a407c2.

finesse/gui/temp_control.py
dc2917: This is just so that the plot always shows a fixed (maximum) number of points, which helps it update nicely. The `None`s get overwritten as more points are collected, rather than appending to a list. 

But yeah, I'm not sure whether it's better/preferable to have a fixed number of points or show a fixed time interval so this may change anyway.

finesse/gui/temp_control.py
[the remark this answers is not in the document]
dc2917: I've now added a placeholder function which returns the time (the _x_ value) and two temperatures (the _y_ values), which is called within this method to update the figure.  I've kept these `append`s here as the lists being appended to pertain only to the figure, but I think this addresses the issue of keeping the two concerns (i.e. getting the data and plotting the data) separate.

finesse/gui/temp_control.py
dc2917: My thinking was that when there is only one data point (or pair of points), `plt.plot` can't draw a line and so just draws the marker. But yeah, I definitely shouldn't be setting these each time the figure is updated, and it doesn't seem worth doing an `if-else` check every time.
I've now changed this so that the marker and linestyle are set in the constructor. I guess there are both scientific and aesthetic reasons for joining up measured data or not, but that's probably not critical at this stage.

pyproject.toml
dc2917: Yes this was to try and deal with the merge conflict 😅

finesse/gui/temp_control.py
alexdewar: Good stuff. If the researchers aren't happy with how it looks we can always tweak it.

pyproject.toml
alexdewar: Ah I see. It looks like you didn't actually merge in `main` though... right? To fix the conflicts you do actually need to do a merge operation otherwise you just confuse git and get more merge conflicts. (The other thing is that if you're merging in changes from somewhere else it's nice to have a commit saying this has happened.)

You will need to do it locally because you'll need to manually regenerate `poetry.lock` when you're done. Lemme know if you get stuck.

finesse/gui/temp_control.py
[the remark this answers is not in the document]
dc2917: I think I've finally sorted this out. In case it's useful for future reference, here's what I had to do:

I'm using `importlib`, which seems to be the preferred way to include static files now. One significant difference between `importlib` and `pkg_resources` is that the directory containing the static files needs an `__init__.py` to be properly treated as a module. This allowed me to load in the data from the image files using `importlib.resources.read_binary()`, but the image files still weren't being picked up by `PyInstaller`. I therefore needed to add the images "module" as a hiddenimport, and also tell `PyInstaller` to include the image files in the build, placing them in the `images` module.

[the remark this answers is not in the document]
alexdewar: Good work! Sorry it was such a faff :grimacing:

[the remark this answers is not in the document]
alexdewar: Lemme know when you're ready for re-review

finesse/gui/main_window.py
```diff
@@ -40,6 +48,20 @@ def __init__(self) -> None:
...
+
+        sub_layout_right = QGridLayout()
+        sub_layout_right.addWidget(bb_monitor, 0, 0, 1, 2)
+        sub_layout_right.addWidget(dp9800, 1, 0, 1, 2)
+        sub_layout_right.addWidget(tc4820_hot, 2, 0, 1, 1)
+        sub_layout_right.addWidget(tc4820_cold, 2, 1, 1, 1)
+        sub_right = QWidget()
+        sub_right.setLayout(sub_layout_right)
+        layout_right.addWidget(sub_right)
+
```
alexdewar, line 64: I don't think you need a sublayout for the right-hand side. You should be able to put all the widgets on the right side in a single `QGridLayout`, as I've done here: https://github.com/ImperialCollegeLondon/FINESSE/blob/gui_merge_for_jon/finesse/gui/main_window.py

The `opus` control will need to span two columns.

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def _update_figure(self, event) -> None:
+        """Updates the matplotlib figure to be contained within the panel."""
+        time = list(self._ax["hot"].lines[0].get_xdata())
+        hot_data = list(self._ax["hot"].lines[0].get_ydata())
+        cold_data = list(self._ax["cold"].lines[0].get_ydata())
+
+        time.pop(0)
+        hot_data.pop(0)
+        cold_data.pop(0)
```
alexdewar, line 127: This is ok for now, but at some point in the future we might want to be able to view more than the last `n` points (e.g. with a button that lets the user view the whole figure).

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,311 @@
...
+        self._ax["cold"].set_ylim(  # Confines "cold" line to lower half of plot
+            [self._ax["cold"].get_ylim()[0], 2 * self._ax["cold"].get_ylim()[1]]
+        )
+
+        xticks = self._ax["hot"].get_xticks()
+        xticklabels = [""] * len(xticks)
+        for i in range(len(xticks)):
+            t = datetime.fromtimestamp(xticks[i])
+            xticklabels[i] = t.strftime("%H:%M:%S")
+        #        self._ax["cold"].set_xticks(xticks)
```
alexdewar, line 152:
```suggestion
```

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+class DP9800(QGroupBox):
+    """Widgets to view the DP9800 properties."""
+
+    def __init__(self, num_channels: int) -> None:
+        """Creates the widgets to monitor DP9800.
+
+        Args:
+            num_channels (int): Number of Pt 100 channels being monitored
```
alexdewar, line 165: Not a problem, but FYI you don't need to specify to type again in the docstring. MkDocs et al. are clever enough to extract the types from the function signature

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,311 @@
...
+            label.setAlignment(Qt.AlignmentFlag.AlignCenter)
+            layout.addWidget(label, 0, i + 1)
+            tbox = QLineEdit()
+            self._channels.append(tbox)
+            self._channels[i].setReadOnly(True)
+            layout.addWidget(self._channels[i], 1, i + 1)
+
+        label = QLabel("POLL")
+        align = Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter
+        label.setAlignment(align)
```
alexdewar, line 195:
```suggestion
        label.setAlignment(Qt.AlignmentFlag.AlignRight | Qt.AlignmentFlag.AlignVCenter)
```

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,311 @@
...
+        layout.addWidget(QLabel("Pt 100"), 1, 0)
+        self._channels = []
+        for i in range(self._num_channels):
+            label = QLabel("CH_%d" % (i + 1))
+            label.setAlignment(Qt.AlignmentFlag.AlignCenter)
+            layout.addWidget(label, 0, i + 1)
+            tbox = QLineEdit()
+            self._channels.append(tbox)
+            self._channels[i].setReadOnly(True)
+            layout.addWidget(self._channels[i], 1, i + 1)
```
alexdewar, line 191: A couple of things:

- You can use an `f`-string for the `QLabel`
- You refer to the same thing as `tbox` and `self._channels[i]`

```suggestion
        self._channels = []
        for i in range(self._num_channels):
            label = QLabel(f"CH_{i+1}")
            label.setAlignment(Qt.AlignmentFlag.AlignCenter)
            layout.addWidget(label, 0, i + 1)

            channel = QLineEdit()
            channel.setReadOnly(True)
            layout.addWidget(channel, 1, i + 1)
            self._channels.append(channel)
```

alexdewar (changes requested):
It's getting there, but there are a few more small things I'd change. Good work on fixing the `pyinstaller` problems btw :smile:

alexdewar (approved):
Looking good! Sorry about all the rounds of re-review but I think it's in a nice state now.

dc2917:
> Looking good! Sorry about all the rounds of re-review but I think it's in a nice state now.

No worries at all - it's good for me to get the feedback!

finesse/gui/temp_control.py
```diff
@@ -0,0 +1,312 @@
...
+    QProgressBar,
+    QPushButton,
+    QSpinBox,
+    QWidget,
+)
+
+poll_on_img_data = resources.read_binary("finesse.gui.images", "poll_on.png")
+poll_off_img_data = resources.read_binary("finesse.gui.images", "poll_off.png")
+alarm_on_img_data = resources.read_binary("finesse.gui.images", "alarm_on.png")
+alarm_off_img_data = resources.read_binary("finesse.gui.images", "alarm_off.png")
```
dalonsoa, line 25: It seems that `read_binary` [is deprecated since Python 3.11](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.read_binary).  

I'm sure you have considered this - and possibly discarded it - but I honestly think it is cleaner to just use the relative location of the images folder with respect to this one and open an IO stream:

```suggestion
images = Path(__file__).parent / "images"
poll_on_img_data = open(images / "poll_on.png", "rb")
poll_off_img_data = open(images /  "poll_off.png", "rb")
alarm_on_img_data = open(images /  "alarm_on.png", "rb")
alarm_off_img_data = open(images /  "alarm_off.png", "rb")
```
(Or something along this lines. I'm not totally sure if this open the file in the correct mode)

This will also save the issue with the hidden imports in Pyinstaller.

dalonsoa (approved):
This looks good! Just a minor comment. I've the feeling it will need to change substantially once we are retrieving the actual data from the sensors, but it is good to have all those placeholders to see what things look like.

Issue #33, opened 2022-11-08
Add panel for setting sequence of mirror positions


alexdewar:
Currently this is done via a bespoke scripting language (#16) and the GUI allows the user to choose a script file.

Depending on the design decisions we make, though, we may not want to use a script file at all but rather let the user choose the positions via a point-and-click interface.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-23

alexdewar:
@jonemurray Any thoughts about this?

I was wondering how complex the scripting language you designed is, e.g. does it have loops or anything in it? Or is it more like a static list of move instructions?

alexdewar:
Thanks @jonemurray.

@dalonsoa Given Jon's response, I think maybe this one needs a bit more consideration and discussion with him, so perhaps we should hold off on it for now and do the other GUI-related tasks first.

The solutions to this as I see it are:

- Port the existing "scripting" language
- Replace it with a point-and-click interface (complicated by the fact that there are loops)
- Allow the user to control it with little Python scripts (which means we would have to expose the interface for moving the mirror somehow)

dalonsoa:
I think the most sustainable solution will new to port the scripting language to a standard, and ideally human readable, format like YAML or TOML, so it is not an obscure combination of numbers and letters. 

Anyway, something to have a chat about, for sure.

jonemurray:
Maybe we should have a brief discussion next week. Better still I can demonstrate the script if you want to pop over. Monday (not between 12-13) or Tuesday (not between 11:30 and 12:30, are good for me the rest of the week is full of meetings.

alexdewar:
Thanks @jonemurray. @dalonsoa shall we have a chat about this today? (I thought I'd stick a note here so I don't forget to bring it up)

jonemurray:
Hi Alex,

Yes let’s talk, I’m free other than between 12 and 1. So just suggest a time.

All the best, Jon.

PR #63, opened 2022-12-02
Add basic support for measure scripts

### LINKED ISSUE #16: Add ability to move mirror to a sequence of positions
Depends on #15.

I believe this behaviour can be controlled with the old software by entering commands into a file; I'm not sure how elaborate this "scripting language" is though.

In the proposal, two possibilities are suggested for achieving this functionality: firstly, by replicating this scripting functionality and secondly by creating a point-and-click GUI control. Which of these seems like the better option will depend on a) getting a better understanding of the problem and b) talking to the researchers (i.e. Jon) about what they'd prefer.

If a more complex scripting functionality is required, we could allow users to do this by writing python code, however I suspect this may be overkill.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-10

alexdewar:
@jonemurray: I've tagged you as a reviewer for this PR, but don't feel any pressure to actually do anything, unless you want to! I just thought you might have an opinion on what I've done here, so I thought I'd tag you in it and give you a chance to have a look at the code, if you like.

This PR adds some basic code and documentation for what I've called "measure scripts". (I think in the old program they're  just referred to as "scripts" but I thought that was a bit ambiguous.)

The idea is that the program needs to be able to move the mirror to a sequence of positions, taking a specified number of recordings at each one. The whole process can be repeated an arbitrary number of times. We decided against keeping the old file format for these measure scripts and so I've implemented this using YAML. I've added the details to a doc in this PR. Obviously, we don't have code for controlling the actual mirror yet, so currently the code just controls a `DummyStepperMotor` by sending `"stepper.move"` messages.

I've also written code for creating and editing these files with a point-and-click interface (see #33), but mindful of discussions we've had elsewhere recently about overlong PRs, I thought it would be best to split it up :wink: so we can discuss the file format separately from the GUI code to manipulate it.

Closes #16.


docs/measure_scripts.md
```diff
@@ -0,0 +1,25 @@
...
+
+```yaml
+measurements:
+  count: 10
+  sequence:
+  - angle: zenith
+    count: 1
+  - angle: 10.0
+    count: 3
+```
```
dalonsoa, line 18: Why the top level `measurement`? This can be written in a more simpler form as:

```suggestion
```yaml
count: 10
sequence:
- angle: zenith
  count: 1
- angle: 10.0
  count: 3
```
```

Comment (dalonsoa) [reply docs/measure_scripts.md:18]: Also, I would replace the top level `count` by `repeat` and the inner `count` by `measurements` (or `meas`), so their meaning is more explicit.

File: tests/test_measure_script.py [hunk 2]
```diff
@@ -0,0 +1,84 @@
+"""Tests for the measure script code."""
+import pytest
+
+from finesse.gui.measure_script.parse import ParseError, parse_script
+
+
+def test_valid_script() -> None:
```
dalonsoa, line 7: As pointed out in another PR, this could be made more concise using parametric testing.

dalonsoa (changes requested):
Looks good. My main comment is that I think the structure of the file should be simplified by removing the top level `measurements` section, as well as renaming some of the fields.

finesse/gui/measure_script/parse.py
```diff
@@ -0,0 +1,91 @@
...
+        {
+            "measurements": {
+                "count": count_type,
+                "sequence": [{"angle": Or(float, valid_preset), "count": count_type}],
+            }
+        }
+    )
+
+    try:
+        return schema.validate(yaml.safe_load(script))["measurements"]
```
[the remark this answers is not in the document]
alexdewar: That case will be caught by the `schema.validate()`.

docs/measure_scripts.md
alexdewar: In my head I had the idea that we might want to add another top-level key at some point. We probably should have a way of versioning the file format (with a `version` key or something) in case we decide to add features in future. And it sort of files like that info shouldn't be at the same level as things that are *actually* to do with how measurements will be recorded, e.g. `count` etc. What do you think?

I take your point about the other keys and I'll change those.

dalonsoa: Ok. Is there any hint that other features will be necessary in a measure script in the future? If not, I would not make the nesting deeper than needed - certainly not if the only other reason is to keep `version` - which I think it is a very appropriate field - separated from the others.

alexdewar: Okey doke. These scripts are only ever going to be for describing the parameters to collect measurements, so I can't see there being any more top-level keys needed. Let's just drop the `measurements` one.

I suppose it would probably be useful to have a top-level `metadata` key with the creation date, program version etc. too as well as the file format version while we're at it.

tests/test_measure_script.py
alexdewar: Good shout. I wrote this before that discussion, but you're quite right! I'll do that.

docs/measure_scripts.md
dalonsoa: Having that top level metadata key with other stuff makes more sense to me than the measurement one, so go ahead with it if you want.

alexdewar:
Right, I've added some parametrised tests now so I think we're good to go! I added some missing checks to the code along the way that I wouldn't have spotted if I hadn't done this, so it has been a useful exercise.

finesse/gui/measure_script/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Measure scripts represent a sequence of mirror positions for a recording.
+
+This module contains code for saving and loading mirror scripts, as well as creating and
```
dc2917, line 3: mirror scripts -> measure scripts?

alexdewar: Good spot!

dalonsoa: Well, actually itbis both, measure and mirror scripts as they set the position pf the mirror

alexdewar: True. I did toy with calling them "mirror scripts" initially, but eventually I settled on calling them "measure scripts" everywhere to avoid confusion.

docs/measure_scripts.md
```diff
@@ -0,0 +1,24 @@
...
+Here is an example:
+
+```yaml
+count: 10
+sequence:
+- angle: zenith
+  count: 1
+- angle: 10.0
+  count: 3
+```
```
dalonsoa, line 17: I thought we have agreed to give a more informative name to the `count` keyword in the two contexts. See https://github.com/ImperialCollegeLondon/FINESSE/pull/63#discussion_r1043039325

finesse/gui/measure_script/parse.py
```diff
@@ -0,0 +1,115 @@
...
+            sequence: Different measurements (i.e. angle + count) to record
+        """
+        self.path = path
+        self.count = count
+        self.sequence = [Measurement(**val) for val in sequence]
+
+    def run(self) -> None:
+        """Run this measure script."""
+        logging.info(f"Running {self.path}")
+        for i in range(self.count):
```
dalonsoa, line 56: I'd use [`tqdm`](https://pypi.org/project/tqdm/) here. Eventually, we will want a GUI-based progress bar somewhere that also allows to cancel the sequence, but that will be work for the future, when the basic stuff is up and running. Something like [QProgressDialog](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QProgressDialog.html). Maybe open an issue about it, so we do not forget.

dalonsoa (changes requested):
I"ve a couple of comments, but it looks OK, overall.

dc2917 (approved):

docs/measure_scripts.md
alexdewar: My bad. I'll do that now

finesse/gui/measure_script/parse.py
alexdewar: This is a neat package. I knocked something together using it, but the problem is that it obscures the console output and atm that's the only acknowledgement we get that the mirror has been "moved" to a given position, so I think we should drop it for now.

I've opened an issue: #71

finesse/gui/measure_script/parse.py
```diff
@@ -0,0 +1,120 @@
...
+from PySide6.QtWidgets import QWidget
+from schema import And, Or, Schema, SchemaError
+
+from ...config import ANGLE_PRESETS
+from ..error_message import show_error_message
+
+
+@dataclass
+class Measurement:
+    """Represents a single measurement (i.e. angle + number of measurements)."""
```
dalonsoa, line 19: Should this be called `Step` rather than `Measurement` to avoid mixing it with the individual acquisitions?

dalonsoa (approved):
Looks good! I've just made a comment.

finesse/gui/measure_script/parse.py
alexdewar: Good point.

PR #65, opened 2022-12-09
Add support for TC4820 temperature controllers

alexdewar:
So this is the result of today's hacking. I managed to get to the stage with the temperature controllers that we can choose the set-point value for the temperature as well as reading various properties back from them. Currently it's entirely standalone and doesn't interact with the GUI in any way.

It would also be good to have documentation and a dummy version of the class for testing, but I thought I might as well get this stuff reviewed and merged in the meantime.


alexdewar:
I just realised I included an unrelated refactor of `SerialPortControl` to hardcode the baud rates, but I can drop this commit if you like. I do plan on using this change elsewhere!

finesse/__init__.py
```diff
@@ -1,17 +1,17 @@
 """The main module for the FINESSE program."""
 __version__ = "0.1.0"
 
-import sys
 
-from PySide6.QtWidgets import QApplication
+def run() -> None:
```
dalonsoa, line 5: I'm totally fine with this, but why did you moved all these imports within the function?

finesse/gui/serial_view.py
```diff
@@ -14,7 +16,6 @@ def __init__(
         self,
         devices: Dict[str, Dict[str, str]],
         avail_ports: Sequence[str],
-        avail_baud_rates: Sequence[str],
```
dalonsoa, line 17: I would not remove this, but rather set as default value the imported `BAUDRATES`. That way, the result is more functional, with the functionality depending entirely on the inputs and not on constants defined somewhere else.

finesse/hardware/tc4820.py
```diff
@@ -0,0 +1,137 @@
...
+            port: Serial port name
+            baudrate: Serial port baudrate
+            serial_args: Extra arguments to Serial constructor
+            serial_kwargs: Extra keyword arguments to Serial constructor
+        """
+        self.serial = Serial(port, baudrate, *serial_args, **serial_kwargs)
+
+    def read_int(self) -> int:
+        """Read a message from the TC4820 and decode the number as a signed integer."""
+        # TODO: Handle errors?
```
dalonsoa, line 33: Most definitely. This application - or more likely the follow up to be flown in a plane - need to be extremely robust, so we will need to implement a way of recovering from errors, when they happen.

finesse/hardware/tc4820.py
```diff
@@ -0,0 +1,137 @@
...
+            raise ValueError("Bad checksum received")
+
+        # Turn the hex string into raw bytes, then convert the raw bytes to a signed int
+        raw_bytes = bytes.fromhex(message[1:5])
+        return int.from_bytes(raw_bytes, byteorder="big", signed=True)
+
+    def read_float(self) -> float:
+        """Read a message from the TC4820 and decode the number as a float.
+
+        Float values are encoded as 10x their value then converted to an int.
```
dalonsoa, line 52: How weird...

dalonsoa (changes requested):
Looks good! Obviously, tests are missing. They should be easy to implement by mocking `Serial`.

On a separate PR, and now that you have a clear picture of the interface for the TC4820, we will need a `MockTC4820` that follows more closely that interface, so we can run the tool without the need of the actual controller.

finesse/config.py
```diff
@@ -8,3 +8,6 @@
 
 ANGLE_PRESETS = ("zenith", "nadir", "hot_bb", "cold_bb", "home")
 """Names of preset angles that the mirror can rotate to."""
+
+BAUDRATES = (4800, 9600, 19200, 38400, 57600, 115200)
```
dc2917, line 12: I guess I have a general question about how to best set and use the baud rates. 

I know that the DP9800 uses a baud rate of 38400 - presumably each of the other components have their own preferred or valid baud rate? I understand that these are just used to populate the list boxes, but do we not also want to check that when a baud rate is selected from the list, it corresponds to the correct baud rate for that instrument?

finesse/__init__.py
alexdewar: There's more info in the commit message: 7b6ba3402587eb546437a05b7e36d926475fea72

The problem was that when I tried running the example code with `python -m finesse.hardware.tc4820` it was failing to import these modules. Putting the imports in `run()` fixed it, but I fear my Python-fu isn't good enough to understand what the root of the problem was :smile:

finesse/gui/serial_view.py
alexdewar: Good point. Not sure why I didn't do this in the first place.

finesse/hardware/tc4820.py
alexdewar: Hmm, that's true.

There's also another way I've thought of that things could be more robust: currently when the checksum isn't valid an error is raised, but really we should try retransmitting our message in this case (maybe 3 retries or so would be sensible).

finesse/hardware/tc4820.py
alexdewar: Fixed-point arithmetic is more common with embedded devices because floating-point is much more complicated so you either need a better CPU or a dedicated floating-point unit, which makes things more expensive. It's a bit odd that it's base 10 though.

finesse/config.py
alexdewar: Kind of. The thing is that the old GUI had this as an option and we're trying to replicate that for now. In theory you could change the baudrate for the device if you're reworking the hardware, in which case you would also want to change the value in the program, but this is hardly going to be a common occurrence! Having it as an option in the main window just increases the chance that someone will change it by accident then wonder why it's stopped working.

I think going forward we should hide this feature in a separate settings panel, with the settings defaulting to whatever the baudrate is for each device.

finesse/__init__.py
dalonsoa: Odd. Well, if it works... it's ok!

finesse/hardware/tc4820.py
```diff
@@ -0,0 +1,212 @@
...
+            baudrate: Serial port baudrate
+            timeout: How long to wait for read/write operation
+            max_attempts: Maximum number of attempts for requests
+            serial_args: Extra arguments to Serial constructor
+            serial_kwargs: Extra keyword arguments to Serial constructor
+        """
+        # If the user hasn't specified an explicit timeout for write operations, then
+        # use the same as for read operations
+        if "write_timeout" not in serial_kwargs:
+            serial_kwargs["write_timeout"] = timeout
```
dalonsoa, line 58: I think we have had this conversation before, but I would put the description of what the method is doing in the docstring. Especially this one, which is not an implementation detail but something that the user should know.

finesse/hardware/tc4820.py
```diff
@@ -0,0 +1,212 @@
...
+            SerialException: An error occurred while reading the device
+        """
+        message_bytes = self.serial.read_until(b"^", size=8)
+
+        # Don't handle decoding errors, because these will be caught by bytes.fromhex()
+        # below
+        message = message_bytes.decode("ascii", errors="replace")
+
+        if len(message) != 8 or message[0] != "*" or message[-1] != "^":
+            raise MalformedMessageError("Malformed message received: {message}")
```
dalonsoa, line 79: I would add an explanation in the docstring about what the correctly formatted message should look like. This will simplify life to future developers when trying to figure out why you are making all these checks. The same applies to the checks below.

finesse/hardware/tc4820.py
```diff
@@ -0,0 +1,212 @@
...
+
+        if message == "*XXXX60^":
+            raise MalformedMessageError("Bad checksum sent")
+
+        if message[5:7] != TC4820.checksum(message[1:5]):
```
dalonsoa, line 84: Why not:
```suggestion
        if message[5:7] != self.checksum(message[1:5]):
```
Static methods can still be called in the normal way!

tests/test_tc4820.py
```diff
@@ -0,0 +1,149 @@
...
+    value: int, message: bytes, raises: Any, dev: TC4820, mocker: MockerFixture
+) -> None:
+    """Test TC4820.read()."""
+    with raises:
+        m = mocker.patch("serial.Serial.read_until", return_value=message)
+        assert value == dev.read()
+        m.assert_called_once_with(b"^", size=8)
+
+
+@pytest.mark.parametrize("value", range(0, 0xFFFF, 200))
```
dalonsoa, line 82: Cool! I didn't know that `range` worked with hex!

dalonsoa (approved):
Great! I've a few comments and suggestions, but very minor stuff. 

For the future `MockTC4820`, you can probably subclass `TC4820` and overwrite the problematic methods only. That will ensure the same interface for both interfaces.

finesse/hardware/tc4820.py
alexdewar: Good point

tests/test_tc4820.py
alexdewar: `range()` isn't doing anything special here. `0xFFFF` is still just an int; it's just a more readable way of saying `65535`.

PR #79, opened 2023-01-24
Add interface to OPUS EM27 software

### LINKED ISSUE #78: Fix OPUS panel
Currently the OPUS panel contains various placeholder methods etc. It needs to be updated so that it can retrieve information from the actual OPUS program/server.

alexdewar:
Closes #78.

This completes the integration with the OPUS software, which is used to control the Bruker EM27 spectrometer. Scans can be started and stopped and the status can be queried by connecting to this software via an HTTP-based protocol. The software is proprietary and there is currently only a licence to run it on a single machine.

Commit summary:

- Move the OPUS backend out of `finesse.gui` into `finesse.hardware`
- Parse the returned HTML-formatted information returned by OPUS in response to requests to get error status etc.
- Handle HTTP requests on a background `QThread` so they don't hang the GUI
- Also log errors during requests in the GUI

Limitations:

- The IP address of the machine running OPUS is hardcoded. Not really a problem (it's a single-machine setup) but it isn't very clean
- Currently when you check the status, the status page is loaded once for parsing and once again to render it. @jonemurray is in favour of removing the web panel in future though, so I think it's fine as a stopgap
- It uses the `beautifulsoup4` library for parsing HTML, which may be overkill
- To stop the request `QThread` on exit, we wait for a message that the main window is closing and explicitly delete the `OPUSInterface`, which a) feels like a hack and b) sometimes temporarily hangs the GUI if the thread is busy doing things
- I haven't implemented automatic enabling/disabling of buttons based on actions that are available (e.g. starting/stopping etc.)
- The "OPUS" button still doesn't do anything

### DIFF
--- a/finesse/gui/opus_view.py
+++ b/finesse/gui/opus_view.py
@@ -40,7 +40,7 @@ class OPUSControl(QGroupBox):
         self.ip = ip
         self.commands = commands if commands is not None else COMMANDS
         self.status: QWebEngineView
-        self.log_handler: OPUSLogHandler
+        self.logger = logging.getLogger("OPUS")
 
         layout = self._create_controls()
         self.setLayout(layout)
@@ -95,7 +95,7 @@ class OPUSControl(QGroupBox):
         """
         log_box = QGroupBox("Error log")
         log_area = QTextBrowser()
-        self.log_handler = OPUSLogHandler.set_handler(log_area)
+        OPUSLogHandler.set_handler(self.logger, log_area)
 
         _layout = QVBoxLayout()
         _layout.addWidget(log_area)
@@ -155,7 +155,7 @@ class OPUSControl(QGroupBox):
         self.status.load(QUrl(self.get_action_url("status")))
         self.status.show()
 
-        logging.getLogger("OPUS").error("Oh, no! Something bad happened!")
+        self.logger.error("Oh, no! Something bad happened!")
 
     def open_opus(self) -> None:
         """Opens OPUS front end somewhere else.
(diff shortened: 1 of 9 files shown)


finesse/gui/opus_view.py
```diff
@@ -15,38 +16,35 @@
...
-    "stop": "opusrs/cmd.htm?opusrsstop",
-    "cancel": "opusrs/cmd.htm?opusrscancel",
-    "opus": "TBC",
-}
+COMMANDS = ["cancel", "stop", "start", "connect"]
+"""The default commands shown for interacting with OPUS."""
 
 
 class OPUSControl(QGroupBox):
     """Class that monitors and control the OPUS interferometer."""
```
dc2917, line 24:
```suggestion
    """Class that monitors and controls the OPUS interferometer."""
```

finesse/hardware/em27_opus.py
```diff
@@ -0,0 +1,147 @@
...
+        try:
+            if response.status_code != 200:
+                raise OPUSError(f"HTTP status code {response.status_code}")
+
+            status: Optional[int] = None
+            text = ""
+            errcode: Optional[int] = None
+            errtext = ""
+            soup = BeautifulSoup(response.content, "html.parser")
+            for td in soup.find_all("td"):
```
dc2917, line 102: I guess as long as the html doesn't decide to capitalise all its tags this should be fine!

alexdewar: Fingers crossed! If it does we can always revisit this.

dalonsoa (approved):
It looks good and nothing to add from my side.

alexdewar:
Great. Thank you for reviews!

PR #80, opened 2023-01-24
Add initial support for ST10 stepper motor controller

alexdewar:
This PR adds a class (`ST10Controller`) for communicating with the stepper motor controller over a USB serial connection, though this support isn't integrated with the GUI yet.

Summary:

- Make a `StepperMotorBase` class as a parent for `ST10Controller` and `DummyStepperMotor`
- Add support for getting and setting the angle of the ST10, including moving it to preset positions

Limitations:

- As with the TC4820, there is no integration with the GUI yet and errors in the backend will only be displayed by the logger

### DIFF
--- a/finesse/hardware/dummy_stepper_motor.py
+++ b/finesse/hardware/dummy_stepper_motor.py
@@ -16,8 +16,8 @@ class DummyStepperMotor:
         Args:
             steps_per_rotation: Number of motor steps for an entire rotation (360°)
         """
-        if steps_per_rotation < len(ANGLE_PRESETS):
-            raise ValueError(f"steps_per_rotation must be >={len(ANGLE_PRESETS)}")
+        if steps_per_rotation < 1:
+            raise ValueError("steps_per_rotation must be at least one")
 
         self.steps_per_rotation = steps_per_rotation
         self.current_step = 0
@@ -25,15 +25,18 @@ class DummyStepperMotor:
         pub.subscribe(self.move_to, "stepper.move")
 
     @staticmethod
-    def get_preset_step(name: str) -> int:
+    def get_preset_angle(name: str) -> float:
         """Get the angle for one of the preset positions.
 
         Args:
             name: Name of preset angle
         Returns:
-            The step number (as int)
+            The angle in degrees
         """
-        return ANGLE_PRESETS.index(name)
+        try:
+            return ANGLE_PRESETS[name]
+        except KeyError as e:
+            raise ValueError(f"{name} is not a valid preset") from e
 
     def move_to(self, target: Union[float, str]) -> None:
         """Move the motor to a specified rotation.
@@ -42,15 +45,11 @@ class DummyStepperMotor:
             target: The target angle (in degrees) or the name of a preset
         """
         if isinstance(target, str):
-            self.current_step = DummyStepperMotor.get_preset_step(target)
-            target_angle = self.current_step * 360 / self.steps_per_rotation
-        else:
-            step = round(self.steps_per_rotation * target / 360.0)
-            if step < 0 or step >= self.steps_per_rotation:
-                raise ValueError("step number is out of range")
-            self.current_step = step
-            target_angle = target
+            target = DummyStepperMotor.get_preset_angle(target)
 
-        logging.info(
-            f"Moving stepper motor to step {self.current_step} (={target_angle}°)"
-        )
+        step = round(self.steps_per_rotation * target / 360.0)
+        if step < 0 or step >= self.steps_per_rotation:
+            raise ValueError("step number is out of range")
+        self.current_step = step
+
+        logging.info(f"Moving stepper motor to step {self.current_step} (={target}°)")
--- a/finesse/hardware/dummy_stepper_motor.py
+++ b/finesse/hardware/dummy_stepper_motor.py
@@ -47,9 +47,9 @@ class DummyStepperMotor:
         if isinstance(target, str):
             target = DummyStepperMotor.get_preset_angle(target)
 
+        if target < 0.0 or target > 270.0:
+            raise ValueError("Angle must be between 0° and 270°")
         step = round(self.steps_per_rotation * target / 360.0)
-        if step < 0 or step >= self.steps_per_rotation:
-            raise ValueError("step number is out of range")
         self.current_step = step
 
         logging.info(f"Moving stepper motor to step {self.current_step} (={target}°)")
--- a/finesse/hardware/dummy_stepper_motor.py
+++ b/finesse/hardware/dummy_stepper_motor.py
@@ -18,16 +18,23 @@ class DummyStepperMotor(StepperMotorBase):
         if steps_per_rotation < 1:
             raise ValueError("steps_per_rotation must be at least one")
 
-        self.steps_per_rotation = steps_per_rotation
-        self.current_step = 0
+        self._steps_per_rotation = steps_per_rotation
+        self._step = 0
 
         pub.subscribe(self.move_to, "stepper.move")
 
-    def get_steps_per_rotation(self) -> int:
-        """Get the number of steps that correspond to a full rotation."""
-        return self.steps_per_rotation
+    @property
+    def steps_per_rotation(self) -> int:
+        """The number of steps that correspond to a full rotation."""
+        return self._steps_per_rotation
 
-    def move_to_step(self, step: int) -> None:
+    @property
+    def step(self) -> int:
+        """The number of steps that correspond to a full rotation."""
+        return self._step
+
+    @step.setter
+    def step(self, step: int) -> None:
         """Move the stepper motor to the specified absolute position."""
-        self.current_step = step
+        self._step = step
         logging.info(f"Moving stepper motor to step {step}")
--- a/finesse/hardware/dummy_stepper_motor.py
+++ b/finesse/hardware/dummy_stepper_motor.py
@@ -1,13 +1,12 @@
 """Code for a fake stepper motor device."""
 import logging
-from typing import Union
 
 from pubsub import pub
 
-from ..config import ANGLE_PRESETS
+from .stepper_motor_base import StepperMotorBase
 
 
-class DummyStepperMotor:
+class DummyStepperMotor(StepperMotorBase):
     """A fake stepper motor device used for unit tests etc."""
 
     def __init__(self, steps_per_rotation: int) -> None:
@@ -24,32 +23,11 @@ class DummyStepperMotor:
 
         pub.subscribe(self.move_to, "stepper.move")
 
-    @staticmethod
-    def get_preset_angle(name: str) -> float:
-        """Get the angle for one of the preset positions.
+    def get_steps_per_rotation(self) -> int:
+        """Get the number of steps that correspond to a full rotation."""
+        return self.steps_per_rotation
 
-        Args:
-            name: Name of preset angle
-        Returns:
-            The angle in degrees
-        """
-        try:
-            return ANGLE_PRESETS[name]
-        except KeyError as e:
-            raise ValueError(f"{name} is not a valid preset") from e
-
-    def move_to(self, target: Union[float, str]) -> None:
-        """Move the motor to a specified rotation.
-
-        Args:
-            target: The target angle (in degrees) or the name of a preset
-        """
-        if isinstance(target, str):
-            target = DummyStepperMotor.get_preset_angle(target)
-
-        if target < 0.0 or target > 270.0:
-            raise ValueError("Angle must be between 0° and 270°")
-        step = round(self.steps_per_rotation * target / 360.0)
+    def move_to_step(self, step: int) -> None:
+        """Move the stepper motor to the specified absolute position."""
         self.current_step = step
-
-        logging.info(f"Moving stepper motor to step {self.current_step} (={target}°)")
+        logging.info(f"Moving stepper motor to step {step}")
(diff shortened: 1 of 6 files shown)


finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,305 @@
...
+        # motor first. I don't know what the input status actually means, but this is
+        # how it was done in the old program, so I'm copying it here.
+        if self._get_input_status(3):
+            self.relative_move(-5000)
+
+        # Send home command; leaves mirror facing upwards
+        self._write_check("SH6H")
+
+        # Turn mirror so it's facing down
+        self.relative_move(-30130)
```
dalonsoa, line 131: This is highly confusing. How is it possible that to move the mirror 180 degrees (from facing upwards to facing downwards) you need to move anything other than `STEPS_PER_ROTATION/2`? I think there is some sort of inconsistency on what steps mean - or a more detailed explanation is missing.

finesse/hardware/dummy_stepper_motor.py
```diff
@@ -1,13 +1,12 @@
...
-from typing import Union
 
 from pubsub import pub
 
-from ..config import ANGLE_PRESETS
+from .stepper_motor_base import StepperMotorBase
 
 
-class DummyStepperMotor:
+class DummyStepperMotor(StepperMotorBase):
```
dalonsoa, line 9: The actual controller `ST10Controller` has a bunch of public methods like `wait_until_stopped`, `relative_move`, etc. that are neither present in the dummy nor in the base class `StepperMotorBase`. If they are not to be used externally, then they should have an underscore. If they are meant to be used externally, then the dummy should also have them (or potentially the base class), right?

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,305 @@
...
+        Raises:
+            SerialException: Error communicating with device
+            SerialTimeoutException: Timed out waiting for response from device
+            ST10ControllerError: Malformed message received from device
+        """
+        response = self._read()
+
+        # These values are referred to as "ack" and "qack" in the old program (nack is
+        # "?"). I don't know what qack is. Could it mean clockwise/anticlockwise
+        # movement?
```
dalonsoa, line 232: I think this clarification is actually a bit confusing. We are not using any of those terms here, so why to include an explanation about them - related to a program no one (but you) have the source code for.

tests/test_st10_controller.py
```diff
@@ -0,0 +1,136 @@
...
+    """A fixture providing an ST10Controller with a patched Serial object."""
+    serial = MagicMock()
+
+    # check_device_id() and home() should both be called by ST10Controller.__init__(),
+    # but patch them for now as we'll test their inner workings elsewhere
+    with patch.object(ST10Controller, "_check_device_id") as check_mock:
+        with patch.object(ST10Controller, "home") as home_mock:
+            st10 = ST10Controller(serial)
+            check_mock.assert_called_once()
+            home_mock.assert_called_once()
```
dalonsoa, line 24: Why are you asserting things in a fixture?

tests/test_st10_controller.py
```diff
@@ -0,0 +1,136 @@
...
+        dev._read()
+
+
+@pytest.mark.parametrize(
+    "response,raises",
+    [
+        (
+            response,
+            does_not_raise()
+            if response == "%" or response == "*"
```
dalonsoa, line 64:
```suggestion
            if response in ("%", "*")
```

tests/test_st10_controller.py
```diff
@@ -0,0 +1,136 @@
...
+            name,
+            value,
+            response,
+            does_not_raise()
+            if response.startswith(f"{name}=")
+            else pytest.raises(ST10ControllerError),
+        )
+        for name in ["hello", "IS", "SP"]
+        for value in ["", "value", "123"]
+        for response in [f"{name}={value}", value, "%", "*", "?4"]
```
dalonsoa, line 99: For things that are immutable in practice, just use tuples. Is more memory efficient and, I think, faster to access. Probably totally irrelevant in this case, but is just a question of getting in the habit of using the proper structure. 
```suggestion
        for name in ("hello", "IS", "SP")
        for value in ("", "value", "123")
        for response in (f"{name}={value}", value, "%", "*", "?4")
```

dalonsoa (changes requested):
It looks good - although borderline too long for a PR. Probably it should have been split in 2-3 chunks. I've just made some comments and suggestions here and there. Also, some docstrings in the methods do not include a description of the input arguments (eg. `_get_input_status`).

finesse/hardware/st10_controller.py
alexdewar: Good point -- I hadn't noticed this. The total number of steps per rotation *should* be 50,800, based on how the old program converts between steps and angles. I've tried it and the motor seems to move by about the correct amount (e.g. when you move to the hot black body it's pointing in the right direction), so I think this value is probably correct. I suspect that it's simply that the motor case isn't aligned with things like the hot and cold black bodies, so that's why it's rotated by an extra ~35°. In any case, I should at least probably just say that we're doing what the old program did in the comment.

@jonemurray Does this make sense based on what you know about how the device was put together?

finesse/hardware/dummy_stepper_motor.py
alexdewar: Good point. I'll tweak things

tests/test_st10_controller.py
alexdewar: I have to mock them because I don't actually want these methods to run for the fixture, but I would like to check that they actually run. I suppose the danger is that if they fail they'll break all the other tests. I'll put them in a separate test

finesse/hardware/st10_controller.py
alexdewar: Maybe it's not the best explanation. But I feel like ack and nack are fairly standard terms, so it's obvious what the meaning was, but it is a mystery as to why there are two versions of "ack". This seems to be out of scope for the manual I linked to, so probably the only person who knows is Alan, who built the hardware and the old software and is now retired (I think).

I think it is worth flagging up in the comment that there are two "ack" responses and that they may mean slightly different things, even if we currently treat them the same in the program. We don't have any documentation for this so the closest thing is the source code for the old program.

tests/test_st10_controller.py
alexdewar: Done

finesse/hardware/st10_controller.py
alexdewar: What do you think?

dalonsoa: Well, "it's obvious what the meaning was" might be an overstatement. I don't know what the meaning of those terms is, so I would suggest you elaborate on that.

alexdewar: Fair enough, sorry. I'll write a better explanation.

alexdewar: Ha, it turns out that the manual does have an explanation for what these responses mean -- I just wasn't looking hard enough.

alexdewar:
Aside from the whole "mirror being rotated by >180°" thing, I think this is ready for re-review.

I've made a few additional small changes. I added a new `stop_moving()` method which is called when initialising and disconnecting from the device, so that the device will stop moving immediately before we do anything else with it (there could conceivably still be move commands in the buffer that are being executed one by one).

Thanks for the comments -- it spurred me to actually look at the manual properly, so now there's better documentation of what's happening.

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+    def _check_device_id(self) -> None:
+        """Check that the ID is the correct one for an ST10.
+
+        Raises:
+            SerialException: Error communicating with device
+            SerialTimeoutException: Timed out waiting for response from device
+            ST10ControllerError: The device ID is not for an ST10
+        """
+        self._write("MV")
+        if self._read() != "107F024":
```
dalonsoa, line 99: Can we make this number a class attribute or a constant? It looks like a sort of serial number, device dependent, and not something you can get from manual.

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+        # motor first. I don't know what the input status actually means, but this is
+        # how it was done in the old program, so I'm copying it here.
+        if self._get_input_status(3):
+            self._relative_move(-5000)
+
+        # Send home command; leaves mirror facing upwards
+        self._write_check("SH6H")
+
+        # Turn mirror so it's facing down
+        self._relative_move(-30130)
```
dalonsoa, line 137: I still don't get why a relative rotation of 180 degrees is not STEPS_PER_ROTATION/2. I understand that this is what works, but still, it seems totally inconsistent. I'd suggest, when the system is available again, we run some tests about this and try to figure out why this is this way.

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+            - Normal acknowledge ("%"), indicating that the command was received and
+              successfully executed
+            - Exception acknowledge ("*"), indicating that the command was received and
+              added to the execution buffer, i.e. it will be executed after other
+              commands. (Why the manual uses the term "exception", who knows, but it
+              also indicates success!)
+
+        If an error occurs, the device returns a negative acknowledge (NACK) response,
+        which is a "?", followed by an error code, e.g. "?4". The meanings of the error
+        codes are listed in the manual linked in the module description.
```
dalonsoa, line 254: This is much, much better! Thanks!

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+        # Set temporary timeout
+        old_timeout, self.serial.timeout = self.serial.timeout, timeout
+        try:
+            if self._read() != "X":
+                raise ST10ControllerError(
+                    "Invalid response received when waiting for X"
+                )
+        finally:
+            # Restore previous timeout setting
+            self.serial.timeout = old_timeout
```
dalonsoa, line 324: Is it the `finally` block executed even in the case of raising an exception? If not, we should restore the timeout before raising the exception.

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+        # Check that we are connecting to an ST10
+        self._check_device_id()
+
+        # In case the motor is still moving, stop it now
+        self.stop_moving()
+
+        # Move mirror to home position
+        self._home_and_reset()
+
+        super().__init__()
```
dalonsoa, line 54: I would call this the first thing, so any customisations of this class are done on top of whatever the base class has setup.

dalonsoa (approved):
It looks good! Just a couple of minor suggestions.

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+        """
+        self._write("MV")
+        if self._read() != "107F024":
+            raise ST10ControllerError("Device ID indicates this is not an ST10")
+
+    def _get_input_status(self, index: int) -> bool:
+        """Read the status of the device's inputs.
+
+        The inputs to the controller include digital inputs, as well as other properties
+        like alarm status. The exact meaning seems to varies between boards.
```
dc2917, line 106:
```suggestion
        like alarm status. The exact meaning seems to vary between boards.
```

finesse/hardware/st10_controller.py
```diff
@@ -0,0 +1,343 @@
...
+
+        Args:
+            index: Which boolean value in the input status array to check
+        """
```
dc2917, line 110: Maybe it's worth quoting something from the manual regarding "input status" here, to clarify what's being checked and thus why there's a comparison to the string "1"?

From the manual:
A closed input is represented by a “0” (zero), and an open input is represented by a “1” (one).

dc2917 (approved):
All looks good to me. Some of the ST10Controller methods were a bit strange to read but only because of the built-in instructions. I guess the link to the manual is there in the docstring so it is possible to look up what all of the seemingly weird commands mean.

finesse/hardware/st10_controller.py
alexdewar: My only worry with doing that in this case is that the base class does a `pub.subscribe()` in its `__init__()` and ideally I think we want that to be the last piece of setup we do, partly so we can check that its the right device with `_check_device_id()` before proceeding but also so that there is no chance of the motor being moved in response to a `stepper.move` message before the commands to move it to the home position have been sent (otherwise it'll move to the wrong position). The message would have to be sent from another thread for this to happen, so I don't think it'll happen with the way the code is written now, but we might as well make it thread-safe if we can!

finesse/hardware/st10_controller.py
alexdewar: Done.

finesse/hardware/st10_controller.py
alexdewar: It will be moving more than 180°, it's just that this is what the engineer wanted to define as the origin.

But yes, let's check it ourselves once we've got the kit.

finesse/hardware/st10_controller.py
alexdewar: Yup!

In practice, if an error occurs during the read it's probably game over anyway, but it seems prudent to reset the timeout.

alexdewar:
@dc2917 I've gone through and added comments to say what the mnemonics for the different commands stand for, so hopefully that'll make it a bit less opaque!

PR #81, opened 2023-01-25
Add panel and dialogs for creating and editing measure scripts

alexdewar:
This PR adds a new panel for creating, editing and running measure scripts. I have also added a point-and-click dialog for creating a measure script (or editing an existing one). Running a measure script currently just prints to the log as I still need to hook up the OPUS integration (the first step is #79, but other changes will be needed).

@dalonsoa I'm afraid this PR is a bit on the longer side too. I worked on it a few months ago before the measure script stuff was properly in place so it represents a fair bit of work (albeit now squashed into one commit). Let me know if you'd like me to break it up.

### DIFF
--- a/finesse/gui/measure_script/script_edit_dialog.py
+++ b/finesse/gui/measure_script/script_edit_dialog.py
@@ -1,0 +1,414 @@
+"""Contains code for a dialog to create and edit measure scripts."""
+
+import logging
+from pathlib import Path
+from typing import Any, List, Optional, Union
+
+import yaml
+from PySide6.QtCore import QAbstractTableModel, QModelIndex, QPersistentModelIndex, Qt
+from PySide6.QtGui import QCloseEvent
+from PySide6.QtWidgets import (
+    QAbstractItemView,
+    QButtonGroup,
+    QDialog,
+    QDialogButtonBox,
+    QFileDialog,
+    QFormLayout,
+    QGroupBox,
+    QHBoxLayout,
+    QMessageBox,
+    QPushButton,
+    QSpinBox,
+    QTableView,
+    QVBoxLayout,
+    QWidget,
+)
+
+from ...config import ANGLE_PRESETS, DEFAULT_SCRIPT_PATH
+from ..error_message import show_error_message
+from .parse import Measurement, Script
+from .script_path_widget import ScriptPathWidget
+
+
+class ScriptEditDialog(QDialog):
+    """A dialog to create and edit measure scripts."""
+
+    def __init__(self, parent: QWidget, script: Optional[Script] = None) -> None:
+        """Create a new ScriptEditDialog.
+
+        Args:
+            parent: Parent widget
+            script: A loaded measure script or None
+        """
+        super().__init__(parent)
+        self.setWindowTitle("Edit measurement script")
+
+        if script:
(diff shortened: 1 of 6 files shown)


finesse/gui/measure_script/script_edit_dialog.py
```diff
@@ -0,0 +1,414 @@
...
+        delete.clicked.connect(sequence.delete_selected)  # type: ignore
+
+        clear = QPushButton("Clear")
+        clear.clicked.connect(sequence.delete_all)  # type: ignore
+
+        layout = QVBoxLayout()
+        layout.addWidget(up)
+        layout.addWidget(down)
+        layout.addWidget(delete)
+        layout.addWidget(clear)
```
dalonsoa, line 395: This button is too close to the `delete` one. When trying the thing, after writing a few lines, I click that one by mistake and all was gone wit question. I would, add a confirmation dialog because otherwise the loss of work can be drastic.

dalonsoa (changes requested):
Looks good! Definitely more user friendly and readable than the original approach. I'm not convinced by the layout and spacing of the widgets, but that something that can be tweak at a later time when polishing the look and feel.

I've left a couple of comments, but the main thing is... yes, tests! There is a lot of code here and zero tests. If you prefer to merge this and then do the test separately, that is fine, but tests should be added.

finesse/gui/measure_script/script_edit_dialog.py
alexdewar: Good idea. I've added a confirmation dialog.

I wasn't sure whether the button should live somewhere else, but then I couldn't think of where else to put it. The dialog makes it less dangerous anyways though.

alexdewar:
I've added the confirmation dialog and opened an issue to add tests: #83.

I'll add the tests later, if that's ok. I'm working on a couple of other bits right now, so I think it'd be good to have this bit merged in as it is for now (partly for ease of demoing on Monday). I promise I'll do it soon though!

alexdewar:
(I've also slightly rearranged the GOTO button + angle spinbox for aesthetic reasons.)

finesse/gui/measure_script/script_edit_dialog.py
```diff
@@ -0,0 +1,437 @@
...
+            self.sequence[higher],
+            self.sequence[lower],
+        )
+        self.model.endMoveRows()
+
+    def move_selected_up(self) -> None:
+        """Move the currently selected instructions up one row."""
+        selected = self._get_selected_rows()
+        if not selected or selected[0] == 0:
+            # We can't move up if the topmost item is selected
```
dc2917, line 284: I wonder if it's best to have these buttons disabled unless a row is selected?

dc2917 (approved):
All looks good and seems to work well!

finesse/gui/measure_script/script_edit_dialog.py
alexdewar: Yeah, they really should be. There are actually a bunch of places on the GUI where we should be disabling buttons when they're not usable and aren't.

I'll maybe look at this when I write the tests for this lot.

alexdewar:
Thanks for the review @dc2917 :smile:

dalonsoa (approved):
All good! As long as there is a record that tests need to be added, that's ok.

PR #82, opened 2023-01-25
Add LED icon class

dc2917:
This PR adds a new class for handling LED indicators on the GUI. The main advantage is that the image data are loaded only once, rather than for each separate GUI component that requires them. I have also provided some additional functionality, allowing the LED icons to flash on for a specified duration, once some process triggers them (e.g. polling the server).

### DIFF
--- a/finesse/gui/led_icons.py
+++ b/finesse/gui/led_icons.py
@@ -29,8 +29,8 @@ class LEDIcon(QLabel):
             is_on (bool): On/off status of LED.
         """
         super().__init__()
-        self._on_img = QImage(on_img)
-        self._off_img = QImage(off_img)
+        self._on_img = on_img
+        self._off_img = off_img
         if is_on:
             self._turn_on()
         else:
--- a/finesse/gui/led_icons.py
+++ b/finesse/gui/led_icons.py
@@ -1,4 +1,6 @@
 """Class for LED Icons."""
+from __future__ import annotations
+
 from importlib import resources
 
 from PySide6.QtCore import QTimer
@@ -38,15 +40,15 @@ class LEDIcon(QLabel):
         self._timer = QTimer()
         self._timer.timeout.connect(self._turn_off)  # type: ignore
 
-    @staticmethod
-    def create_poll_icon() -> "LEDIcon":
+    @classmethod
+    def create_poll_icon(cls) -> LEDIcon:
         """Creates the LED icon for polling the server."""
-        return LEDIcon(on_img=_poll_on_img, off_img=_poll_off_img)
+        return cls(on_img=_poll_on_img, off_img=_poll_off_img)
 
-    @staticmethod
-    def create_alarm_icon() -> "LEDIcon":
+    @classmethod
+    def create_alarm_icon(cls) -> LEDIcon:
         """Creates the LED icon to indicate alarm status."""
-        return LEDIcon(on_img=_alarm_on_img, off_img=_alarm_off_img)
+        return cls(on_img=_alarm_on_img, off_img=_alarm_off_img)
 
     def _turn_on(self) -> None:
         """Turns the LED on."""
--- a/finesse/gui/led_icons.py
+++ b/finesse/gui/led_icons.py
@@ -1,6 +1,7 @@
 """Class for LED Icons."""
 from importlib import resources
 
+from PySide6.QtCore import QTimer
 from PySide6.QtGui import QImage, QPixmap
 from PySide6.QtWidgets import QLabel
 
@@ -25,6 +26,8 @@ class LEDIcon(QLabel):
         self._status = status
         self._on_img = QImage()
         self._off_img = QImage()
+        self._timer = QTimer()
+        self._timer.timeout.connect(self._turn_off)
 
     def _turn_on(self):
         """Turns the LED on."""
@@ -36,15 +39,14 @@ class LEDIcon(QLabel):
         self._status = 0
         self.setPixmap(QPixmap(self._off_img))
 
-    def _flash(self, duration: int):
+    def _flash(self, duration: int = 1000):
         """Turns the LED on for a specified duration.
 
         Args:
             duration (int): Number of milliseconds to keep LED lit for
         """
         self._turn_on()
-        # wait for <duration> ms
-        self._turn_off()
+        self._timer.singleShot(duration, self._turn_off)
 
 
 class PollIcon(LEDIcon):
--- a/finesse/gui/led_icons.py
+++ b/finesse/gui/led_icons.py
@@ -20,16 +20,21 @@ _alarm_off_img = QImage.fromData(_alarm_off_img_data)
 class LEDIcon(QLabel):
     """QLabel object to represent an LED with on/off status."""
 
-    def __init__(self, is_on: bool = False) -> None:
+    def __init__(self, on_img: QImage, off_img: QImage, is_on: bool = False) -> None:
         """Creates the LED icon, sets its status and stores corresponding image data.
 
         Args:
+            on_img (QImage): QImage for LED on state.
+            off_img (QImage): QImage for LED off state.
             is_on (bool): On/off status of LED.
         """
         super().__init__()
-        self._is_on = is_on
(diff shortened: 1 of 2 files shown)


finesse/gui/led_icons.py
```diff
@@ -0,0 +1,85 @@
...
+img_files = resources.files("finesse.gui.images")
+poll_on_img_data = img_files.joinpath("poll_on.png").read_bytes()
+poll_off_img_data = img_files.joinpath("poll_off.png").read_bytes()
+alarm_on_img_data = img_files.joinpath("alarm_on.png").read_bytes()
+alarm_off_img_data = img_files.joinpath("alarm_off.png").read_bytes()
+
+poll_on_img = QImage.fromData(poll_on_img_data)
+poll_off_img = QImage.fromData(poll_off_img_data)
+alarm_on_img = QImage.fromData(alarm_on_img_data)
+alarm_off_img = QImage.fromData(alarm_off_img_data)
```
alexdewar, line 17: All these variables are public so they'll be exported from the module. I'd make them private (by putting an underscore in front).

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class LEDIcon(QLabel):
+    """QLabel object to represent an LED with on/off status."""
+
+    def __init__(self, status: int) -> None:
```
alexdewar, line 23: I'd make `status` a `bool` and rename it to something like `is_on` (throughout).

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,85 @@
...
+        """Creates the LED icon, sets its status and stores corresponding image data."""
+        super().__init__()
+        self._status = status
+        self._on_img = QImage()
+        self._off_img = QImage()
```
alexdewar, line 28: Rather than defining these as empty images, I'd have `on_img` and `off_img` as arguments to this function, then your child classes can just set these in their calls to `super().__init__`.

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,85 @@
...
+        Args:
+            status (int): On/off status of LED. 0 = off, !0 = on
+        """
+        super().__init__(status=status)
+        self._on_img = poll_on_img
+        self._off_img = poll_off_img
+        if status:
+            self._turn_on()
+        else:
+            self._turn_off()
```
alexdewar, line 67: These lines could be in the parent class's `__init__()` instead, if you make the change I suggested above.

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,85 @@
...
+            duration (int): Number of milliseconds to keep LED lit for
+        """
+        self._turn_on()
+        self._timer.singleShot(duration, self._turn_off)
+
+
+class PollIcon(LEDIcon):
+    """QLabel object to represent an LED for polling server."""
+
+    def __init__(self, status: int) -> None:
```
alexdewar, line 55: I think `status` should default to off (and in the child classes).

alexdewar (changes requested):
I've made some small suggestions but this looks good :-)

finesse/gui/led_icons.py
```diff
@@ -5,38 +5,47 @@
...
     def _turn_on(self):
         """Turns the LED on."""
-        self._status = 1
+        self._is_on = True
         self.setPixmap(QPixmap(self._on_img))
 
     def _turn_off(self):
         """Turns the LED off."""
-        self._status = 0
+        self._is_on = 0
```
alexdewar, line 48:
```suggestion
        self._is_on = False
```

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,85 @@
...
+        """Turns the LED on."""
+        self._status = 1
+        self.setPixmap(QPixmap(self._on_img))
+
+    def _turn_off(self):
+        """Turns the LED off."""
+        self._status = 0
+        self.setPixmap(QPixmap(self._off_img))
+
+    def _flash(self, duration: int = 1000):
```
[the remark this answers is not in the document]
alexdewar: What about this? ^^

alexdewar (changes requested):
Found a couple of really minor things.

Something else I probably should have asked earlier is, where did you get these icons from? What licence (if any) are they released under?

finesse/gui/led_icons.py
dc2917: > Found a couple of really minor things.

Thanks! I should have been more careful with converting int to bool!

> Something else I probably should have asked earlier is, where did you get these icons from? What licence (if any) are they released under?

Good question - I made them on Inkscape haha so fully free and open source!

alexdewar (approved):
LGTM.

Although in the commit message for 875e3ee you say you're setting the timeout to 100 ms when you've set it to 250 ms. It doesn't particularly matter though.

dc2917:
> LGTM.
> 
> Although in the commit message for [875e3ee](https://github.com/ImperialCollegeLondon/FINESSE/commit/875e3ee4f241aa9e17efa8d8c412a4f5adb65384) you say you're setting the timeout to 100 ms when you've set it to 250 ms. It doesn't particularly matter though.

Wow, I'm really not with it today 😅

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,90 @@
...
+        """Turns the LED on for a specified duration.
+
+        Args:
+            duration (int): Number of milliseconds to keep LED lit for
+        """
+        self._turn_on()
+        self._timer.singleShot(duration, self._turn_off)
+
+
+class PollIcon(LEDIcon):
```
dalonsoa, line 61: I think this and the alarm version should not be subclasses of LEDIcon, but factory methods of LEDicon that create the object with slightly different inputs.

So these two init methods should be converted into two class methods of LEDicon, called constructors. Then, in the tenperature controler, yo use those as:

```python
normalLED = LEDIcon.create_poll_icon()
alarmLED = LEDIcon.create_alarm_icon()
```

In summary, just one class- you just create it in a slightly different way.

dalonsoa (changes requested):
Great improvement of modularity. Just a suggestion to ise constructor.

finesse/gui/led_icons.py
dc2917: Thanks! I knew there must be a neater way haha. I think what I've now done achieves this.

finesse/gui/led_icons.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @staticmethod
+    def create_poll_icon() -> "LEDIcon":
+        """Creates the LED icon for polling the server."""
+        return LEDIcon(on_img=_poll_on_img, off_img=_poll_off_img)
+
+    @staticmethod
+    def create_alarm_icon() -> "LEDIcon":
+        """Creates the LED icon to indicate alarm status."""
+        return LEDIcon(on_img=_alarm_on_img, off_img=_alarm_off_img)
```
dalonsoa, line 49: The usual approach when dealing with factory methods like these is:
```suggestion
    @clasmethod
    def create_poll_icon(cls) -> LEDIcon:
        """Creates the LED icon for polling the server."""
        return cls(on_img=_poll_on_img, off_img=_poll_off_img)

    @clasmethod
    def create_alarm_icon(cls) -> LEDIcon:
        """Creates the LED icon to indicate alarm status."""
        return cls(on_img=_alarm_on_img, off_img=_alarm_off_img)
```

For this to work, you will need to import at the top of the file:

```python
from __futures__ import annotations
```

dalonsoa (approved):
Just a small correction, but looks good otherwise!

alexdewar:
Looks like this is being affected by #95. Should be fixed once #86 is merged.

alexdewar:
Would you like me to merge this when it's fixed btw?

dc2917:
> Looks like this is being affected by #95. Should be fixed once #86 is merged.

Yep, I was glad to see it wasn't me breaking something 😅

dc2917:
> Would you like me to merge this when it's fixed btw?

Yes please, thanks!

PR #98, opened 2023-02-22
Interferometer monitor

### LINKED ISSUE #47: Add panel to display interferometer environment info
Corresponds to green box in #24.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-30

dc2917:
This PR adds the panel for displaying the EM27 properties (the panel labelled "EM27 SOH Monitor" on the original GUI) which are scraped off the web server. There is currently placeholder functionality for polling the server and updating the GUI with the returned values. 

It is currently compatible with the separate module I have been working on do the web-scraping, in that it expects the results in the same format. There will also need to be some functionality to listen for when the connection to the EM27 is opened. I am currently intending to achieve these via `pubsub`.

Force pushed as the branch was rebased, since I initially branched from the wrong place.

Closes #47.


finesse/gui/interferometer_monitor.py
```diff
@@ -0,0 +1,204 @@
...
+    data_table = {
+        "PSF27 Temp": [28.151062, "deg. C"],
+        "Cryo Temp": [0.0, "deg. K"],
+        "Blackbody Hum": [2.463968, "%"],
+        "Source Temp": [70.007156, "deg. C"],
+        "Aux Volt": [6.285875, "V"],
+        "Aux Curr": [0.910230, "A"],
+        "Laser Curr": [0.583892, "A"],
+    }
+    return data_table
```
alexdewar, line 26: I feel like the data structure isn't quite right here, because you're effectively using keys as values and the lists contain two objects of different types. How about defining a dataclass for this, with `name`, `value` and `unit` fields?

finesse/gui/interferometer_monitor.py
```diff
@@ -0,0 +1,204 @@
...
+    def __init__(self, prop_labels: list[str], prop_units: list[Optional[str]]) -> None:
+        """Creates a sequence of widgets to view properties monitored by the EM27."""
+        super().__init__("EM27 SOH Monitor")
+
+        self._prop_labels = deepcopy(prop_labels)
+        self._prop_units = deepcopy(prop_units)
+
+        self._psf27_temp_box = QLineEdit(self._prop_units[0])
+        self._psf27_temp_box.setReadOnly(True)
+        self._psf27_temp_box.setAlignment(Qt.AlignmentFlag.AlignCenter)
```
alexdewar, line 41: All these controls could be added dynamically in a `for` loop. The problem you have at the mo is the code makes assumptions that all the values will be present and in a specific order, which isn't very robust.

finesse/gui/interferometer_monitor.py
```diff
@@ -0,0 +1,204 @@
...
+
+        return layout
+
+    def set_psf27_temp(self, val: tuple[str, float]) -> None:
+        """Sets the PSF27 temperature text box.
+
+        Args:
+            val: value polled from server
+        """
+        self._psf27_temp_box.setText(f"{val[0]:.6f} {val[1]}")
```
alexdewar, line 105: Again, these functions are v boilerplate-y, so I'd figure out a way to do all this without repeating yourself (i.e. so you can just update the relevant text field from the name given to you by the API).

finesse/gui/interferometer_monitor.py
```diff
@@ -0,0 +1,204 @@
...
+        "Aux Curr": [0.910230, "A"],
+        "Laser Curr": [0.583892, "A"],
+    }
+    return data_table
+
+
+class EM27Monitor(QGroupBox):
+    """Widgets to view the EM27 properties."""
+
+    def __init__(self, prop_labels: list[str], prop_units: list[Optional[str]]) -> None:
```
alexdewar, line 32: I don't think we want the `prop_labels` and `prop_units` arguments here. What you've done is similar to what we've done for other placeholder controls, but I would request these values from the server, rather than passing them into the constructor (i.e. just use your `get_vals_from_server()`).

finesse/gui/interferometer_monitor.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def set_laser_current(self, val: tuple[str, float]) -> None:
+        """Sets the laser current text box.
+
+        Args:
+            val: value polled from server
+        """
+        self._laser_current_box.setText(f"{val[0]:.6f} {val[1]}")
+
+    def poll_server(self) -> None:
```
alexdewar, line 155: It seems like you're not calling this from anywhere atm. I'd make it actually work: add a `QTimer` to invoke it every 2s or something.

alexdewar (changes requested):
Thanks for this :smile:. I've made some suggestions. The main one is that you really want this to be more dynamic (i.e. adding controls based on what values the server tells you there are) rather than hardcoding everything.

The other thing is that I would write a dummy backend for this, like I've done for the stepper motor. You could request the values with an `"em27.request"` message and then the backend could return an `"em27.response"` message (or something similar). Bonus points if it returns another message in case of errors. I don't mind whether you do this as part of this PR or separately.

finesse/gui/interferometer_monitor.py
dalonsoa: Or how about using [`pint`](https://pint.readthedocs.io/en/stable/)? Dealing with this properly is the whole point of the tool.

finesse/gui/interferometer_monitor.py
dalonsoa: I disagree here. If initialising this class depends on a connection to a server, it is necessary to mock that in order to test it. For this type of things, I would keep `__init__` as Dan has done it and implement a factory method (like the ones used for `LEDicon`) where  `get_vals_from_server()` is called.

dalonsoa (changes requested):
I've made a couple of comments and suggestions, but otherwise agreeing with @alexdewar with 👍 .

dc2917:
Thanks @alexdewar and @dalonsoa for your feedback.

Yeah, it does feel strange having everything hard-coded. However, the order of the properties is the same as that on the webpage, which I assume _is_ hardcoded - if this is indeed the case then the `data_table` will always look like that in the `get_vals_from_server()` function. Furthermore, the method calls in `poll_server()` make use of the property names ("PSF27 Temp", "Cryo Temp" etc.) rather than their row index, so they would always show values corresponding to the correct properties. 

But yes, I'll have a look at how I can re-work some of this (a lot of it was written back in November 😬) and perhaps it'll be helpful if I include my web-scraping module here as well.

alexdewar:
I assume the order will be hardcoded on the server too, though it is possible that we will connect to a different EM27 (i.e. one with different sensors or a different firmware version etc.) that produces a different output. @jonemurray did say that their old tool didn't work with someone else's EM27 box, so we want to make this one more robust if possible!

Even if we *could* be certain about the format we get the values in though, I'd still add the controls dynamically, as I think it'll make for cleaner code.

finesse/gui/interferometer_monitor.py
alexdewar: The thing is that we're not actually doing anything with the values other than displaying them, so I'm not sure `pint` would help us. It would be different if we wanted to convert between unit types or do dimensional analysis etc.

finesse/gui/interferometer_monitor.py
alexdewar: Maybe this should be done in a separate setter? My point was that if we add the controls dynamically, then in theory we won't know what `prop_labels` we will have upfront and will have to wait until we first poll the server before we know how many controls to add and what their labels will be etc.

alexdewar:
PS -- I'd submit the webscraper as a separate PR after this one, because I think this stands alone.

dc2917:
> I assume the order will be hardcoded on the server too, though it is possible that we will connect to a different EM27 (i.e. one with different sensors or a different firmware version etc.) that produces a different output. @jonemurray did say that their old tool didn't work with someone else's EM27 box, so we want to make this one more robust if possible!
> 
> Even if we _could_ be certain about the format we get the values in though, I'd still add the controls dynamically, as I think it'll make for cleaner code.

Yes, that's a good point

alexdewar (changes requested):
Getting there!

There are a couple of things I've spotted though. First, you should really be using a `Decimal` type for the data values, otherwise you could get rounding errors when you convert them from the strings the server gives you.

Second, I think you could merge `_add_widgets` and `_update_widgets` into a single function to make it more robust. What I had in mind was that when you get new data back from the server, you could go through each returned value and check if there is a text label with that name already there. If so, you can update the corresponding value and if not you could add another row to the layout with the name and value. This way it would also be robust to changes between when you poll the server. Again, we don't expect this to happen but I think it'll make for cleaner code!

Finally, you could probably use some tests for all this, but I've been pretty sloppy with tests myself so I can't really hold it against you! You'd probably need to use `pytest-qt` -- I'm doing some development using this on another branch, so maybe you could have a go once that's merged? In the meantime, could you open an issue for this?

alexdewar:
PS -- Just checking: have you got your `pre-commit` hooks installed? It looks like you had to fix something up for `mypy`, but `pre-commit` should be catching this.

finesse/gui/interferometer_monitor.py
```diff
@@ -47,53 +48,61 @@ def __init__(self) -> None:
...
+        self.setLayout(self._layout)
+
+    def _display_props(self) -> None:
+        """Creates and populates the widgets to view the EM27 properties."""
+        for prop in self._data_table:
+            if prop.name not in self._prop_names:
+                # Update list of monitored properties and create corresponding label
+                self._prop_names.append(prop.name)
+                prop_label = QLabel(prop.name)
+                self._prop_labels.append(prop_label)
```
alexdewar, line 74: I don't think you're actually using `self._prop_labels` anywhere now, right? If so, you can drop it.

finesse/gui/interferometer_monitor.py
```diff
@@ -47,53 +48,61 @@ def __init__(self) -> None:
...
+                val_lineedit.setReadOnly(True)
+                val_lineedit.setAlignment(Qt.AlignmentFlag.AlignCenter)
+                self._val_lineedits.append(val_lineedit)
+                self._layout.addWidget(val_lineedit, self._num_props, 1)
+
+                # Add poll server label and icon to bottom
+                self._layout.addWidget(poll_server_label, self._num_props + 1, 0)
+                self._layout.addWidget(self._poll_light, self._num_props + 1, 1)
+
+                self._num_props += 1
```
alexdewar, line 96: You don't need to keep track of the number of properties. Just use `len(self._prop_names)`.

finesse/gui/interferometer_monitor.py
```diff
@@ -47,53 +48,61 @@ def __init__(self) -> None:
...
+                self._prop_names.append(prop.name)
+                prop_label = QLabel(prop.name)
+                self._prop_labels.append(prop_label)
+
+                # Remove poll server label and icon before adding new property widgets
+                poll_server_label = self._layout.itemAtPosition(
+                    self._num_props, 0
+                ).widget()
+                self._layout.removeWidget(poll_server_label)
+                self._layout.removeWidget(self._poll_light)
```
alexdewar, line 81: Rather than doing this dance every time you add a property, I would just have the poll-related controls in their own `QGridLayout`. You can house this layout along with `self._layout` in a parent `QVBoxLayout`.

finesse/gui/interferometer_monitor.py
```diff
@@ -47,53 +48,61 @@ def __init__(self) -> None:
...
+
+                self._layout.addWidget(prop_label, self._num_props, 0)
+
+                val_lineedit = QLineEdit()
+                val_lineedit.setText(f"{prop.value:.6f} {prop.unit}")
+                val_lineedit.setReadOnly(True)
+                val_lineedit.setAlignment(Qt.AlignmentFlag.AlignCenter)
+                self._val_lineedits.append(val_lineedit)
```
alexdewar, line 89: I would turn `self._val_lineedits` into a `dict`, so you can look it up based on its name. The values you're currently housing in `self._prop_names` could just be used as the keys.

finesse/gui/interferometer_monitor.py
```diff
@@ -47,53 +48,61 @@ def __init__(self) -> None:
...
-        """Updates the widgets with the latest values."""
-        for i, prop in enumerate(self._data_table):
-            self._prop_labels[i].setText(prop.name)  # in case order changes?
-            self._val_lineedits[i].setText(f"{prop.value:.6f} {prop.unit}")
+        self.setLayout(self._layout)
+
+    def _display_props(self) -> None:
+        """Creates and populates the widgets to view the EM27 properties."""
+        for prop in self._data_table:
+            if prop.name not in self._prop_names:
```
alexdewar, line 70: This is looking much better! Good job :smile: 

Aside from the minor comments I've made below, I think this function is a little long. I'd put most of it into a new function that returns a `QLineEdit` based on the name you give it.

It could look something like:

```py
if name in self._my_dict:
    return self._my_dict[name]

# ... otherwise create a new control as you're doing now, then return it
```

alexdewar (changes requested):
It's looking good! I think there's just one more round of (small) changes needed, I think.

finesse/gui/interferometer_monitor.py
```diff
@@ -0,0 +1,120 @@
...
+                poll_server_label = self._layout.itemAtPosition(
+                    self._num_props, 0
+                ).widget()
+                self._layout.removeWidget(poll_server_label)
+                self._layout.removeWidget(self._poll_light)
+
+                self._layout.addWidget(prop_label, self._num_props, 0)
+
+                val_lineedit = QLineEdit()
+                val_lineedit.setText(f"{prop.value:.6f} {prop.unit}")
```
dalonsoa, line 86: You can make this conversion the `__str__` method of the `EM27Property`:

```python
@dataclass
 class EM27Property:
     ...
     def __str__(self):
          return f"{self.value:.6f} {self.unit}"
```

And here just leave (I think):

```suggestion
                val_lineedit.setText(prop)
```

Or maybe:
```suggestion
                val_lineedit.setText(str(prop))
```

dalonsoa (changes requested):
Looks really good! I've just made a minor comment and agreed with Alex's comments.

finesse/gui/interferometer_monitor.py
dc2917: Yes, this is a good suggestion - thanks. I've now done this but using a new `val_str()` method rather than `__str__()`, since here I'm only concerned with the value and unit, and felt a `__str__()` method would/should contain the name too. (In fact, in the web-scraping module where the data come from, I have defined this same data class and have the `__str__()` method return "{name} = {value} {unit}" and so wanted to avoid a conflict).

finesse/gui/interferometer_monitor.py
```diff
@@ -48,55 +67,59 @@ def __init__(self) -> None:
...
+
+                prop_label, val_lineedit = self._create_prop_widgets(prop)
+                self._val_lineedits[prop.name] = val_lineedit
+
+                self._prop_wid_layout.addWidget(prop_label, num_props, 0)
+                self._prop_wid_layout.addWidget(val_lineedit, num_props, 1)
             else:
-                idx = self._prop_names.index(prop.name)
-                self._val_lineedits[idx].setText(f"{prop.value:.6f} {prop.unit}")
+                self._val_lineedits[prop.name].setText(prop.val_str())
```
alexdewar, line 122: It's good that you've encapsulated the creation of the prop widgets, but I think the adding of the widgets to the layout really belongs in there too.

And I think this can be encapsulated a bit further so that this block doesn't have to worry about finding and creating the relevant widgets at all, just setting the label. (I had to scrutinise the code to check that you were setting the value of the `QLineEdit` correctly on both branches of the `if`-statement as it is.)

What I was suggesting is that you could have a function called `_get_prop_lineedit()` or something and then this block would look like:

```py
for prop in self._data_table:
    lineedit = _get_prop_lineedit(prop.name)
    lineedit.setText(prop.val_str())
```

If the relevant `QLineEdit` already exists, then `_get_prop_lineedit()` will just return it, otherwise it will create it and then return it.

alexdewar (changes requested):
I think it just needs one last tweak... Sorry!

dalonsoa (approved):
Looks good! Just missing tests, as Codecov is pointing to.

alexdewar (approved):
LGTM! Sorry about the many rounds of review...

PR #102, opened 2023-02-28
Add mock EM27 device

alexdewar:
I've added a dummy EM27 device so we can try things out while we don't have the hardware. All it does is transition between states when the various commands are sent, sending an error if it's in an invalid state (e.g. you're trying to start a new measurement, but you're not connected to the device).

I implemented it using the `python-statemachine` package: https://github.com/fgmacedo/python-statemachine/


alexdewar:
Currently, the buttons for controlling the EM27 are always enabled, but we should only enable the ones which are valid for the current state (e.g. only show the "Start" button when the device is connected).

finesse/hardware/__init__.py
```diff
@@ -1,8 +1,14 @@
 """This module contains code for interfacing with different hardware devices."""
+import sys
+
 from pubsub import pub
 
+if "--dummy-em27" in sys.argv:
```
dc2917, line 6: Do we want equivalent statements for all other dummy devices?

alexdewar: Yes, probably. We want some way to be able to switch between using dummy devices or real ones and probably the easiest way to do that is with command-line arguments or environment variables.

Atm you *only* have the option to have a dummy stepper motor though -- even though I've written the code for the real one. My plan is to make the GUI controls for connecting to the serial devices actually do something and then we can integrate it into the main program.

finesse/hardware/dummy_opus.py
```diff
@@ -0,0 +1,168 @@
...
+        3,
+        "Status is not 'Running' or 'Finishing' although required for current command",
+    )
+    UNKNOWN_COMMAND = (4, "Unknown command")
+    WEBSITE_NOT_FOUND = (5, "Website not found")
+    NO_RESULT = (6, "No result available")
+    NOT_CONNECTED = (7, "System not connected")
+
+    def to_tuple(self) -> Optional[tuple[int, str]]:
+        """Convert to a code, message tuple or None if no error."""
```
dalonsoa, line 34:
```suggestion
        """Convert to a (code, message) tuple or None if no error."""
```

finesse/hardware/dummy_opus.py
```diff
@@ -0,0 +1,168 @@
...
+        """Timer signalling the end of a measurement."""
+        self.measure_timer.setInterval(round(measure_duration * 1000))
+        self.measure_timer.setSingleShot(True)
+        self.measure_timer.timeout.connect(self.finish_measuring)  # type: ignore
+
+    def __del__(self) -> None:
+        """Stop the background timer."""
+        self.measure_timer.stop()
+
+    def _send_response(self, type: str) -> None:
```
dalonsoa, line 110: What does `type` represent? Need docstring explaining it.

finesse/hardware/dummy_opus.py
```diff
@@ -0,0 +1,168 @@
...
+            try:
+                self.state_machine.start_measuring()
+                self.measure_timer.start()
+            except TransitionNotAllowed:
+                self.last_error = OPUSError.NOT_CONNECTED
+        elif command == "connect":
+            try:
+                self.state_machine.connect()
+            except TransitionNotAllowed:
+                self.last_error = OPUSError.NOT_IDLE
```
dalonsoa, line 156: This is a dummy device and we should not invest more time than necessary here, but I feel this could be abstracted somehow. Repeating the same block of code 4 times with just small differences seems a waste. At the very least, I'd try to use the new [`match` syntax](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching), available from Python `3.10` (are we using Python 3.10?`)

finesse/hardware/opus_interface_base.py
```diff
@@ -0,0 +1,34 @@
+"""Provides a base class for interfacing with the OPUS program."""
+import logging
+import traceback
+
+from pubsub import pub
+from PySide6.QtCore import QObject
+
+
+class OPUSInterfaceBase(QObject):
```
dalonsoa, line 9: I guess than `QObject` and `ABC`, to make this an abstract class, do not get along and that's why you are using `NotImplementedError` below? If they do get along, I would use the abstract class and set those methods as `@abstractmethod` as that would be the canonical approach.

dalonsoa (changes requested):
Very nice use of the state machine! I didn't know that one. Just a few suggestions/comments, but it looks good, otherwise.

finesse/hardware/dummy_opus.py
alexdewar: This section is a bit boilerplate-y :disappointed:, though the branches are subtly different from one another.

One option would be to have each command represented as a member function (i.e. you'd have `start()`, `stop()` etc.) and then we could just call the corresponding function for the command. Then we could probably have a decorator which returns a specified error when `TransitionNotAllowed` is raised. What do you think?

I've never used the `match` syntax, but it looks super useful! The minimum requirement for FINESSE is currently 3.9, but I was thinking about bumping it to 3.10. Do you think it'd be worth doing a separate PR for this?

finesse/hardware/opus_interface_base.py
alexdewar: Yep, that's exactly the problem. `QObject`s have their own subtle metaclass business going on. There are some workarounds but they were all pretty gnarly and didn't actually seem to work for me.

dc2917 (approved):
Looks good to me!

finesse/hardware/dummy_opus.py
alexdewar: Ping @dalonsoa ?

alexdewar: Ok cool.

What do you think about this bit?

> One option would be to have each command represented as a member function (i.e. you'd have `start()`, `stop()` etc.) and then we could just call the corresponding function for the command. Then we could probably have a decorator which returns a specified error when `TransitionNotAllowed` is raised. What do you think?

I do have some work based on top of this now, so it might be easier to merge this as is and do the refactoring in my next PR.

dalonsoa: Separating things into different functions won't be a bad idea, although the decorator part feels a bit overkilling. What about having the following function (appropriately formatted and typed):

```python
def execute(self, *args, error: OPUSError):
    try:
        for fun in args:
            fun()
    except TransitionNotAllowed:
            self.last_error = error
```

And then call it:

```python
        if command == "cancel":
            self.execute(self.state_machine.cancel, self.measure_timer.stop, error = OPUSError.NOT_RUNNING)
        elif command == "stop":
            self.execute(self.state_machine.stop, self.measure_timer.stop, error =  OPUSError.NOT_RUNNING_OR_FINISHING)
        ...
```

Or something along these lines?

alexdewar: Good idea.

I've done something similar. I realised it would probably be simpler if `OPUSStateMachine` was responsible for starting/stopping the `QTimer`, which removes one of the differences between commands. I've tried using a `dict` for each of the different error codes instead.

Let me know what you think!

dalonsoa (approved):

PR #110, opened 2023-03-07
EM27 webscraper

### LINKED ISSUE #109: Add webscraper for reading sensor values from EM27

### LINKED ISSUE #136: Initial delay in reading EM27 sensor values
The problem is that polling the server is done by a `QTimer`, but that currently means that the program waits for 2s before polling the server. Ideally we want to poll the server as soon as the program loads. This wouldn't be so much of a problem, but as the sensor values are added dynamically, it means that my window suddenly gets bigger 2s after loading, which is annoying.

@dc2917 You could probably fix this as part of #110.

dc2917:
This PR adds a module for obtaining the EM27 properties (the properties displayed in the "EM27 SOH Monitor" panel on the original GUI).

The properties are provided on a web server set up by the EM27's computer, thus this module provides web-scraping functionality to obtain the data from the corresponding web page. I have also included a snapshot of the webpage for testing functionality when access to the live server is not possible.

The module contains simple methods for opening and closing the connection to the server, reading the source HTML, and parsing this to return the data relating to the monitored properties. The expectation is that these data will be transmitted using `pubsub` to be displayed on the GUI.

Closes #109
Closes #136


alexdewar:
Hey @dc2917. I'm guessing you forgot to tag me as a reviewer for this...?

dc2917:
> Hey @dc2917. I'm guessing you forgot to tag me as a reviewer for this...?

I was going to give it one last look over before requesting reviews, but yeah I reckon it's review-ready now

alexdewar (changes requested):
I've made some minor suggestions throughout, but tbh I think this needs a bit of a rewrite.

I'd start off by separating out the loading of the page and the parsing into separate places. It will allow you to test them separately and will also mean that your `DummyEM27Scraper` can just pass in the HTML from the file rather than a URL.

For loading the page, use the `requests` package rather than Python's built-in libraries; it's much nicer. (See [the other EM27 code](https://github.com/ImperialCollegeLondon/FINESSE/blob/main/finesse/hardware/em27_opus.py) if you want an example.) You should then be able to do all the loading of the page within a single function. The loading and the parsing don't actually need to keep track of any state, so they don't have to be class methods. You might even be able to get away with dropping the class altogether.

There are a couple of extra things I'd like you to add. Firstly, could you please add code to make it controllable with `pubsub`, because that's how we've been communicating with hardware from the frontend. I'd have one message to request values and one that responds with the values retrieved. (Again, the other EM27 code has an example of this.) Ideally, we will need another message signalling that an error has occurred, but if you get stuck with this then just submit it anyway and we can talk about it. Secondly, this needs some unit tests. Again, don't spend ages bashing your head against the wall if you can't get it to work -- ask if you need help!

Finally, I'd choose a catchier name for the files (blame Bruker for the naming). How about `em27_scraper.py` or `em27_sensors.py` or something?

dalonsoa (changes requested):
I think this is a bit odd, with the webpage for the Dummy to load and commenting the `from ..config` line etc. I would remove the `if __init__` section, convert it into a test (a sort of integration or functional test) and leave the `em27_diag_autom` file as it would look in production. Otherwise, it looks rather half-baked.

And obviously, add the missing unit tests, but that can be incorporated as later stage. Open an issue about it.

finesse/hardware/em27_diag_autom.py
```diff
@@ -0,0 +1,184 @@
...
+from urllib.request import urlopen
+
+from pubsub import pub
+
+# from ..config import EM27_IP
+EM27_IP = "10.10.0.1"  # while calling as __main__ for testing
+
+
+@dataclass
+class EM27Property:
```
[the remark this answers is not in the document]
dc2917: Yeah exactly, I think it belongs here rather than in the module containing the GUI component.

finesse/hardware/em27_diag_autom.py
```diff
@@ -0,0 +1,184 @@
...
+from dataclasses import dataclass
+from decimal import Decimal
+from importlib import resources
+from urllib.error import HTTPError, URLError
+from urllib.request import urlopen
+
+from pubsub import pub
+
+# from ..config import EM27_IP
+EM27_IP = "10.10.0.1"  # while calling as __main__ for testing
```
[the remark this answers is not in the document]
dc2917: The commented-out line doesn't work if I run this module as main (I get an `ImportError`).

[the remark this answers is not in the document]
alexdewar: Does that happen even when you run it with `python -m finesse.hardware.em27_diag_autom`?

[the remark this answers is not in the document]
dc2917: Ah, that works - thanks! Learnt something new

alexdewar:
Btw you need to request a re-review from @dalonsoa because it won't let you merge this if one of the reviews is stuck at "changes requested".

tests/test_em27_scraper.py
```diff
@@ -0,0 +1,73 @@
...
+def test_get_psf27sensor_data(dev):
+    """Test EM27Scraper's get_psf27sensor_data() method."""
+    content = dev.read()
+    dev.get_psf27sensor_data(content)
+    if dev._is_read:
+        assert dev._data_table != []
+        for row in dev._data_table:
+            assert type(row) == EM27Property
+    else:
+        assert dev._data_table == []
```
alexdewar, line 50: You don't need to call `dev.read()` at all here, as a) it won't succeed and b) we're already testing it elsewhere. This test should just test `get_psf27sensor_data()`. There's no point in even checking `dev._is_read` because we know it will be `False` on any machine other than the one connected to the EM27.

The problem with this is that it only tests what happens when you pass an empty string into `get_psf27_sensor_data()`, which isn't particularly interesting! Define your own valid inputs for `get_psf27_sensor_data()` (and probably some invalid ones too) and check that the returned results are correct. Use a parametrised test for this.

tests/test_em27_scraper.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_send_data(dev, send_message_mock):
+    """Test EM27Scraper's send_data() method."""
+    try:
+        dev.send_data()
+    except Exception:
+        pass
```
alexdewar, line 58: Why are you silently ignoring all exceptions here?

tests/test_em27_scraper.py
```diff
@@ -0,0 +1,73 @@
...
+    assert not dev._is_read
+
+
+def test_error_occurred(dev, send_message_mock):
+    """Test EM27Scraper's error_occurred() method."""
+    try:
+        raise PSF27Error("Error occurred")
+    except PSF27Error as e:
+        dev._error_occurred(e)
+        send_message_mock.assert_called_with("psf27.error", message="Error occurred")
```
alexdewar, line 73: This is good! But you don't actually need to raise the exception:

```suggestion
    dev._error_occurred(PSF27Error("Error occurred"))
    send_message_mock.assert_called_with("psf27.error", message="Error occurred")
```

tests/test_em27_scraper.py
```diff
@@ -0,0 +1,73 @@
...
+    else:
+        send_message_mock.assert_called_with(
+            "psf27.data.response", data=dev._data_table
+        )
+
+    assert not dev._is_read
+
+
+def test_error_occurred(dev, send_message_mock):
+    """Test EM27Scraper's error_occurred() method."""
```
alexdewar, line 68:
```suggestion
    """Test EM27Scraper's _error_occurred() method."""
```

finesse/gui/interferometer_monitor.py
```diff
@@ -71,8 +29,7 @@ def __init__(self) -> None:
         self._data_table: list[EM27Property] = []
 
         self._poll_light = LEDIcon.create_poll_icon()
-        self._poll_light._timer.timeout.connect(self.poll_server)  # type: ignore
-        self._poll_light._timer.start(2000)
+        self._poll_light._timer.timeout.connect(self._poll_server)  # type: ignore
```
dalonsoa, line 32: Why you need to ignore this line? Why is mypy complaining?

Not sure if related, but any tribute that starts with `_` should be treated as internal to the class and not be used externally. So, you should not use `self._poll_light._timer`. If this is a legitimate use of `_timer`, then it should not be "private" and the underscore should be removed.

finesse/gui/interferometer_monitor.py
```diff
@@ -127,10 +88,22 @@ def _display_props(self) -> None:
...
             lineedit.setText(prop.val_str())
 
-    def poll_server(self) -> None:
+    def get_data_table(self, data: list[EM27Property]):
+        """Receive the data table from the server."""
+        self._data_table = data
+
+    def begin_polling(self) -> None:
+        """Initiate polling the server."""
+        self._poll_light._timer.start(2000)
```
dalonsoa, line 97: Same here.

finesse/gui/interferometer_monitor.py
```diff
@@ -127,10 +88,22 @@ def _display_props(self) -> None:
             lineedit = self._get_prop_lineedit(prop)
             lineedit.setText(prop.val_str())
 
-    def poll_server(self) -> None:
+    def get_data_table(self, data: list[EM27Property]):
+        """Receive the data table from the server."""
+        self._data_table = data
+
```
dalonsoa, line 94: Maybe call this `set_data_table`. In any case, if this is all it is doing, why not making `data_table` public (ie. remove the underscore) and just assigned `data` to it in the calling function?

finesse/gui/interferometer_monitor.py
```diff
@@ -127,10 +88,22 @@ def _display_props(self) -> None:
...
+    def begin_polling(self) -> None:
+        """Initiate polling the server."""
+        self._poll_light._timer.start(2000)
+
+    def end_polling(self) -> None:
+        """Terminate polling the server."""
+        self._poll_light._timer.stop()
```
dalonsoa, line 101: And here.

finesse/gui/interferometer_monitor.py
```diff
@@ -127,10 +88,22 @@ def _display_props(self) -> None:
...
+
+    def _poll_server(self) -> None:
         """Polls the server to obtain the latest values."""
         self._poll_light._flash()
```
dalonsoa, line 105: I missed this one - if it is going to be used outside of `poll_light`, `flash` also need to be public.

finesse/hardware/em27_scraper.py
```diff
@@ -0,0 +1,128 @@
...
+            table_header = (
+                "<TR><TH>No</TH><TH>Name</TH><TH>Description</TH>"
+                + "<TH>Status</TH><TH>Value</TH><TH>Meas. Unit</TH></TR>\n"
+            )
+            table_start = content.find(table_header)
+            try:
+                if table_start == -1:
+                    raise PSF27Error("PSF27Sensor table not found")
+            except Exception as e:
+                self._error_occurred(e)
```
dalonsoa, line 100: This is a bit odd because there is only one exception to raise here and you know the error message, so what's the need to the `try... except`? Just do:
```suggestion
            if table_start == -1:
                self._error_occurred("PSF27Sensor table not found")
```
or:
```suggestion
            if table_start == -1:
                self._error_occurred(PSF27Error("PSF27Sensor table not found"))
```

tests/test_dummy_em27_scraper.py
```diff
@@ -0,0 +1,30 @@
...
+def test_read_fail(dev):
+    """Test failure of DummyEM27Scraper's read() method."""
+    dev._url = dev._url + "chars"
+    try:
+        content = dev.read()
+        assert dev._is_read
+        assert type(content) == str
+        assert content.count("PSF27Sensor") == 1
+    except Exception:
+        assert content == ""
```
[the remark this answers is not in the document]
dalonsoa: In general, `try... except` blocks do not make sense in tests. If you want to check if an exception occurred, use [`pytest.raises`](https://stackoverflow.com/a/29855337/3778792).

dalonsoa (changes requested):
It's looking much better - well done! I've made a few comments of my own, but otherwise agreed with Alex's main comments.

finesse/hardware/em27_scraper.py
```diff
@@ -0,0 +1,128 @@
...
+        """Print a property's value and unit in required format.
+
+        Returns:
+            str: The value and unit of a property in the format consistent with
+                 the previous FINESSE GUI.
+        """
+        return f"{self.value:.6f} {self.unit}"
+
+
+class PSF27Error(Exception):
```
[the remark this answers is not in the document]
dc2917: Honestly, no idea. Maybe point spread function if it's monitoring the optics? The table on the webpage is titled "EM27 SN_93 Automation units Diagnostics" which contains a load of information, some of which comes under sections whose titles start with "PSF27", so it seems like PSF27 is just a specific part of the EM27. The specific table I search for comes under "PSF27Sensor" so that's what I went with. Even more confusing is that the GUI panel which shows these is titled "EM27 SOH Monitor". Maybe I should ask Jon.

finesse/gui/interferometer_monitor.py
dc2917: The `_timer` attribute is a `QTimer`, and `mypy` and Qt don't seem to get along.

But yes, I've not been consistent with my public and private attributes as the code has evolved -I'll fix this.

finesse/gui/led_icons.py
```diff
@@ -37,8 +37,8 @@ def __init__(self, on_img: QImage, off_img: QImage, is_on: bool = False) -> None
             self._turn_on()
         else:
             self._turn_off()
-        self._timer = QTimer()
-        self._timer.timeout.connect(self._turn_off)  # type: ignore
+        self.timer = QTimer()
+        self.timer.timeout.connect(self._turn_off)  # type: ignore
```
alexdewar, line 41: Why is `timer` public?

finesse/hardware/dummy_em27_scraper.py
```diff
@@ -22,10 +22,7 @@ def read(self) -> str:
         try:
             with open(self._url, "r") as page:
                 content = page.read()
-            self._is_read = True
             logging.info("Read PSF27Sensor table")
             return content
         except FileNotFoundError:
```
alexdewar, line 27:
```suggestion
        except FileNotFoundError as e:
```

finesse/hardware/dummy_em27_scraper.py
```diff
@@ -22,10 +22,7 @@ def read(self) -> str:
...
-            self._is_read = False
-            self._error_occurred(PSF27Error("Dummy EM27 server file not found"))
-            return ""
+            raise EM27Error(f"Dummy EM27 server file {self._url} not found")
```
alexdewar, line 28:
```suggestion
            raise EM27Error(f"Dummy EM27 server file {self._url} not found") from e
```

finesse/hardware/em27_scraper.py
```diff
@@ -58,71 +91,42 @@ def __init__(self, url: str = f"{EM27_URL}") -> None:
...
+
+            content = request.text
+            logging.info("Read PSF27Sensor table")
+            return content
+        except ConnectionError:
+            raise EM27Error(f"Error connecting to {self._url}")
+        except HTTPError:
+            raise EM27Error(f"{self._url} not found")
+        except Timeout:
+            raise EM27Error("Request timed out")
```
alexdewar, line 118:
```suggestion
            logging.info("Read PSF27Sensor table")
            return request.text
        except (ConnectionError, HTTPError, Timeout) as e:
            raise EM27Error(f"Error connecting to {self._url}") from e
```

finesse/gui/led_icons.py
dalonsoa: Because it is used outside of the class. See my comments above.

alexdewar (approved):
I've made a couple of minor suggestions and there's one query, but otherwise I think this is good to go!

finesse/gui/led_icons.py
alexdewar: Ohhh I see.

finesse/gui/interferometer_monitor.py
```diff
@@ -129,14 +89,31 @@ def _get_prop_lineedit(self, prop: EM27Property) -> QLineEdit:
...
             lineedit.setText(prop.val_str())
 
-    def poll_server(self) -> None:
+    def _set_data_table(self, data: list[EM27Property]):
+        """Receive the data table from the server.
+
+        Args:
+            data: the data received from the server
+        """
+        self._data_table = data
```
dalonsoa, line 103: Maybe mention in the docstring that this function is needed because it is use via `pubsub`, otherwise, I would make the same comment as before - why you need a method that just set the data.

finesse/hardware/__init__.py
```diff
@@ -4,14 +4,17 @@
...
     from .opus.dummy import DummyOPUSInterface as OPUSInterface
 else:
+    from .em27_scraper import EM27Scraper  # type: ignore
     from .opus.em27 import OPUSInterface  # type: ignore
 
 from .stepper_motor import create_stepper_motor_serial_manager
 from .temperature import create_temperature_controller_serial_managers
 
 opus: OPUSInterface
+scraper: EM27Scraper
```
dalonsoa, line 17: As you're defining `scraper` below, you do not need to assign a type to it here.

finesse/hardware/dummy_em27_scraper.py
```diff
@@ -0,0 +1,28 @@
...
+
+from .em27_scraper import EM27Error, EM27Scraper
+
+
+class DummyEM27Scraper(EM27Scraper):
+    """An interface for testing monitoring EM27 properties."""
+
+    def __init__(self) -> None:
+        """Create a new EM27 property monitor."""
+        dummy_em27_fp = resources.files("finesse.hardware").joinpath("diag_autom.htm")
```
dalonsoa, line 13: In our course of software engineering [we suggest that loading data in the constructor is a bad practice](https://imperialcollegelondon.github.io/grad_school_software_engineering_course/l1-05-abstractions/index.html) as there's a hidden dependency within the constructor that prevents initialising the class if not present - if should be done in a factory method. I don't have strong feelings - and I don't suggest to change this now as it will require significant updates in several places - but I just point this out for the future.

dalonsoa (approved):
I've made a couple of comments/suggestions, but otherwise is good to go.

PR #111, opened 2023-03-08
Refactor OPUS code

alexdewar:
I noticed a small issue with the current OPUS code (both the dummy version and the "real" one) after my last PR had been reviewed, but I thought it would be best to make a new PR for it as it involves a few small changes all over.

Currently, when the response to a command is received from OPUS, we send a `"opus.command.response"` message. The problem is that information about *which* command has been executed is not included, which makes it annoying to work with. I decided to use separate messages for each command of the form `f"opus.response.{command}`, which allows the users of the interface to subscribe to responses from specific commands if desired (or they can subscribe to `opus.response` to get all of them). In addition, I've turned `"status"` into a regular command like the others. In a sense it is special, as we have to request a separate URL from OPUS and it behaves slightly differently, but this isn't a distinction the frontend needs to care about.

I also added a test for the OPUS HTML parser and fixed some unrelated stepper motor tests which were giving spurious failures.

### DIFF
--- a/tests/test_em27_opus.py
+++ b/tests/test_em27_opus.py
@@ -17,7 +17,7 @@ def opus(mock) -> OPUSInterface:
 def test_request_status(opus: OPUSInterface) -> None:
     """Test OPUSInterface's request_status() method."""
     with patch.object(opus, "submit_request") as request_mock:
-        opus.request_status()
+        opus.request_command("status")
         request_mock.emit.assert_called_once_with("stat.htm", "opus.response.status")
 
 
@@ -26,7 +26,7 @@ def test_request_command(opus: OPUSInterface) -> None:
     with patch.object(opus, "submit_request") as request_mock:
         opus.request_command("hello")
         request_mock.emit.assert_called_once_with(
-            "cmd.htm?opusrshello", "opus.response.command"
+            "cmd.htm?opusrshello", "opus.response.hello"
         )
 
 
--- a/tests/test_em27_opus.py
+++ b/tests/test_em27_opus.py
@@ -1,6 +1,6 @@
 """Tests for the interface to the EM27's OPUS control program."""
 from typing import Any, Optional
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, Mock, patch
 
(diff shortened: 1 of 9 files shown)


dc2917 (approved):
Makes sense, and seems sensible!

tests/test_em27_opus.py
```diff
@@ -64,3 +65,77 @@ def test_make_request_error(requests_mock) -> None:
...
+    (
+        (http_status, status, text, errcode, errtext)
+        for http_status in (200, 403, 404)
+        for status in (None, 0, 1)
+        for text in (None, "", "status text")
+        for errcode in (None, 0, 1)
+        for errtext in (None, "", "error text")
+    ),
+)
+def test_parsing(
```
dalonsoa, line 109: Having parametric testing is really useful if you can provided the straight, expected output as an input to the test, so the logic of the test is also simple. In this case, You have a complex logic of nested conditional statement (in blocks and inline) that, in my view, defeat the purpose of having para metric testing since it does not make the test simpler. 

I would suggest splitting this test into several ones, potentially each parametrized, but much less than this one, such that the test logic involves no or minimal conditionals.

dalonsoa (changes requested):
It looks good, although Codecov highlight a couple of untested things and one of your tests is overly complicated and I would suggest simplifying it.

tests/test_em27_opus.py
alexdewar: Good point. This *felt* wrong when I was writing it actually :laughing:. I'll try breaking it up.

alexdewar:
I've split up the tests and I think it's cleaner now. Thanks! I also added coverage for a couple of lines I missed.

The lines that codecov was warning about weren't really added by me; I modified a bunch of lines which already weren't covered and didn't add new tests for them. The problem is that this code is GUI-based and currently most of the GUI stuff isn't tested. I've got another branch queued up which adds some tests using `pytest-qt` so I think it would make sense to wait until then before adding more GUI tests, if you agree?

dalonsoa (approved):
Much better!

PR #123, opened 2023-03-16
Add tests for measure script code + reorganise files

### LINKED ISSUE #83: Add tests for measure script dialogs
Currently they're entirely lacking in unit tests.

alexdewar:
This PR is primarily to add tests for the measure script code, including the GUI parts, and closes #83. For this I've used `pytest-qt`, which should be useful for any other GUI bits we want to test.

I also reorganised the file structure a little, putting `.py` files and their corresponding tests into separate subfolders. I also split up one of the larger `.py` files. No functional change is intended anywhere.

This is a little on the longer side (sorry!) but I'm hoping that as it just adds test coverage it shouldn't be too controversial.


dalonsoa (approved):
It looks good - massive and confusing with all those files moving around - but good. Definitely, it should have been two separate PR, one for testing and another for reorganising things.

alexdewar:
Yep, I'll def do that for next time. Sorry!

PR #124, opened 2023-03-16
Set size policies of EM27 SOH Monitor columns to show full sensor values

### LINKED ISSUE #122: EM27 sensor panel needs a minimum width for sensor values
Tagging @dc2917.

Currently on my machine it looks like this:

image

Ideally we want to be able to see the sensor values, whatever the size of the window. I'm not sure what the right fix is. Either we need to set a minimum width (in pixels) for the `QGridLayout`'s column or for the `QLineEdit`s themselves. Alternatively there may be an option to keep a `QLineEdit` wide enough that its text is always visible.

dc2917:
This PR contains a fix to set the column widths of the EM27 SOH Monitor panel, so that the full sensor values are shown, regardless of the window's size.

This is achieved by setting the `QSizePolicy` attributes of the `QLabel`s and `QLineEdits` in each row.

Closes #122.

### DIFF
--- a/finesse/gui/interferometer_monitor.py
+++ b/finesse/gui/interferometer_monitor.py
@@ -109,9 +109,15 @@ class EM27Monitor(QGroupBox):
         """
         if prop.name not in self._val_lineedits:
             prop_label = QLabel(prop.name)
+            prop_label.setSizePolicy(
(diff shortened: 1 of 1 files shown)


alexdewar (approved):
Works for me. Good job!

PR #126, opened 2023-03-17
Add `SerialManager` class for opening/closing devices on backend

alexdewar:
This is part of the work that I've been doing on issue #31 (i.e. making the open/close buttons on the `SerialControl` work). I realised that the changes needed to the backend are substantial enough that they probably deserve their own PR, so here it is.

Aside from adding the new class, there were some other preparatory changes I made for it:

1. Use constants for the hardware-related topic names (I had been making mistakes, so I thought this would be safer)
2. Add a base class for serial devices, which requires that all its subclasses have a `close()` method (used by `SerialManager`)

### DIFF
--- a/finesse/hardware/serial_manager.py
+++ b/finesse/hardware/serial_manager.py
@@ -1,0 +1,78 @@
+"""Common functionality for opening and closing USB serial devices."""
+from collections.abc import Callable
+
+from pubsub import pub
+from serial import Serial
+
+from ..config import DUMMY_DEVICE_PORT
+from .device_base import DeviceBase
+
+
+class SerialManager:
+    """A class for managing the opening and closing of USB serial devices.
+
+    When the object receives the "serial.{name}.open" message, a new instance of the
+    device will be created, using the serial argument. If the argument is None, a dummy
+    device will be called instead. When the "serial.{name}.open" message is received,
+    the device will be closed.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        device_ctor: Callable[[Serial], DeviceBase],
+        dummy_device_ctor: Callable[[], DeviceBase],
+    ) -> None:
+        """Create a new SerialManager object.
+
+        Args:
+            name: A unique name for the device to be used in pubsub topic names
+            device_ctor: A constructor for the real device (accepting a Serial object)
+            dummy_device_ctor: A constructor for the dummy device
+        """
+        self.name = name
+        self.device: DeviceBase
+        self.device_ctor = device_ctor
+        self.dummy_device_ctor = dummy_device_ctor
+
+        # Listen for open events for this device
+        pub.subscribe(self._open, f"serial.{name}.open")
+
+    def _open(self, port: str, baudrate: int) -> None:
+        """Open the device.
+
+        If the port is "Dummy", then a dummy device will be created.
+
+        Args:
+            port: The serial port to use
+            baudrate: The baudrate to use
+        """
+        try:
+            if port == DUMMY_DEVICE_PORT:
+                self.device = self.dummy_device_ctor()
+            else:
+                serial = Serial(port, baudrate)
+                self.device = self.device_ctor(serial)
+        except Exception as error:
+            pub.sendMessage(f"serial.{self.name}.error", error=error)
+        else:
+            # Listen for close events for this device
+            pub.subscribe(self._close, f"serial.{self.name}.close")
+
+            # For now, assume all errors are fatal so close the port
+            pub.subscribe(self._send_close_message, f"serial.{self.name}.error")
+
+            # Signal that serial device is now open
+            pub.sendMessage(f"serial.{self.name}.opened")
+
+    def _send_close_message(self, error: BaseException) -> None:
+        pub.sendMessage(f"serial.{self.name}.close")
+
+    def _close(self) -> None:
+        """Close the device.
+
+        It is ensured that devices will only be closed once.
+        """
+        pub.unsubscribe(self._close, f"serial.{self.name}.close")
+        self.device.close()
+        del self.device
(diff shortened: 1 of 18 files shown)


finesse/hardware/serial_manager.py
```diff
@@ -0,0 +1,78 @@
...
+        Args:
+            port: The serial port to use
+            baudrate: The baudrate to use
+        """
+        try:
+            if port == DUMMY_DEVICE_PORT:
+                self.device = self.dummy_device_ctor()
+            else:
+                serial = Serial(port, baudrate)
+                self.device = self.device_ctor(serial)
```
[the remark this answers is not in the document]
alexdewar: That's a good point. I'll have a crack

finesse/hardware/serial_manager.py
```diff
@@ -0,0 +1,78 @@
...
+
+        Args:
+            name: A unique name for the device to be used in pubsub topic names
+            device_ctor: A constructor for the real device (accepting a Serial object)
+            dummy_device_ctor: A constructor for the dummy device
+        """
+        self.name = name
+        self.device: DeviceBase
+        self.device_ctor = device_ctor
+        self.dummy_device_ctor = dummy_device_ctor
```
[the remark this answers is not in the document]
alexdewar: I'm actually using `ctor` as an abbreviation for constructor not controller here, but it's obviously not clear so I'll change it.

alexdewar:
Actually I've found a bug :cry: so maybe hold off on reviewing for now

alexdewar:
Closes #129.

finesse/hardware/serial_manager.py
```diff
@@ -0,0 +1,93 @@
...
+                            returning a constructed device
+        """
+        self.name = name
+        self.device: DeviceBase
+        self.device_factory = device_factory
+
+        # Listen for open events for this device
+        pub.subscribe(self._open, f"serial.{name}.open")
+
+    def _open(self, port: str, baudrate: int) -> None:
```
dc2917, line 61: Do we want to write to the log when devices are opened and closed?

alexdewar: Good point!

finesse/hardware/serial_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+
+    def create_device(port: str, baudrate: int) -> DeviceBase:
+        if port == DUMMY_DEVICE_PORT:
+            return dummy_device_factory()
+        else:
+            serial = Serial(port, baudrate)
+            return device_factory(serial)
+
+    return create_device
```
dalonsoa, line 31: This just moves the problem to a separate function, creating, on top of that, a nested function impossible to test directly. Also, for what I can tell, all those `device_factory` and `dummy_device_factory` often need `partial` functions with the arguments half populated. And, on top of that, the temperature controller, for example, has its own way of creating a serial connection (the `create` method), with the constructor accepting a `name` that is not used anywhere. 

I know we don't have time to spare, but I really think you need to go back to the drawing board and re-think the architecture of how serial devices are managed because, as it now, is really not clean. I'm happy to be involved in that discussion.

dalonsoa:
Let's be proactive. Let's say we have `DeviceBase` registering all its subclasses with some metadata:

```python
REGISTERED_DEVICES: dict = {}

class DeviceBase(ABS):

    model: str
    device_type: str

    @classmethod
    def __init_subclass__(cls, model: str, device_type: str):
	cls.model = model
	cls.device_type = device_type
        REGISTERED_DEVICES[model] = cls

    @abstractmethod
    @classmethod
    def factory(cls, port, baudrate, subtopic):
	# Open serial port IF RELEVANT for this model of device. Otherwise those inputs are ignored 
        serial = Serial(...)
	# Connect signals using the topic
	topic = f"{cls.device_type}.{subtopic}" if subtopic else cls.device_type
	# Create device with the relevant inputs.
        return cls(...)
```

This requires the individual devices to be defined as:

```python
class DummyStepperMotor(StepperMotorBase, model="Dummy", device_type=STEPPER_MOTOR_TOPIC):
```

In the relevant `__init__.py` file, we create the devices when they are open. Using the stepper motor example:

```python
_stepper_motor: StepperMotorBase

def create_steper_motor(model, port, baudrate, subtopic)
	global _stepper_motor
	_stepper_motor = REGISTERED_DEVICES[model].factory(port, baudrate, subtopic)

pub.subscribe(create_steper_motor, f"serial.{STEPPER_MOTOR_TOPIC}")
```

And finally, in the front end, probably the biggest deviation from the current implementation, we create the list of devices and serial ports based on the type of devices and the ones that are available of each type. So, there will be a "Stepper Motor" frame with 3 dropdown lists:

- Valid devices, obtained as `[name for name, value in REGISTERED_DEVICES.items() if value.device_type == STEPPER_MOTOR_TOPIC]`, and similar for the other types of devices.
- Ports, as it is now, although this could be filtered further based on extra metadata added to the class.
- Baudrates, as it is now, although this could be filtered further based on extra metadata added to the class.

When user clicks the Open button, then the following is sent:

```python
pub.sendMessage(f"serial.{STEPPER_MOTOR_TOPIC}", model=model, port=port, baudrate=baudrate, subtopic=subtopic)
```

And similar with the others. `subtopic` is actually only relevant for the temperature controllers, with a hot and cold sides. 

This approach treats real and dummy devices in exactly the same way and allow expansion in the future with new models of devices (even provided externally via plugins, if needed), as their availability is not hardcoded but registered using subclassing.

Does it make sense? Of course, proper typing and, potentially code reorganisation to keep mypy and flake8 happy might be needed.

dc2917 (approved):
This looks good to me as it is. Perhaps I'll get a sense of what Diego has mentioned when I come to using it to upgrade my DP9800 classes.

alexdewar:
@dc2917 Cool, thanks for review!

@dalonsoa I see what you mean. Yes I just moved the problem elsewhere but you noticed my ruse :wink:. I think something along the lines of what you suggested could work... I'll have a tinker and see how I get on

alexdewar:
As discussed with @dalonsoa, although it might make for a better design to refactor this, in the interests of time, let's just merge it for now and refactor later if there's time.

PR #127, opened 2023-03-20
codecov: Don't fail CI if coverage drops

alexdewar:
Currently the CI is failing on `main`, because the codecov GitHub Action fails if the coverage drops. We only want codecov to advise us rather than fail the CI, so change the settings to reflect this.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,6 @@
-ignore:
-  - "*/tests/*"
+coverage:
+  status:
+    project:
+      default:
+        # Don't fail CI if coverage drops
+        informational: true


codecov.yml
```diff
@@ -1,2 +1,6 @@
-ignore:
-  - "*/tests/*"
```
dalonsoa, line 2: Do we want to keep ignoring tests or it does not matter?

dalonsoa (approved):
Good point! Codecov should always be informational, not making things fails, otherwise it is a pain.

dc2917 (approved):

codecov.yml
alexdewar: I don't think it matters. I copied and pasted this config option when I set up codecov, but it doesn't actually seem to be affecting our coverage.

PR #135, opened 2023-03-21
DP9800 interface

### LINKED ISSUE #19: Add ability to read from DP9800 temperature logger
This task is concerned with interfacing with the following pieces of hardware:

- [DP9800: 8 channel PT100 RTD temperature logger (Omega)](https://www.omega.co.uk/pptst/DP9800.html)

Note that while this device will be used in conjunction with the temperature controller (see #18), it is not needed to set the temperature as the controller can do this by itself; its role is to allow the user to see what the current temperature is.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-7

### LINKED ISSUE #20: Plot the temperature on a graph in real time
The user should be able to view the current temperature in the GUI (see #19).

Probably the easiest way to do this is with ``matplotlib``, but we can decide this closer to the time.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-6

- [ ] Replace current scheme for updating graph (clicking on button) with one that calls the backend
- [ ] Swap placeholder function for communicating via `pubsub`

dc2917:
This PR adds a module for reading temperatures from the DP9800 device, displaying them on the GUI and updating the blackbody temperature plot.

There are separate modules for dummy and real DP9800 devices, which broadcast the temperatures via `pubsub`. The DP9800 GUI class polls the device and listens for these broadcasts, updating the GUI accordingly. Similarly, the blackbody monitor class (`BBMonitor`) listens for these broadcasts and plots the blackbody temperatures over time.

Closes #19. Closes #20


finesse/gui/temp_control.py
```diff
@@ -188,10 +220,20 @@ def _create_controls(self) -> QGridLayout:
...
+        """Display the latest Pt 100 temperatures.
+
+        Args:
+            values: the temperatures retrieved from the DP9800
+        """
+        for i in range(self._num_channels):
+            self._channels[i].setText(f"{values[i]: .2f}")
+
 
 class TC4820(QGroupBox):
```
dalonsoa, line 238:
```suggestion
class TC4820Controls(QGroupBox):
```
As above?

finesse/hardware/dp9800.py
```diff
@@ -0,0 +1,242 @@
...
+        states that there should be 8. The device shows 8 values, corresponding
+        to the values with indices 1 to 8. The first value (at index 0) is
+        therefore ignored.
+
+        Returns:
+            data: the sequence of bytes read from the device
+        """
+        num_bytes_to_read = self.serial.in_waiting
+        if num_bytes_to_read == 0:
+            data = b""
```
dalonsoa, line 140: I would return after this line and the unindent everything else rather than leaving it within an `else` block. It simplifies readability.

finesse/hardware/dp9800.py
```diff
@@ -0,0 +1,242 @@
...
+                data = self.serial.read(num_bytes_to_read)
+            except SerialException as e:
+                self._error_occurred(DP9800Error(e))
+
+            # Perform message integrity checks
+            # Check characters we know
+            if data[0] != 2:  # STX
+                self._error_occurred(
+                    DP9800Error("Start transmission character not detected")
+                )
```
dalonsoa, line 152: I think it will be more readable if each of this checks is a different function.

dalonsoa (changes requested):
I've suggested a couple of changes - the biggest one renaming the topic use by pubs throughout the code - but otherwise it looks good. It will need to be adapted to the final form of the Serial manager (or whatever replaces it).

finesse/gui/temp_control.py
```diff
@@ -1,41 +1,27 @@
...
-    """
-    time_now = datetime.now().timestamp()
-
-    # Placeholder. Later read from sensors.
-    hot_bb_temp = (60 * time_now + 50) % 70
-    cold_bb_temp = (5 * time_now + 1) % 8
-    return (time_now, hot_bb_temp, cold_bb_temp)
-
-
 class BBMonitor(QGroupBox):
```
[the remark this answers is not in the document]
dc2917: Yes this is just for the plot area. Only two of the temperatures monitored by the DP9800 are plotted.

alexdewar (changes requested):
I think the `DP9800` code would benefit from a similar restructure to the one you did for the EM27 webscraper. In particular the error handling needs a bit of work because it's currently buggy.

I'd structure the code like this:

1. Have a method that just reads a single message from the DP9800 (i.e. until the NUL terminator) and returns the bytes. It should only raise an error if the serial device fails or times out
2. Have a non-member function which checks that the message is valid, parses the temperatures and returns both a list of temperatures and `sysflag`
3. Have a convenience method, which just calls (1) and pipes the result into (2)
4. Do all the error handling in your `send_temperatures` function

For future reference, I think the changes to the frontend should have been a separate PR.

finesse/hardware/dp9800.py
```diff
@@ -0,0 +1,263 @@
...
+
+class DP9800:
+    """An interface for DP9800 temperature readers.
+
+    The manual for this device is available at:
+    https://assets.omega.com/manuals/M5210.pdf
+    """
+
+    NUM_CHANNELS = 8
+    """The number of channels on the DP9800 device."""
```
alexdewar, line 22: I don't think you need to hardcode it. Just parse however many temperatures the device gives you in a given reading.

finesse/hardware/dp9800.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        if max_attempts < 1:
+            raise ValueError("max_attempts must be at least 1")
+
+        self.serial = serial
+        self.max_attempts = max_attempts
+        self._sysflag: str = ""
+
+        logging.info(f"Opened connection to DP9800 on port {self.serial.port}")
+        pub.sendMessage("temperature_monitor.open")
```
alexdewar, line 39:
```suggestion
```

finesse/hardware/dp9800.py
```diff
@@ -0,0 +1,263 @@
...
+            F - bit 0: temperature unit: 0 = deg C, 1 = deg F
+            A - bit 1: audible button:   0 = silence, 1 = audible
+            S - bit 2: autoscan:         0 = no scan, 1 = autoscan active
+            x - bit 3: must be 0
+            L - bit 4: logging:          0 = no logging, 1 = logging active
+            x - bit 5: must be 0
+            x - bit 6: must be 0
+            T - bit 7: instrument type:  0 = TC, 1 = PT
+        """
+        if self._sysflag != "":
```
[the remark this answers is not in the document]
alexdewar: Again, you don't need to check the input argument either. If the user is daft enough to pass in an empty string, they can live with the consequences.

finesse/hardware/dp9800.py
```diff
@@ -0,0 +1,263 @@
...
+        return num_bytes_written
+
+    def send_temperatures(self) -> None:
+        """Perform the complete process of reading from the DP9800.
+
+        Writes to the DP9800 requesting a read operation.
+        Reads the raw data from the DP9800.
+        Parses the data and broadcasts the temperatures.
+        """
+        time_now = datetime.now().timestamp()
```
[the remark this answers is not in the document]
dc2917: I was thinking that this might be more relevant when writing to the log and/or csv file (which contains the time as well as all the temperatures read by the DP9800), though I've no idea how they use the data and how precise the timing needs to be (maybe they sync it up with something else?).

Either way, I was thinking it makes sense to just calculate it once and broadcast it to all listeners.

finesse/hardware/dp9800.py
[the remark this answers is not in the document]
dc2917: I'm tempted to get rid of `sysflag`-related things altogether. The `sysflag` is contained within the data returned during a read operation, but it isn't explicitly used anywhere in what we're doing.

dc2917:
I think this is ready for another review. A couple of comments:
- I've left in the `pubsub` calls related to opening and closing of the serial port. These will later be replaced by calls from the serial manager. 
- I _could_ improve modularity by using a dynamic number of `QLineEdits`, based on how many temperatures are returned by the device (similar to how the EM27 SOH Monitor panel updates from web-scraping). However, given that the DP9800 device always returns 8 values, I've left this is static and provided a default number.

finesse/gui/temp_control.py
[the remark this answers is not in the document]
alexdewar: I agree with @dalonsoa. I realise I didn't mention it before, but the name is a little obscure. My vote is for `TemperaturePlot`.

finesse/gui/temp_control.py
```diff
@@ -100,15 +89,27 @@ def _create_figure(self) -> None:
         self._canvas.draw()
 
     def _toggle_axis_visibility(self, name: str) -> None:
-        """Shows or hides BB plots."""
+        """Shows or hides BB plots.
```
alexdewar, line 92: The abbreviation BB is probably not obvious to someone who doesn't know the system well. Just spell it out.

finesse/gui/temp_control.py
```diff
@@ -144,15 +144,27 @@ def _update_figure(self, event) -> None:
 
         self._canvas.draw()
 
+    def _plot_bb_temps(self, values: list[Decimal], time: float) -> None:
```
alexdewar, line 147: Logically I feel these arguments should be the other way round. It's a bit odd to have your `y` coming after your `x`.

finesse/gui/temp_control.py
```diff
@@ -255,7 +298,7 @@ def _create_controls(self) -> QGridLayout:
         self._power_bar.setTextVisible(False)
         self._power_bar.setOrientation(Qt.Orientation.Horizontal)
         layout.addWidget(self._power_bar, 1, 1, 1, 3)
-        layout.addWidget(QLineEdit("40"), 1, 4)
+        layout.addWidget(QLineEdit(), 1, 4)
```
alexdewar, line 301: What does this `QLineEdit` do? Is it just a placeholder for now?

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+    if data.find(3) == -1:  # ETX
+        raise DP9800Error("End transmission character not detected")
+    if data[-1] != 0:  # NUL
+        raise DP9800Error("Null terminator not detected")
+
+    bcc = data[-2]
+    bcc_chars = data[1:-2]
+    byte_sum = 0
+    for byte in bcc_chars:
+        byte_sum ^= byte
```
alexdewar, line 39: I'd have a separate function for calculating the checksum. It'd be clearer and make testing easier.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+    representing each of the temperatures, and finally into floats.
+
+    Returns:
+        vals: A list of Decimals containing the temperature values recorded
+              by the DP9800 device.
+        sysflag: string representation of the system flag bitmask
+    """
+    try:
+        data_ascii = data.decode("ascii")
+    except Exception as e:
```
alexdewar, line 58:
```suggestion
    except UnicodeDecodeError as e:
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+
+    vals_begin = 3
+    vals_end = 74
+    etx_index = data.find(b"\x03")
+
+    vals = [Decimal(val) for val in data_ascii[vals_begin:vals_end].split()]
+
+    sysflag = bin(int(data_ascii[vals_end:etx_index], 16))
+
+    return (vals[1:], sysflag[2:])
```
alexdewar, line 69: Small stylistic suggestion:

```suggestion
    return vals[1:], sysflag[2:]
```

I think you need a comment to say *why* you're discarding some of these values.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Raises:
+            SerialException: Error communicating with device
+        """
+        try:
+            self.serial.close()
+            pub.sendMessage("temperature_monitor.close")
+            logging.info("Closed connection to DP9800")
+        except SerialException as e:
+            raise DP9800Error(e)
```
alexdewar, line 106: As you're only expecting to get exceptions on one line, I'd do this:

```suggestion
        try:
            self.serial.close()
        except SerialException as e:
            raise DP9800Error(e)
        else:
            pub.sendMessage("temperature_monitor.close")
            logging.info("Closed connection to DP9800")
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+            instr_type = ["TC", "PT"][int(sysflag[0])]
+            logging_state = ["no logging", "logging active"][int(sysflag[3])]
+            scanning_state = ["no scan", "autoscan active"][int(sysflag[5])]
+            audible_state = ["silence", "audible"][int(sysflag[6])]
+            temp_unit = ["deg C", "deg F"][int(sysflag[7])]
+            print(f"Instrument type: {instr_type}")
+            print(f"Logging: {logging_state}")
+            print(f"Autoscan: {scanning_state}")
+            print(f"Audible button: {audible_state}")
+            print(f"Temperature unit: {temp_unit}")
```
alexdewar, line 135: It's up to the user to provide sane input so you don't need to check it:

```suggestion
        instr_type = ["TC", "PT"][int(sysflag[0])]
        logging_state = ["no logging", "logging active"][int(sysflag[3])]
        scanning_state = ["no scan", "autoscan active"][int(sysflag[5])]
        audible_state = ["silence", "audible"][int(sysflag[6])]
        temp_unit = ["deg C", "deg F"][int(sysflag[7])]
        print(f"Instrument type: {instr_type}")
        print(f"Logging: {logging_state}")
        print(f"Autoscan: {scanning_state}")
        print(f"Audible button: {audible_state}")
        print(f"Temperature unit: {temp_unit}")
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Returns:
+            data: the sequence of bytes read from the device
+
+        Raises:
+            DP9800Error: Malformed message received from device
+        """
+        if self.serial.in_waiting == 0:
+            raise DP9800Error("No data waiting to be read")
+
```
alexdewar, line 171: If you follow my suggestion below you won't need this.

```suggestion
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+        try:
+            data = self.serial.read_until(b"\x00")
+        except SerialTimeoutException:
+            raise DP9800Error("Read request timed out")
+
+        return data
```
alexdewar, line 177: Unfortunately, `read_until()` *doesn't* raise an exception if it times out; it just returns fewer bytes than expected (usually zero bytes): https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.read_until

I think you want to set the `size` argument to 4 (or something close). (I'm thinking you need 3 bytes for the header and one for the terminator.) Then you need to check whether `len(data)` is big enough and raise an exception if not (which I'd do in this function rather than your parsing one).

`read_until()` *might* throw a `SerialException`, though, so that's the one you need to check for.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+        Parses the data and broadcasts the temperatures.
+        """
+        time_now = datetime.now().timestamp()
+        try:
+            self.request_read()
+            data = self.read()
+            check_data(data)
+            temperatures, _ = parse_data(data)
+            pub.sendMessage(
+                "temperature_monitor.data.response", values=temperatures, time=time_now
```
alexdewar, line 211: I feel like the argument should also be called `temperatures` rather than `values`.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+        except Exception as e:
+            self._error_occurred(e)
```
alexdewar, line 214: As above, I'd be wary about wrapping code in a `try`/`except` block when it doesn't need to be. The danger here is that if an exception occurs in one of the pubsub listeners, it will percolate up the stack and `self._error_occurred` will be called, which will make debugging annoying.

```suggestion
        except Exception as e:
            self._error_occurred(e)
        else:
            pub.sendMessage(
                "temperature_monitor.data.response", values=temperatures, time=time_now
            )
```

finesse/hardware/temperature/dummy_dp9800.py
```diff
@@ -0,0 +1,50 @@
+"""This module provides an interface to dummy DP9800 temperature readers."""
+import logging
+
+from pubsub import pub
+
+from .dp9800 import DP9800, DP9800Error
+
+
+class DummyDP9800(DP9800):
```
alexdewar, line 9: As I said before, I'd rewrite this class so that it doesn't use the `DP9800` parser at all. These various bits of state make it unnecessarily complicated and bugprone.

Just have a member function called `read_temperatures` for both `DP9800` and `DummyDP9800` and the dummy version can just return static data (or random data).

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,219 @@
...
+    bcc_chars = data[1:-2]
+    byte_sum = 0
+    for byte in bcc_chars:
+        byte_sum ^= byte
+
+    if byte_sum != bcc:
+        raise DP9800Error("BCC check failed")
+
+
+def parse_data(data: bytes) -> tuple[list[Decimal], str]:
```
alexdewar, line 45: I'd call  `check_data()` directly from this function. I don't think it makes sense to have checking the data be a separate step that the user has to remember to do.

alexdewar (changes requested):
Getting there! I think some of my suggestions from last time got lost along the way.

I think you should replace the `DummyDP9800` class with a more generic `DummyTemperatureMonitor` one. Make an abstract base class with one abstract method called `read_temperatures()` or something and then `DummyTemperatureMonitor` can just have a variant that returns static or random data. Then subscribing to and sending pubsub messages can be dealt with in the base class too so you're not repeating yourself.

alexdewar:
Another thing: codecov is complaining that the test coverage has dropped and is failing the CI as a result. I've disabled this in #137, so we'll want to merge that first so that this branch can get the fix too.

dalonsoa (changes requested):
I generally agree with all of Alex's comments, including that some comments from the previous review were lost along the way.

alexdewar:
@dc2917 One thing you can do if you have lots of comments is resolve the comment threads as you fix the issues so you don't lose track.

alexdewar:
@dc2917 My PR has now been merged, so if you merge `main` into this branch that should fix the codecov CI failure.

finesse/gui/temp_control.py
dc2917: This is the one I was referring to in our meeting yesterday. I _think_ it just shows the power level numerically, but I'm not certain as on the screenshot of the old GUI it shows 40 on both the hot and cold TC4820s, while the power level `QProgressBar`s are different (~40 and 0). It's updated in the TC4820 branch I'm yet to push - not sure why I bothered changing this here.

finesse/gui/temp_control.py
```diff
@@ -61,6 +49,10 @@ def _create_controls(self) -> QGridLayout:
             partial(self._toggle_axis_visibility, name="cold")
         )
         self._create_figure()
+        self._canvas.setMinimumSize(QSize(640, 120))
+        self._canvas.setSizePolicy(
+            QSizePolicy.MinimumExpanding, QSizePolicy.MinimumExpanding  # type: ignore
```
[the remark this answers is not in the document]
dc2917: `error: "Type[QSizePolicy]" has no attribute "MinimumExpanding"  [attr-defined]`

finesse/gui/temp_control.py
alexdewar: I seeeee. I'd maybe just update it based on the power level then. We can always ask Jon about it when we show it to him.

finesse/gui/temp_control.py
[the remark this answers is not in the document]
alexdewar: Does it work if you use `QSizePolicy.Policy.MinimumExpanding` instead?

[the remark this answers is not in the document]
dc2917: Yep, that works - thanks! Will also fix everywhere else I've set size policies.

finesse/hardware/temperature/dp9800.py
alexdewar: I don't think we should be hardcoding that though. I'd just make `size` be the minimum required for the code that comes after to work, i.e. the size of the header plus checksum plus terminator. Otherwise if someone wanted to change the device for one that returns, say, six or ten values instead, they'd have to figure out that they also had to change this one random magic number.

dalonsoa: @alexdewar , do not think that much on possible use cases. This class is specific for DP9800, which returns 9 temperature values no matter what, so let's write code design to work with that. If this were a base class, or a generic interface, it would be different, but this is bespoken to a particular device.

dalonsoa: So, the dummy should be generic, but not this one.

alexdewar: Yeah, I don't think I expressed myself very well there. I'm pretty sure that device will never be replaced.

Originally the way @dc2917 was extracting the temperatures assumed there were nine channels (which is fine because there are!) but I suggested it would be a bit cleaner to just `split()` it rather than hardcoding a bunch of offsets etc. So this suggestion about the number of bytes is in a similar vein.

finesse/gui/temp_control.py
```diff
@@ -161,6 +173,25 @@ def __init__(self, num_channels: int) -> None:
...
 
+        self._begin_polling()
+
+        pub.subscribe(self._begin_polling, "temperature_monitor.open")
+        pub.subscribe(self._end_polling, "temperature_monitor.close")
+        pub.subscribe(self._update_pt100s, "temperature_monitor.data.response")
+
+    def _begin_polling(self) -> None:
+        """Initiate polling the DP9800 device."""
+        self._poll_light.timer.start(2000)
```
dalonsoa, line 184: Could we make this 2000 a class attribute,  with the appropriate docstring, so it is easy to identify and change, if needed? Otherwise, it just pops up out of the blue with no clear meaning.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+    if bcc != data[-2]:
+        raise DP9800Error("BCC check failed")
+
+    try:
+        data_ascii = data.decode("ascii")
+    except UnicodeDecodeError as e:
+        raise DP9800Error(e)
+
+    vals_begin = 3
+    vals_end = 74  # assuming 9 temperatures
```
dalonsoa, line 78: Why 3 and 74? Need a bit of an explanation - maybe in the docstring - of the meaning of these numbers.

dalonsoa (approved):
Looks good! I've just asked for a couple of clarifications.

finesse/gui/temp_control.py
alexdewar: I'd call `self._poll_light_timer.setInterval()` in the constructor and let the user pass the poll interval as an argument (with a sensible default value).

finesse/gui/temp_control.py
```diff
@@ -161,6 +173,25 @@ def __init__(self, num_channels: int) -> None:
         layout = self._create_controls()
         self.setLayout(layout)
 
+        self._begin_polling()
```
alexdewar, line 176: As with the EM27 scraper, it might be worth polling the device once before starting the timer so the GUI is populated with values as soon as it opens.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Args:
+        data: the message to check
+
+    Raises:
+        DP9800Error: Malformed message received from device
+    """
+    header_width = 2
+    value_width = 8
+    num_values = 9
```
alexdewar, line 27: As I said before, I don't think we should be hardcoding the number of channels in various places. Just check that the header + terminator are there. If the device ever *did* decide to give us fewer than 9 channels' worth of data (it won't) then we can just pass however many values we get along to the `pubsub` listeners, which is fine.

finesse/hardware/temperature/dp9800.py
alexdewar: I'd write `vals_end` as a negative index, i.e. so the position is relative to the end of the string. It'll be more obvious that way to whoever's reading the code.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+            serial: Serial device
+        """
+        super().__init__("DP9800")
+        self.serial = serial
+
+    def close(self) -> None:
+        """Close the connection to the device.
+
+        Raises:
+            SerialException: Error communicating with device
```
alexdewar, line 113:
```suggestion
            DP9800Error: Error communicating with device
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+            dev_settings: dictionary containing the current device settings
+        """
+        dev_settings = {
+            "instrument_type": ["TC", "PT"][int(sysflag[0])],
+            "logging_state": ["no logging", "logging active"][int(sysflag[3])],
+            "scanning_state": ["no scan", "autoscan active"][int(sysflag[5])],
+            "audible_state": ["silence", "audible"][int(sysflag[6])],
+            "temperature_unit": ["deg C", "deg F"][int(sysflag[7])],
+        }
+        return dev_settings
```
alexdewar, line 150: Much better! Small suggestion:
```suggestion
        Returns:
            dictionary containing the current device settings
        """
        return {
            "instrument_type": ["TC", "PT"][int(sysflag[0])],
            "logging_state": ["no logging", "logging active"][int(sysflag[3])],
            "scanning_state": ["no scan", "autoscan active"][int(sysflag[5])],
            "audible_state": ["silence", "audible"][int(sysflag[6])],
            "temperature_unit": ["deg C", "deg F"][int(sysflag[7])],
        }
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+        therefore ignored.
+
+        Returns:
+            data: the sequence of bytes read from the device
+
+        Raises:
+            DP9800Error: Malformed message received from device
+        """
+        # require at least 4 bytes else checks will fail
+        min_length = 4
```
alexdewar, line 185: I think this is the same minimum length you should use in `check_data()`.

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+        try:
+            data = self.serial.read_until(b"\x00", size=min_length)
+        except SerialException as e:
+            raise DP9800Error(e)
+
+        if len(data) < min_length:
+            raise DP9800Error("Insufficient data read from device")
+        else:
+            return data
```
alexdewar, line 194: You don't need `else` here:

```suggestion
        return data
```

finesse/hardware/temperature/dp9800.py
```diff
@@ -0,0 +1,238 @@
...
+        """Perform the complete process of reading from the DP9800.
+
+        Writes to the DP9800 requesting a read operation.
+        Reads the raw data from the DP9800.
+        Parses the data and broadcasts the temperatures.
+        """
+        time_now = datetime.now().timestamp()
+        try:
+            self.request_read()
+            data = self.read()
```
alexdewar, line 224: Generally with timestamps I'd get the time just after reading rather than before, in case the request blocks for a while. (Probably not likely here, but still.)

```suggestion
        try:
            self.request_read()
            data = self.read()
            time_now = datetime.now().timestamp()
```

finesse/hardware/temperature/dummy_temperature_monitor.py
```diff
@@ -0,0 +1,43 @@
...
+        temperatures = [
+            Decimal(19 + noise),
+            Decimal(17 + noise),
+            Decimal(26 + noise),
+            Decimal(850.00),
+            Decimal(24 + noise),
+            Decimal(68 + noise),
+            Decimal(69 + noise),
+            Decimal(24 + noise),
+        ]
```
alexdewar, line 38: A couple of things:

1. You don't need to worry about setting a random seed (I used a fixed seed on purpose)
2. This currently uses the same random number for each channel

How about something like this:

```suggestion
        _BASE_TEMPS = (19, 17, 26, 850, 24, 68, 69)
        rand = NoiseProducer()

        time_now = datetime.now().timestamp()
        temperatures = [rand() + temp for temp in _BASE_TEMPS]
```

finesse/hardware/temperature/temperature_monitor_base.py
alexdewar, line 1: Good!

alexdewar (changes requested):
Neaaaarly there. Just one more round of changes needed I think.

finesse/hardware/temperature/dummy_temperature_monitor.py
alexdewar: Ahh my bad. So it will. In that case I'd make the `NoiseProducer` an attribute of the class. Setting up random number generators is a little slow so you don't want to do it more often than you have to. It won't be a bottleneck here, but it's a good habit to get into.

finesse/hardware/__init__.py
```diff
@@ -18,16 +18,14 @@
...
-dp9800: DP9800
 opus: OPUSInterface
 
 
 def _init_hardware():
     global opus, dp9800
 
-    opus = OPUSInterface()
-
     dp9800 = DP9800()
```
alexdewar, line 27: This will raise an error unless you pass the `--dummy-dp9800` argument, because it expects a `Serial` as an argument. (`mypy` would ordinarily catch this, but it's missing it because of the `type: ignore` shenanigans above.) I'd just drop the `if "--dummy-dp9800"` business above and unconditionally use the `DummyTemperatureMonitor` for now. We'll obvs replace this bit anyway once when we tackle #141.

alexdewar (changes requested):
One small change, then I think we're good to go!

PR #137, opened 2023-03-23
Send pubsub messages for stepper motor errors

### LINKED ISSUE #112: Communicate stepper motor errors via `pubsub` messages
While there is error handling for `ST10Controller`, at present it just consists of raising exceptions, meaning that the frontend cannot be informed of errors.

We could send a `"stepper.error"` message in this case.

alexdewar:
Closes #112.

### DIFF
--- a/finesse/hardware/stepper_motor/stepper_motor_base.py
+++ b/finesse/hardware/stepper_motor/stepper_motor_base.py
@@ -16,10 +16,13 @@ class StepperMotorBase(DeviceBase):
 
         Subscribe to stepper.move messages.
         """
-        pub.subscribe(self.move_to, f"serial.{STEPPER_MOTOR_TOPIC}.move.begin")
-        pub.subscribe(self.stop_moving, f"serial.{STEPPER_MOTOR_TOPIC}.stop")
         pub.subscribe(
-            self.notify_on_stopped, f"serial.{STEPPER_MOTOR_TOPIC}.notify_on_stopped"
+            self._move_to,
+            f"serial.{STEPPER_MOTOR_TOPIC}.move.begin",
+        )
+        pub.subscribe(self._stop_moving, f"serial.{STEPPER_MOTOR_TOPIC}.stop")
+        pub.subscribe(
+            self._notify_on_stopped, f"serial.{STEPPER_MOTOR_TOPIC}.notify_on_stopped"
         )
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


dalonsoa:
The informational flag should work, but given that it does not, you can try to set a threshold of how much coverage can go down before considering that a failure. https://docs.codecov.com/docs/commit-status

finesse/hardware/stepper_motor/stepper_motor_base.py
```diff
@@ -95,3 +103,21 @@ def move_to(self, target: Union[float, str]) -> None:
             raise ValueError("Angle must be between 0° and 270°")
 
         self.step = round(self.steps_per_rotation * target / 360.0)
+
+    def _move_to(self, target: Union[float, str]) -> None:
+        try:
+            self.move_to(target)
+        except Exception as error:
+            self.send_error_message(error)
```
dalonsoa, line 111: This is a counter intuitive construction. As `_move_to` is called by `pubsub` - which responds to an external caller -, that should be the public method, and the one called inside the `try` block would be the private method.

finesse/hardware/stepper_motor/stepper_motor_base.py
```diff
@@ -95,3 +103,21 @@ def move_to(self, target: Union[float, str]) -> None:
...
+
+    def _stop_moving(self) -> None:
+        try:
+            self.stop_moving()
+        except Exception as error:
+            self.send_error_message(error)
```
dalonsoa, line 117: And here.

finesse/hardware/stepper_motor/stepper_motor_base.py
```diff
@@ -95,3 +103,21 @@ def move_to(self, target: Union[float, str]) -> None:
...
+
+    def _notify_on_stopped(self) -> None:
+        try:
+            self.notify_on_stopped()
+        except Exception as error:
+            self.send_error_message(error)
```
dalonsoa, line 123: And here.

dalonsoa (changes requested):
There's a mixup of what public and private methods are, but otherwise looks good!

dc2917 (approved):
Looks good to me

finesse/hardware/stepper_motor/stepper_motor_base.py
alexdewar: The reason I did it this way is because I wanted `move_to` to be the interface to the class for "normal" use (i.e. if you just want to make an object and call a method on it). The private version of the method just adds error handling and you'd only ever really want it to be called by pubsub, because otherwise you won't see raised exceptions in the usual way.

Also it seemed more intuitive to me to have the child classes override public methods rather than private ones, which seems a little odd.

I'm not totally wedded to this though. What do you think?

dalonsoa (approved):
OK. I did not have strong feelings on the public/private method thing. 

Having said that, you keep considering use cases - like calling this object method's directly or providing multiple inputs to some dummy devices to customise them - that are not really use cases with the architecture we are designing for the software. If those really become a thing, ever, the changes to the flow of the program would have been so large that it does not really matter because it will be a different program altogether. 

It is important to thing on the future, but sometimes that leads to cumbersome coding. Let's be flexible, but focus on solving the problem we have here and now in an elegant way, and not on "What if...".

Issue #133, opened 2023-03-20
Remember user's choices for device ports and baudrates


alexdewar:
This feature is provided by the existing software. It would be annoying and error-prone to have the user re-enter these details every time they run the program.

PR #139, opened 2023-03-23
Connect up `SerialControl` to backend

### LINKED ISSUE #131: `SerialControl`: Make open/close buttons send appropriate commands to backend
Currently the buttons are non-functional.

alexdewar:
This PR allows the user to open and close connections to serial devices by clicking buttons in the `SerialControl`. The user can choose the desired COM port and baudrate for the device. The list of available COM ports is obtained by querying `PySerial`, whereas the list of available baudrates is hardcoded. The program remembers the last selected COM port and baudrate for each device. If there is no stored setting available for the baudrate, each device has its own preferred baudrate which will be selected by default.

Closes #131, #133.

### DIFF
--- a/finesse/gui/serial_view.py
+++ b/finesse/gui/serial_view.py
@@ -81,11 +81,10 @@ class DeviceControls:
             avail_ports: Possible serial ports to choose from
             avail_baudrates: Possible baudrates to choose from
         """
-        self.name = device.name
-        self.label = device.label
+        self.device = device
 
         # Add a label showing the device name
-        layout.addWidget(QLabel(self.label), row, 0)
+        layout.addWidget(QLabel(self.device.label), row, 0)
 
         self.ports = QComboBox()
         """The available serial ports for the device."""
@@ -95,7 +94,10 @@ class DeviceControls:
         # Try to load the port from settings
         if avail_ports:
             saved_port = cast(
-                str, settings.value(f"{_KEY_PREFIX}/{self.name}/port", avail_ports[0])
+                str,
+                settings.value(
+                    f"{_KEY_PREFIX}/{self.device.name}/port", avail_ports[0]
+                ),
             )
             self.ports.setCurrentText(saved_port)
 
@@ -109,7 +111,8 @@ class DeviceControls:
             saved_baudrate = cast(
                 int,
                 settings.value(
-                    f"{_KEY_PREFIX}/{self.name}/baudrate", device.default_baudrate
+                    f"{_KEY_PREFIX}/{self.device.name}/baudrate",
+                    device.default_baudrate,
                 ),
             )
             self.baudrates.setCurrentText(str(saved_baudrate))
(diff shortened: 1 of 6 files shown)


finesse/gui/serial_view.py
```diff
@@ -1,138 +1,206 @@
...
-            _open_close_btn.setCheckable(True)
-            _open_close_btn.released.connect(  # type: ignore
-                partial(
-                    self.on_open_close_clicked, device_name=name, button=_open_close_btn
-                )
-            )
-            _open_close_btn.setChecked(False)
-            layout.addWidget(_open_close_btn, i, 3)
+        self.name = device.name
+        self.label = device.label
```
dalonsoa, line 85: Why not storing this as `self.device = device`? It seems a waste to define a data class to put together the information concerning a device and then unpack it almost immediately.

finesse/gui/serial_view.py
```diff
@@ -1,138 +1,206 @@
...
-        logging.info(
-            f"{device_name} - old baud_rate: {self._devices[device_name]['baud_rate']} "
-            f"- new baud_rate: {new_baud_rate}"
-        )
-        self._devices[device_name]["port"] = new_baud_rate
+        # Try to load the port from settings
+        if avail_ports:
+            saved_port = cast(
+                str, settings.value(f"{_KEY_PREFIX}/{self.name}/port", avail_ports[0])
+            )
```
dalonsoa, line 99: This an odd construction. Why are you using `_KEY_PREFIX` is it refers always to `serial`? You are indeed using `serial` directly in the `pubsub` commands.
```suggestion
            saved_port = cast(
                str, settings.value(f"serial/{self.name}/port", avail_ports[0])
            )
```

finesse/gui/serial_view.py
```diff
@@ -1,138 +1,206 @@
...
+        ).exec()
+
+    def _open_device(self) -> None:
+        """Open the specified serial device."""
+        port = self.ports.currentText()
+        baudrate = int(self.baudrates.currentText())
+
+        # Remember these settings for the next time program is run
+        settings.setValue(f"{_KEY_PREFIX}/{self.name}/port", port)
+        settings.setValue(f"{_KEY_PREFIX}/{self.name}/baudrate", baudrate)
```
dalonsoa, line 150:
```suggestion
        settings.setValue(f"serial/{self.name}/port", port)
        settings.setValue(f"serial/{self.name}/baudrate", baudrate)
```
Also, I would set these values after trying to connect to the device and being successful. If there's any error, the new settings should not be saved in the settings.

finesse/settings.py
```diff
@@ -0,0 +1,7 @@
+"""A module with a single settings object for the program settings."""
+from PySide6.QtCore import QSettings
+
+from .config import APP_AUTHOR, APP_NAME
+
+settings = QSettings(APP_AUTHOR, APP_NAME)
```
dalonsoa, line 6: Nice!

dalonsoa (changes requested):
Looks nice! I've just left a couple of comments/suggestions.

finesse/settings.py
alexdewar: Yeah, Qt is the best sometimes!

finesse/gui/serial_view.py
alexdewar: Both good points!

dc2917 (approved):
All looks good and seems sensible to me!

dalonsoa (approved):
A couple of suggestions, but otherwise good to go.

finesse/gui/serial_view.py
```diff
@@ -1,138 +1,211 @@
...
+@dataclass
+class Device:
+    """The parameters to use for a particular device."""
+
+    label: str
+    """A human-readable label for the device"""
+    name: str
+    """The name of the device as used in pubsub topic"""
+    default_baudrate: int
+    """The default baudrate to select for the device"""
```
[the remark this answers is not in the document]
alexdewar: Whoops, I missed this. I'll do it later.

PR #149, opened 2023-03-29
Enable/disable panels on device connect/disconnect

alexdewar:
This doesn't actually relate to a specific issue, but I think it's a usability improvement and I had the code for this lying around anyway.

In future we should also probably enable/disable more of the controls (e.g. the various OPUS buttons) depending on the program state to indicate to the user what controls they can use at any given time.

### DIFF
--- a/finesse/gui/serial_device_panel.py
+++ b/finesse/gui/serial_device_panel.py
@@ -1,0 +1,53 @@
+"""Contains a panel which enables/disables child controls when device opens/closes."""
+from typing import Any
+
+from pubsub import pub
+from PySide6.QtWidgets import QGroupBox, QWidget
+
+
+class SerialDevicePanel(QGroupBox):
+    """A QGroupBox which enables/disables child controls when device opens/closes."""
+
+    def __init_subclass__(cls, **kwargs):
+        """Disable controls after construction."""
+        super().__init_subclass__(**kwargs)
+
+        def init_decorator(previous_init):
+            def new_init(self, *args, **kwargs):
+                previous_init(self, *args, **kwargs)
+                self.disable_controls()
+
+            return new_init
+
+        cls.__init__ = init_decorator(cls.__init__)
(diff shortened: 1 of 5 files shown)


finesse/gui/serial_device_panel.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Args:
+            enabled: Whether to enable or disable the controls
+        """
+        for widget in self.findChildren(QWidget):
+            widget.setEnabled(enabled)
+
+    def disable_controls(self) -> None:
+        """Disable the controls in this panel."""
+        self.enable_controls(False)
```
dalonsoa, line 53: Why are these two public methods and why `disable_controls` is not just `_on_device_closed`, which is when it is used?

dalonsoa (approved):
Looks useful! Just a comment on the methods.

finesse/gui/serial_device_panel.py
alexdewar: They could be private, I guess. They're only called by `pubsub`.

I had it in mind that it was a better idea to name functions according to what they do, rather than when they're called, so that was the logic here, but I've definitely not been consistent with that throughout the project!

dalonsoa: IF that's the logic, then there should be no `_on_device_open`. Having said that, it is common to name callback functions based on the when they are called, so I don't think that's a bad approach, either. I've the feeling you've mix both approaches here. I would suggest to rename the functions to one of these two sets (names are in the order the functions are written):

- enable_controls, toggle_enable, disable_controls
- _on_device_open, toggle_enable, _on_device_close

I prefer the later set.

alexdewar: Good shout. It is a bit inconsistent atm

dc2917 (approved):
Looks good to me!

PR #154, opened 2023-03-30
Fix log spam from `EM27Scraper`

alexdewar:
Currently the `EM27Scraper` logs every time it loads data, which is every other second or so. This means that if you're running the program your console log will very quickly be filled up with these messages obscuring whatever other output you might be looking for.

My bad for not clocking this during the code review.


alexdewar:
Nw. I don't actually think it's a bad thing to log, it's just a bit much to have it on by default. I tried reducing the logging level to `debug` but it seems we're printing these to console by default (which we might want to change), so removing them seemed like the easiest thing to do for now.

PR #159, opened 2023-04-03
Bump Python version to 3.10

### LINKED ISSUE #107: Bump minimum Python version to 3.10
There a few syntactic niceties it would be helpful to be able to use (such as the `match` syntax) so let's increase the minimum Python version.

alexdewar:
Closes #107.

This also fixes an issue meaning that GitHub Actions can't update the pre-commit hooks because it is using the wrong version of Python.


dc2917:
I approved this PR then realised the readme needs updating too.

alexdewar:
Good spot!

PR #166, opened 2023-04-03
Fix: Wait for ST10 to stop moving in constructor

### LINKED ISSUE #162: Stop user from interacting with ST10 before calibration is complete
Otherwise the mirror's calibration will be wrong, which we don't want!

alexdewar:
This PR just adds a call to `wait_to_stop_moving()` to the homing code, so that it will
block until the motor has stopped moving. This prevents the user from interacting with
the device while calibration in progress.

It's a bit gross that we're blocking the main thread while this happens, but this will
do as a stopgap.

Closes #162.

### DIFF
--- a/finesse/hardware/stepper_motor/st10_controller.py
+++ b/finesse/hardware/stepper_motor/st10_controller.py
@@ -276,7 +276,7 @@ class ST10Controller(StepperMotorBase):
         self._write_check("SP0")
 
         # Make sure the motor has stopped
-        self.wait_until_stopped(20.0)
+        self.wait_until_stopped(3.0)
 
     def _relative_move(self, steps: int) -> None:
         """Move the stepper motor to the specified relative position.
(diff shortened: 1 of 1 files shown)


finesse/hardware/stepper_motor/st10_controller.py
```diff
@@ -275,6 +275,9 @@ def _home_and_reset(self) -> None:
         # Tell the controller that this is step 0 ("set variable SP to 0")
         self._write_check("SP0")
 
+        # Make sure the motor has stopped
+        self.wait_until_stopped(20.0)
```
dc2917, line 279: Is this 20 seconds? This seems like quite a long time, but you'd know better than me! I'm just wondering if it's possible to calculate the maximum time this could take based on how fast it rotates and the maximum possible angle (270 degrees?)

alexdewar: Yup! It is a conservative value, but it's not likely that we will ever actually wait that long, unless something v specific goes wrong. I'm not sure how fast it actually moves off the top of my head although I suspect 10s might be long enough.

Do you have a better idea @jonmurray?

alexdewar:
@jonemurray and I settled on a figure of 3 s for the delay in the end.

dc2917 (approved):

PR #168, opened 2023-04-03
Improvements to blackbody temperature figure

### LINKED ISSUE #160: Change DP9800 plot to cover last 15 mins
Currently it only covers the last 10 points.

We should probably change the API to allow passing in a desired duration as an argument to the constructor.

### LINKED ISSUE #161: Use smaller points on the plot
The crosses are rather large. Maybe dots would be better?

### LINKED ISSUE #167: matplotlib warning in TemperaturePlot
I get warnings like this whenever I run the program:

```
/home/alex/code/FINESSE/finesse/gui/temp_control.py:145: UserWarning: FixedFormatter should only be used together with FixedLocator
  self._ax["hot"].set_xticklabels(xticklabels)
```

dc2917:
This PR contains some improvements to the figure showing the hot and cold blackbody temperatures, following feedback from Jon as well as an improvement I thought necessary.

The plot now shows 15 minutes worth of data, as per Jon's request, rather than the last 10 data points. The plot also now just shows a line rather than joined-up markers. The re-scaling of the y axes has also been improved, and is performed when data visibility is toggled in addition to when new points are added.

I have also cleaned up the setting of the x axis tick labels, and set a maximum number of x ticks to display.

Closes #160. Closes #161. Closes #167.


finesse/gui/temp_control.py
```diff
@@ -68,7 +69,9 @@ def _create_figure(self) -> None:
         self._ax = {"hot": ax}
         self._canvas = FigureCanvasQTAgg(self._figure)
 
-        self._figure_num_pts = 10
+        poll_interval = 2  # seconds
+        interval_to_show = 15  # minutes
```
alexdewar, line 73: How about putting these in `config.py` with the other constants?

alexdewar (changes requested):
One small suggestion, but otherwise it looks good. The plot is much nicer now!

I was also thinking maybe the "hot" line should be red and the "cold" one blue. What do you think?

finesse/gui/temp_control.py
dc2917: Ah yes, good idea. I was wondering about where to set the poll interval as it belongs to the DP9800 but gets used in various places.

I always find any of the default matplotlib settings triggering, so I'm happy to change the line colours 😅

alexdewar (approved):
I think the poll interval and plot time range should be class attributes rather than being hardcoded to use the constants. I'd just have optional arguments for the class, with their defaults set to these constants. It makes it cleaner from a testing perspective. But it's up to you whether you can be bothered to do this.

Otherwise looking good!

dc2917:
> I think the poll interval and plot time range should be class attributes rather than being hardcoded to use the constants. I'd just have optional arguments for the class, with their defaults set to these constants. It makes it cleaner from a testing perspective. But it's up to you whether you can be bothered to do this.
> 
> Otherwise looking good!

I'm tempted to leave them as being set via the config file, as the poll interval applies to the `DP9800Controls` class but is also used by the `TemperaturePlot` class, and there doesn't seem to be a clean way to communicate this to `TemperaturePlot` if it's set in the `DP9800Controls` constructor.

alexdewar:
Fair enough!

PR #179, opened 2023-04-05
Remove web view from OPUS control

### LINKED ISSUE #169: Remove web view from `OPUSControl`
It's unnecessary and wastes space in the GUI.

alexdewar:
This has been agreed with the researchers. Having a web view there doesn't add anything -- all the info you need is in the log view anyway.

Removing it also means that we can remove the `url` argument from OPUS `pubsub` messages, as it is now unused.

Closes #169.


dc2917 (approved):
Nice!

PR #181, opened 2023-04-05
Use Qt network module for EM27 backends cf. `requests`

### LINKED ISSUE #175: Connecting to devices should not be done on main thread
It could potentially take on the order of seconds and we don't want to block the main thread for this long. In particular, I'm thinking of the `ST10Controller` class, which needs to wait for the motor to stop moving before it is truly initialised, but we should probably fix this for all devices while we're at it. It should be simple enough to call the device constructors from a separate thread.

### LINKED ISSUE #180: DummyEM27Scraper doesn't work with pyinstaller binary
The issue seems to be that the diag_autom.htm file is not being bundled.

alexdewar:
It turns out that I was reinventing the wheel with my `OPUSRequester` class: Qt already has a networking module that can make HTTP requests and return the result/error via a signal. I've switched both my `OPUSInterface` code and @dc2917's `EM27Scraper` code to use Qt networking instead, which means that HTTP requests are no longer being made on the main thread.

Closes #175. Closes #180.


finesse/hardware/em27_scraper.py
```diff
@@ -81,46 +82,48 @@ class EM27Error(Exception):
...
     """An interface for monitoring EM27 properties."""
 
-    def __init__(self, url: str = f"{EM27_URL}") -> None:
+    def __init__(self, url: str = EM27_URL) -> None:
         """Create a new EM27 property monitor.
 
         Args:
             url: Web address of the automation units diagnostics page.
         """
+        super().__init__()
```
dc2917, line 114: There is no parent class

alexdewar: Ohh yeah :-p

Issue #182, opened 2023-04-05
Allow for pausing measure scripts


alexdewar:
We should add a button to the progress dialog (when that's merged) that allows the user to pause the script partway through. The current software can do this.

Jon says that if the script is running a measurement, it should wait till the end of that measurement, but just not progress onto the next.

alexdewar:
It is possible that the `statemachine` library can help us here: https://python-statemachine.readthedocs.io/en/latest/guards.html

PR #183, opened 2023-04-05
Display dialog to show measure script progress

### LINKED ISSUE #71: Show progress bar when running measure scripts
Currently they just execute without any feedback to the user. Ideally we want a progress bar showing what is happening, how long there is to go and giving the user an option to cancel.

### LINKED ISSUE #158: Remember path to previously run measure script
It's getting annoying having to select the same file every time.

alexdewar:
This PR adds a dialog which shows the progress of a running measure script. The dialog appears when the user clicks "Run script" and disappears when the script finishes or the user clicks "Cancel" (or closes the window). Cancelling the dialog causes the measure script to stop. I haven't added a "Pause" button yet (#182) but that needs to be added at some point.

If you want to tinker with it, you might want to merge #178 into the branch so you can better see what the `DummyStepperMotor` is doing.

Closes #71. Closes #158.


dc2917 (approved):
Looks good and seems to work as expected!

PR #193, opened 2023-04-14
Temperature monitor refactor

alexdewar:
I came across this while I was working on the CSV code and thought I should fix it up. It will create merge conflicts with your new DP9800 branch, but I really don't mind rebasing this one once yours is merged if you'd prefer that.

At the moment `TemperatureMonitorBase` has `send_temperatures()` as an abstract method and the `DP9800` and `DummyTemperatureMonitor` classes each have their own implementation. I've rewritten it so that the child classes each have a `get_temperatures()` method instead and the `send_temperatures()` method which implements common functionality is in `TemperatureMonitorBase`.

I also noticed we were seeding the RNG with the current time, which is unnecessary. If you want a random seed you can just pass `None`.

### DIFF
--- a/finesse/hardware/temperature/dp9800.py
+++ b/finesse/hardware/temperature/dp9800.py
@@ -1,6 +1,5 @@
 """This module provides an interface to DP9800 temperature readers."""
 import logging
-from datetime import datetime
 from decimal import Decimal
 
 from pubsub import pub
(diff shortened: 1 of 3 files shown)


dc2917 (approved):
Makes sense! Looks much cleaner

alexdewar:
Are you happy for me to merge this now or would you like me to hold off till you've merged your DP9800 one?

dc2917:
> Are you happy for me to merge this now or would you like me to hold off till you've merged your DP9800 one?

Merge this and I'll merge into my DP9800 one

PR #195, opened 2023-04-14
Add `pubsub` decorators for convenience and safety

alexdewar:
In the process of doing something else, I started this refactoring and I thought it should be its own PR.

This PR adds a couple of decorators -- `pubsub_errors` and `pubsub_broadcast` -- which can be used to decorate functions. `pubsub_errors` catches any raised exceptions and forwards them as a pubsub message. `pubsub_broadcast` does the same thing but additionally will send a different message in the case that an exception is *not* thrown (i.e. that the function has succeeded). The user needs to supply the names of the keyword arguments that will be passed to the pubsub message in this case.

### DIFF
--- a/finesse/hardware/pubsub_decorators.py
+++ b/finesse/hardware/pubsub_decorators.py
@@ -1,8 +1,21 @@
 """Provides a decorator for catching and forwarding errors via pubsub."""
+import logging
+import traceback
 from collections.abc import Callable
 
 from decorator import decorator
 from pubsub import pub
+
+
+def _error_occurred(error_topic: str, error: BaseException) -> None:
+    """Signal that an error occurred."""
+    traceback_str = "".join(traceback.format_tb(error.__traceback__))
+
+    # Write details including stack trace to program log
+    logging.error(f"Caught error ({error_topic}): {traceback_str}")
+
+    # Notify listeners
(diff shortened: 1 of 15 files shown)


finesse/hardware/pubsub_decorators.py
```diff
@@ -0,0 +1,70 @@
...
+            _error_occurred(error_topic, error)
+        else:
+            # Convert result to a tuple of the right size
+            if result is None:
+                result = ()
+            elif not isinstance(result, tuple):
+                result = (result,)
+
+            # Make sure we have the right number of return values
+            assert len(result) == len(kwarg_names)
```
dalonsoa, line 64: In principle, you should not use `assert` to check runtime conditions. Use a conditional statement and raise the appropriate exception, if needed.

dalonsoa (approved):
The `decorator` package looks very handy. It is mostly syntactic sugar, I think, but quite convenient. It might obscure some of the decorators' magic, already quite esoteric, but I think it should be fine.

finesse/hardware/pubsub_decorators.py
dalonsoa: I think you can simply skip the assert and leave it fail wherever the two need to be the same.

PR #197, opened 2023-04-17
Add support for writing CSV files

### LINKED ISSUE #21: Ability to save sensor data to CSV file
The user should be able to save data from the sensors to a CSV file.

In the proposal it is not stated exactly what sensor data is required, but I suppose we should be aiming to save as much of the sensor data as possible.

We should ask researchers for more detail about this in case they have a preference in terms of data format etc. We could try to maintain a certain amount of compatibility with the data format used by the old program.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-5

### LINKED ISSUE #44: Adding additional data to the text logfile
The c# generated GUI outputs a logfile giving time, temperatures, mirror positions ect... I'd like to include the 12V power output from the TC-4820, given as a percentage of full power. This value is set and monitored by the TC-4820 and can be queried, in fact the c# code does this and displays the percentage power as a green bar in the lower right of the interface.

A text record of this value will enable me to calculate the balance between the radiative cooling of the hot blackbody, via its temperature with the input power from the TC-4820s.

Aha! Link: https://imperial-college-ict.aha.io/requirements/RSE-122-26

alexdewar:
This PR adds the option for the user to write sensor data to a CSV file. The format of the data files produced is the same as it was with the old program, with a couple of differences:

1. The files now have headers containing metadata (via `pycsvy`) and have column headings
2. The power of the hot blackbody's temperature controller is now also written to the file

Both of these changes have been signed off by @jonemurray.

The ST10, TC4820 (hot BB) and DP9800 must all be connected otherwise 

Closes #21. Closes #44.


finesse/hardware/data_file_writer.py
```diff
@@ -0,0 +1,117 @@
...
+            path: The path of the file to write to
+        """
+        self._writer = Writer(path, _get_metadata(path.name))
+
+        # Write column headers
+        self._writer.writerow(
+            (
+                "Date",
+                "Time",
+                *(f"Temp{i+1}" for i in range(config.NUM_TEMPERATURE_MONITOR_CHANNELS)),
```
dc2917, line 78: Do we want to add a format specifier here? Otherwise it will write the `Decimal`s with silly numbers of decimal places

finesse/hardware/data_file_writer.py
```diff
@@ -0,0 +1,117 @@
...
+        """Write temperature readings to the CSV file."""
+        # Also include timestamp as seconds since midnight
+        midnight = datetime(time.year, time.month, time.day)
+        secs_since_midnight = floor((time - midnight).total_seconds())
+
+        self._writer.writerow(
+            (
+                time.strftime("%Y%m%d"),
+                time.strftime("%H:%M:%S"),
+                *temperatures,
```
dc2917, line 108: Do we want to add a format specifier here? Otherwise it will write the `Decimal` values with silly numbers of decimal places

dc2917 (changes requested):
This all looks good, though I've made a small suggestion regarding the number of decimal places to write for the temperatures. 

I also wonder whether we always want to open the file in write mode? I don't know what a typical workflow might be, but I wonder if there's the possibility that one would begin logging, pause (for whatever reason) and then restart? Currently, this would overwrite the file.

finesse/hardware/data_file_writer.py
alexdewar: Good point. I'll look see what they did in the old program

alexdewar:
> I also wonder whether we always want to open the file in write mode? I don't know what a typical workflow might be, but I wonder if there's the possibility that one would begin logging, pause (for whatever reason) and then restart? Currently, this would overwrite the file.

That's a good point. Come to think of it, I think in the old program it would just append to the file if it already existed (which is easier to do if you don't have a header!). Maybe what I'll is raise a dialog to ask the user whether they want to overwrite it for now and open an issue re appending to existing files. Does that sound sensible?

finesse/hardware/data_file_writer.py
alexdewar: I've had a think about this and I think it's actually better if we *don't* use a format specifier. The real DP9800 has a fixed number of decimal places and we don't do any arithmetic with it so we should just end up with however many DPs the DP9800 gives us, which isn't many. Maybe let's just try it with the real device and see how the data look?

The downside is that we will end up with a silly no. of DPs for our simulated data.

dc2917: Ah yes, I guess because they're converted to `Decimal` from strings this should be fine

jonemurray:
The “old” would create a time_stamped log file name when the log started. I never tried stopping and starting it again once we had it running I can’t see the point in that but if you do pause I’d want to append when it was restarted. If we stop the logging entirely restart the GUI we get a new log file again with the filename incorporating the timestamp.

Does this help?

Jon

alexdewar:
That makes sense. Thanks Jon!

@dc2917 In that case, let's not bother with appending to files for now.

dc2917 (approved):

PR #204, opened 2023-04-18
Enable/disable buttons on device connect/disconnect

### LINKED ISSUE #191: Only enable "Run script" button when valid
Measure scripts can only be run when the EM27 and ST10 are connected, so the button should be disabled until this point.

### LINKED ISSUE #203: Only enable "Start recording" button when relevant devices are connected

alexdewar:
There are a couple of buttons -- the "Run script" and "Start recording" ones -- which should only be enabled when certain devices are connected. They should be disabled again when any of the devices disconnects. This means we need a counter to track how many of the required devices are connected at any one time, updating the GUI when this changes.

To do this, I've made a simple class -- `EventCounter` -- which takes a target count and two callbacks as parameters. The first callback is invoked when the target count is reached (i.e. all of the devices are connected) and the second is invoked when the count drops below the target (i.e. one of the devices has disconnected).

Closes #203. Closes #191.

### DIFF
--- a/finesse/gui/measure_script/script_view.py
+++ b/finesse/gui/measure_script/script_view.py
@@ -5,7 +5,8 @@ from typing import Optional, cast
 from pubsub import pub
 from PySide6.QtWidgets import QFileDialog, QGridLayout, QGroupBox, QPushButton
 
-from ...config import DEFAULT_SCRIPT_PATH
+from ...config import DEFAULT_SCRIPT_PATH, STEPPER_MOTOR_TOPIC
+from ...event_counter import EventCounter
 from ...settings import settings
 from ..path_widget import OpenPathWidget
 from .script import Script, ScriptRunner
@@ -41,6 +42,13 @@ class ScriptControl(QGroupBox):
 
         run_btn = QPushButton("Run script")
         run_btn.clicked.connect(self._run_btn_clicked)
+        run_btn.setEnabled(False)
+
+        self._enable_counter = EventCounter(
+            2, lambda: run_btn.setEnabled(True), lambda: run_btn.setEnabled(False)
+        )
+        """A counter to enable/disable the "Run" button."""
+        self._enable_counter.change_on_device_open(STEPPER_MOTOR_TOPIC)
 
         layout = QGridLayout()
         layout.addWidget(create_btn, 0, 0)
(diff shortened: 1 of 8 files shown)


finesse/gui/measure_script/script_view.py
```diff
@@ -41,6 +42,13 @@ def __init__(self) -> None:
 
         run_btn = QPushButton("Run script")
         run_btn.clicked.connect(self._run_btn_clicked)
+        run_btn.setEnabled(False)
+
+        self._enable_counter = EventCounter(
+            2, lambda: run_btn.setEnabled(True), lambda: run_btn.setEnabled(False)
+        )
+        """A counter to enable/disable the "Run" button."""
+        self._enable_counter.change_on_device_open(STEPPER_MOTOR_TOPIC)
```
dalonsoa, line 51: I would do this step as part of the initialisation of the `EventCounter`, so adding another argument to the constructor with the channel(s) this should be subscribed to and calling internally this method.

finesse/hardware/data_file_writer.py
```diff
@@ -54,13 +55,30 @@ def __init__(self) -> None:
         self._writer: Writer
         """The CSV writer."""
 
+        self._enable_counter = EventCounter(3, self.enable, self.disable)
+        self._enable_counter.change_on_device_open(
+            config.STEPPER_MOTOR_TOPIC,
+            config.TEMPERATURE_MONITOR_TOPIC,
+            f"{config.TEMPERATURE_CONTROLLER_TOPIC}.hot_bb",
+        )
```
dalonsoa, line 63: And the same with these. A 2-step initialisation does not look necessary.

dalonsoa (approved):
I've left a couple of comments, but it looks ok otherwise.

dc2917 (approved):
Looks good to me, and works nicely

finesse/gui/measure_script/script_view.py
alexdewar: That makes sense.

Note that you don't *have* to use the device topics with it though. In one place I'm increasing/decreasing the counter depending on the EM27's reported status. It probably does make sense to subscribe to the serial-related messages in the constructor though.

PR #207, opened 2023-04-19
Add support for pausing measure scripts

### LINKED ISSUE #182: Allow for pausing measure scripts

alexdewar:
Researchers want the option to pause running measure scripts, with the option of resuming them again at a later point.

I have implemented this so that when the user pauses the script, the program will finish the current stepper motor movement or interferometer measurement before stopping. (If there are repeat measurements to be done, the next measurement won't be started until the script is unpaused.) This is (I think!) what the existing program does!

I've implemented this by adding a couple of new possible `State`s to `ScriptRunner`: `waiting_to_move` and `waiting_to_measure`. In the various places where we currently transition to `moving` or `measuring` directly, we now instead transition to these "waiting" states. When the `ScriptRunner` enters these states, it will do nothing if paused and will transition directly to `moving` or `measuring` if unpaused. If the user unpauses the script while it is in one of these states, it will then transition to `moving` or `measuring`.

Closes #182.


alexdewar:
> It makes sense, and I think it is a useful feature.
> 
> Documentation-wise, for future developers (including ourselves) it would be useful to have a diagram of the state machine and its possible transitions, otherwise it might be hard to grasp what it is doing.

Done! It turned out to be surprisingly easy. We could even get extra fancy and automatically generate the state diagrams when building the documentation, but that's probably a job for another day.

PR #208, opened 2023-04-19
Fix initial poll of EM27

### LINKED ISSUE #136: Initial delay in reading EM27 sensor values

alexdewar:
The problem was that the updating of the GUI was being done at the same time as the
polling, whereas it actually needs to be done when the EM27 properties are received from
the server.

While I was at it, I renamed a file for consistency and made the polling interval a
constant in config.py.

Closes #136.


dc2917 (approved):
Nice one, seems to work well!

PR #215, opened 2023-04-27
Fix plot resizing without function

dalonsoa:
This PR makes small changes in the size policy and preferred size of the buttons, so that the temperature solution resizes correctly without having to use the `_resize_plot` method.

A more robust solution could be to use two `QVBoxLayout` side by side (one for buttons and other for the canvas), as it is done in the Opus view.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,8 @@
 {
-    "python.formatting.provider": "black",
+    "python.formatting.provider": "none",
     "[python]": {
         "editor.formatOnSave": true,
+        "editor.defaultFormatter": "ms-python.black-formatter",
     },
     "editor.rulers": [88],
 
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-    "python.formatting.provider": "none",
+    "python.formatting.provider": "black",
     "[python]": {
         "editor.formatOnSave": true,
         "editor.defaultFormatter": "ms-python.black-formatter",
--- a/finesse/gui/temp_control.py
+++ b/finesse/gui/temp_control.py
@@ -44,11 +44,6 @@ class TemperaturePlot(QGroupBox):
         layout = self._create_controls()
         self.setLayout(layout)
 
-        self.setSizePolicy(
-            QSizePolicy.Policy.Preferred,
-            QSizePolicy.Policy.MinimumExpanding,
-        )
-
         pub.subscribe(
             self._plot_bb_temps, f"serial.{TEMPERATURE_MONITOR_TOPIC}.data.response"
         )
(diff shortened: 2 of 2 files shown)


.vscode/settings.json
```diff
@@ -2,6 +2,7 @@
     "python.formatting.provider": "black",
     "[python]": {
         "editor.formatOnSave": true,
+        "editor.defaultFormatter": "ms-python.black-formatter",
```
alexdewar, line 5: I'm surprised that the `"python.formatting.provider"` option above wasn't enough. Didn't autoformatting work for you?

alexdewar (approved):
LGTM, though obvs the now-unused function should be ripped out before we merge into `main`.

If @dc2917 is also happy with this, then I think we should close both this PR and #209 and open a new PR with just this small change in (which I'm happy to do).

.vscode/settings.json
dalonsoa: I don't know. Sometimes VSCode has its own agenda and adds things to the settings without asking...

finesse/gui/temp_control.py
```diff
@@ -71,15 +66,15 @@ def _create_controls(self) -> QGridLayout:
         self._btns["cold"].clicked.connect(
             partial(self._toggle_axis_visibility, name="cold")
         )
+        self._btns["hot"].setMaximumWidth(80)
+        self._btns["cold"].setMaximumWidth(80)
```
dc2917, line 70: I think a cleaner way would be to just prevent them from expanding, via:
`self._btns["hot"].setSizePolicy(QSizePolicy.Policy.Fixed, QSizePolicy.Policy.Fixed)
self._btns["cold"].setSizePolicy(QSizePolicy.Policy.Fixed, QSizePolicy.Policy.Fixed)`

finesse/gui/temp_control.py
```diff
@@ -71,15 +66,15 @@ def _create_controls(self) -> QGridLayout:
...
         self._create_figure()
         self._canvas.setMinimumSize(QSize(640, 120))
-        self._canvas.setSizePolicy(
-            QSizePolicy.Policy.MinimumExpanding, QSizePolicy.Policy.MinimumExpanding
-        )
 
         layout.addWidget(self._btns["hot"], 0, 0)
         layout.addWidget(self._btns["cold"], 1, 0)
-        layout.addWidget(self._canvas, 0, 1, 2, 1)
+        layout.addWidget(self._canvas, 0, 1, 3, 1)
```
dc2917, line 77: I never would have thought to try that 😅. I think this makes it look much nicer with the buttons in the top corner as well.

dc2917 (changes requested):
This is a much cleaner solution than my hacky resizing function. I've made a small suggestion to set the size policy of the buttons rather than hard-coding a maximum size, but otherwise looks good!

finesse/gui/temp_control.py
dalonsoa: I tried this, but the problem is that the buttons will have different width, as the text inside has different width. The result looks slightly annoying with the two buttons a bit different.

alexdewar:
Did we reach a conclusion with all this resizing business?

alexdewar (approved):

alexdewar:
Dammit, I accidentally merged this. Sorry! Let me know if you want me to revert.

PR #216, opened 2023-04-27
Misc tidy-ups of OPUS code

alexdewar:
I've made some small cleanups to the OPUS code in the process of trying (and failing) to solve a separate issue. I figure there's probably enough here for a PR anyway, so here it is.

Commit summary:

- Use an enum to represent EM27 status in frontend and backend
- Tidy up `OPUSControl`'s `__init__`
- Make `OPUSInterfaceBase` a proper abstract class
- Remove duplicate fixture

### DIFF
--- a/finesse/em27_status.py
+++ b/finesse/em27_status.py
@@ -1,0 +1,22 @@
+"""Provides an enum representing the EM27's state."""
+from enum import Enum
+
+
+class EM27Status(Enum):
+    """The state of the EM27 interferometer.
+
+    These values are taken from the manual.
+    """
+
+    IDLE = 0
+    CONNECTING = 1
+    CONNECTED = 2
+    MEASURING = 3
+    FINISHING = 4
+    CANCELLING = 5
+    UNDEFINED = 6
+
+    @property
+    def connected(self) -> bool:
+        """Whether the state represents a connected status."""
+        return 2 <= self.value <= 5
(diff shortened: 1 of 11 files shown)


finesse/gui/measure_script/script_view.py
```diff
@@ -127,13 +128,10 @@ def _hide_run_dialog(self) -> None:
...
     def _on_opus_message(
-        self, status: int, text: str, error: Optional[tuple[int, str]]
+        self, status: EM27Status, text: str, error: Optional[tuple[int, str]]
     ) -> None:
         """Increase/decrease the enable counter when the EM27 connects/disconnects."""
-        # According to the manual, these are the possible states the EM27 can be in
-        # while connected
-        connected = 2 <= status <= 5
-
+        connected = status.connected
```
dalonsoa, line 134: This assignment does not seem necessary. Just use `status.connected` directly where relevant.

finesse/em27_status.py
```diff
@@ -0,0 +1,22 @@
...
+    IDLE = 0
+    CONNECTING = 1
+    CONNECTED = 2
+    MEASURING = 3
+    FINISHING = 4
+    CANCELLING = 5
+    UNDEFINED = 6
+
+    @property
+    def connected(self) -> bool:
```
dalonsoa, line 20: While it is clearly different than `CONNECTED` as it is lower case, it is the same word, which might be confusing when reading the code. I would use `is_connected`, which will give a better sense of this being a flag.

dalonsoa (approved):
Looks good! Using Enum is definitely the right choice there. I've just suggested a couple of changes to further improve readability.

dc2917 (approved):
Looks good to me!

PR #221, opened 2023-05-04
Provide initial x and y values for temperature plot

### LINKED ISSUE #220: Error in initialisation of temperature plot on Windows
On Windows, `datetime.datetime.fromtimestamp(t)` raises `OSError: [Errno 22] Invalid argument]` if `t` is a negative `float`. This causes a problem when the temperature plot is initialised, as the _x_ values are initially centred on zero, with tick values between -0.1 and 0.1, the tick labels at which are converted to datetimes using `datetime.datetime.fromtimestamp()`.

A simple solution is to simply set an initial (positive, non-zero) _x_ value and a corresponding _y_ value so that the plot doesn't initialise centred on (0, 0). It shouldn't matter what the value is - as soon as real data is plotted this value will be `pop`ped from the _x_ and _y_ value lists.

dc2917:
This PR provides a fix for the bug in initialisation of the temperature plot.
Initial _x_ value is set as the current time, and initial _y_ values are set to 25 degrees, representative of room temperature.

Requires some comments to satisfy `mypy` - I _think_ I've gone about this the right way.

Closes #220

### DIFF
--- a/finesse/gui/temp_control.py
+++ b/finesse/gui/temp_control.py
@@ -2,6 +2,7 @@
 from datetime import datetime
 from decimal import Decimal
 from functools import partial
+from typing import Optional  # noqa
 
 import matplotlib.pyplot as plt
 import matplotlib.ticker as ticker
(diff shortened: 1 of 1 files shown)


finesse/gui/temp_control.py
```diff
@@ -83,9 +84,13 @@ def _create_figure(self) -> None:
         self._figure_num_pts = int(
             TEMPERATURE_PLOT_TIME_RANGE / TEMPERATURE_MONITOR_POLL_INTERVAL
         )
-        t = [None] * self._figure_num_pts
-        hot_bb_temp = [None] * self._figure_num_pts
-        cold_bb_temp = [None] * self._figure_num_pts
+        t = [None] * self._figure_num_pts  # type: list[Optional[float]]
+        hot_bb_temp = [None] * self._figure_num_pts  # type: list[Optional[float]]
+        cold_bb_temp = [None] * self._figure_num_pts  # type: list[Optional[float]]
```
dalonsoa, line 89: I would do this in the following way:

```suggestion
        t: list[Optional[float]] = [None] * self._figure_num_pts
        hot_bb_temp: list[Optional[float]] = [None] * self._figure_num_pts
        cold_bb_temp: list[Optional[float]] = [None] * self._figure_num_pts
```

If done this way, you can also remove the noqa tag when importing Optional.

dalonsoa (approved):
Looks good. I've just suggested a change when defining the types.

PR #224, opened 2023-05-09
Removed newline character for cross-platform compatibility

dc2917:
This PR provides a fix for a bug in the web-scraping module.

A search is performed for the table header, which previously included a newline character. This caused an issue on Windows (`\n` vs `\r\n`), so the newline character has simply been removed.


dalonsoa (approved):
LGTM

PR #233, opened 2023-06-08
Misc hw fixes

### LINKED ISSUE #234: Write TC4820 power to data file as a percentage
`finesse/hardware/data_file_writer.py` currently writes the absolute power of the TC4820 temperature monitor to file. However, we would like it to be written as a percentage of its maximum power, `TC4820_MAX_POWER`, which is defined in `finesse/config.py`.

dc2917:
This PR contains some small fixes to hardware-related bits of the code following testing in the lab:

- Jon recommended polling the EM27 properties every minute (2s -> 60s)
- The TC4820 power meter is now normalised by the maximum TC4820 power, which has been moved to the config file
- TC4820 power output is written to data file as a percentage (closes #234)
- Reading from the DP9800 temperature monitor was incorrectly using the expected message size argument, which has now been removed
- DP9800 device is now polled as soon as connection is opened, for consistency with the other devices
- If error with serial device occurs, and connection closes, Open/Close button does not become appropriately "unchecked". This is now manually unchecked on device close.

### DIFF
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -39,7 +39,7 @@ DEFAULT_DATA_FILE_PATH = Path.home()
 EM27_URL = "http://10.10.0.1/diag_autom.htm"
 """The URL of the EM27 monitoring web server."""
 
-EM27_PROPERTY_POLL_INTERVAL = 2.0
+EM27_PROPERTY_POLL_INTERVAL = 60.0
 """Poll rate for EM27 properties."""
 
 STEPPER_MOTOR_TOPIC = "stepper_motor"
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -80,3 +80,6 @@ TEMPERATURE_MONITOR_HOT_BB_IDX = 6
 
 TEMPERATURE_MONITOR_COLD_BB_IDX = 7
 """Position of the cold blackbody on the temperature monitoring device."""
+
+TC4820_MAX_POWER = 511
+"""The maximum value for the power property."""
--- a/finesse/hardware/temperature/dp9800.py
+++ b/finesse/hardware/temperature/dp9800.py
@@ -161,13 +161,13 @@ class DP9800(TemperatureMonitorBase):
         Raises:
             DP9800Error: Malformed message received from device
         """
-        # require at least 4 bytes else checks will fail
-        min_length = 4
         try:
-            data = self.serial.read_until(b"\x00", size=min_length)
+            data = self.serial.read_until(b"\x00")
         except SerialException as e:
             raise DP9800Error(e)
 
+        # require at least 4 bytes else checks will fail
+        min_length = 4
         if len(data) < min_length:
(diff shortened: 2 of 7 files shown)


finesse/config.py
```diff
@@ -80,3 +80,6 @@
 
 TEMPERATURE_MONITOR_COLD_BB_IDX = 7
 """Position of the cold blackbody on the temperature monitoring device."""
+
+TC4820_MAX_POWER = 511
+"""The maximum value for the power property."""
```
alexdewar, line 85: What does this value mean? I'm guessing it's just a proportion rather than a physical unit?

How does Jon know this is the right maximum?

finesse/hardware/temperature/dp9800.py
```diff
@@ -161,13 +161,13 @@ def read(self) -> bytes:
         Raises:
             DP9800Error: Malformed message received from device
         """
-        # require at least 4 bytes else checks will fail
-        min_length = 4
         try:
-            data = self.serial.read_until(b"\x00", size=min_length)
+            data = self.serial.read_until(b"\x00")
```
alexdewar, line 165: Why has this been changed? There's no info in the commit message about why this change was needed.

alexdewar (commented):
Thanks for doing this! I've got a few queries, but I'm sure it's all fine.

Do you mind if we hold off on merging it until Monday though, just so we can check it with the real hardware one last time?

finesse/hardware/temperature/dp9800.py
dc2917: [read_until()](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.read_until) reads until the expected sequence, in this case `b"\x00"`, has been found, unless `size` bytes have been read, at which point it stops reading.

When reading from the actual device, reading would stop once `min_length` bytes had been read. This is set to 4 bytes, as this is the minimum that is required for the integrity checks to work, but we want to read the entire message, not just the first 4 bytes.

finesse/config.py
dc2917: Not sure what the units are, the manual states "511 represent 100% output". (Note you had this in `finesse/hardware/temperature/tc4820.py` but it was unused - I've just moved it to here so we can use it in `finesse/gui/temp_control.py`.)

finesse/hardware/temperature/dp9800.py
alexdewar: Ahh I see. That makes sense.

finesse/config.py
alexdewar: Okey doke. Whoops, it looks like I knew what it meant at some point then forgot and assumed it was 0-100 :laughing:

dc2917: I wonder if this will also be needed in the CSV-writing routine - I think we want the power output as a percentage of maximum, so we'll need to divide by 5.11.

alexdewar (approved):

PR #236, opened 2023-06-12
Fix pre-commit autoupdate hook

### LINKED ISSUE #235: pre-commit autoupdate action is failing
It is failing with this message:

```
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.10'
```

The culprit seems to be the `default_language_version` specified in
`.pre-commit-config.yaml`. The version of Ubuntu used for the autoupdate doesn't have
Python 3.10, so the hook fails.

alexdewar:
The problem seems to be that we have explicitly specified a Python version in
`.pre-commit-config.yaml` that doesn't exist on the CI runner used for the autoupdate
hook. I've tackled this by:

1. Unpinning the Python version used by `pre-commit`
2. Updating the Ubuntu version used by the hook to `ubuntu-latest`

The mystery remains as to why this wasn't causing an issue before.

NB: By using `ubuntu-latest`, the hook may break again if the hooks require an *older*
version of Python than that provided by whatever version of Ubuntu is being used, but
this will likely be trivial to fix anyway so probably isn't worth worrying about.

Fixes #235.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,3 @@
-default_language_version:
-    python: python3
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-    python: python3.10
+    python: python3
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.4.0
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-    python: python3.10
+    python: python3
```
dalonsoa, line 2: Actually, you can remove these lines altogether, as far as I can tell.

dalonsoa (approved):
Just a minor suggestion, but looks good otherwise. I think that sometimes there are just unholy combinations of package versions that work most of the time until they do not.

dc2917 (approved):

PR #257, opened 2023-07-04
Fix stepper motor timeout

### LINKED ISSUE #237: Connection to ST10 times out
The problem seems to be that the `wait_until_stopped()` method is timing out waiting for a response from the device (to the "send string" command). Curiously, it seems that the response is actually being received by the program, but seemingly it is not being returned appropriately to `wait_until_stopped()`. Presumably some of the logic around the `_SerialReader` class is broken.

alexdewar:
Fixes #237.

It took a while for me to figure out what was going on here, but it turns out there were two problems.

Firstly, the homing routine wasn't waiting for long enough before timing out. We forgot that the timeout is for *all* of the homing movements, not just the last one, so I've extended it to 10s.

Secondly, the logic around timing out for synchronous reads was broken. As we do all the reading for the device on a separate thread in order to intercept asynchronous messages (viz. the "I've stopped moving" message) that can occur at some unspecified time, we need the read timeout for the serial device to be infinity. On the other hand, we don't want the synchronous reads to wait forever (in case the device has disconnected), but the solution isn't to do what I did before and temporarily change the read timeout; what happens in that case is that the reading thread will start a new read with this new timeout then raise an exception when no new message appears, even though no message is expected. Now the timeout is set to infinity and instead we check whether synchronous reads have timed out by adding a timeout for reading the messages from `_SerialReader`'s `out_queue`.

I also changed things to allow the user to set the log level with an environment variable.

### DIFF
--- a/finesse/logger.py
+++ b/finesse/logger.py
@@ -15,8 +15,7 @@ def initialise_logging() -> None:
     filename = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
 
     # Allow user to set log level with environment variable
-    log_level = os.environ.get("FINESSE_LOG_LEVEL") or "INFO"
-    log_level = log_level.upper()
+    log_level = (os.environ.get("FINESSE_LOG_LEVEL") or "INFO").upper()
     if not hasattr(logging, log_level):
         raise ValueError(f"Invalid log level: {log_level}")
 
--- a/finesse/logger.py
+++ b/finesse/logger.py
@@ -1,5 +1,6 @@
 """Set up the program's logger."""
 import logging
+import os
 from datetime import datetime
 
 from platformdirs import user_log_path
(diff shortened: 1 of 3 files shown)


alexdewar:
@dalonsoa @dc2917 If either/both of you have a chance to review this today, that'd be great. I'm going into the lab tomorrow so ideally I'd like to test this branch after making any suggested changes if possible.

No biggie if you're busy though!

finesse/logger.py
```diff
@@ -13,9 +14,15 @@ def initialise_logging() -> None:
     log_path.mkdir(parents=True, exist_ok=True)
     filename = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
 
+    # Allow user to set log level with environment variable
+    log_level = os.environ.get("FINESSE_LOG_LEVEL") or "INFO"
+    log_level = log_level.upper()
```
dalonsoa, line 19:
```suggestion
    log_level = (os.environ.get("FINESSE_LOG_LEVEL") or "INFO").upper()
```
I've never been a big fan of assigning a variable just to change immediately afterwards.

dalonsoa (approved):
Looks good! Let's hope this works.

dc2917 (approved):
Looks good to me!

alexdewar:
Thanks both. I'll hold off on merging this until I've tested it with the real hardware.

PR #266, opened 2023-07-12
Fix stuck measure scripts

### LINKED ISSUE #244: Measure script fails to progress past first step
Seemingly the polling of the EM27's status sometimes stops at some point, meaning that the program never learns that the current measurement has finished and that it can therefore progress to the next step.

alexdewar:
When the EM27 is recording measurements during a measure script, it is repeatedly polled to check when the measurement has finished and the script can move onto the next step. Sometimes it would stop polling mid-run and get stuck forever.

After some digging, I finally figured out that the problem was that the `ScriptRunner` object responsible for, among other things, the timer polling the EM27, was sometimes being GC'd mid-run, which would stop the timer. Presumably every time the timer ran, the `ScriptRunner` was temporarily "revived" and therefore spared from the GC. This took me all afternoon to figure out and the result is a one-line fix :laughing: 

Fixes #244.

### DIFF
--- a/finesse/gui/em27_monitor.py
+++ b/finesse/gui/em27_monitor.py
@@ -1,4 +1,6 @@
 """Panel and widgets related to monitoring the interferometer."""
+import logging
+
 from pubsub import pub
 from PySide6.QtCore import Qt
 from PySide6.QtWidgets import (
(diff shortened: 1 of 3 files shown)


dc2917 (approved):
Interesting! Well done for finding the fix!

dalonsoa (approved):
Just one line, but it was an important line!

PR #267, opened 2023-07-12
Fix: Make step property return None if motor is moving

### LINKED ISSUE #250: ST10 error while recording data and running measure script
The following error occurs:

```
2023-06-27 16:49:02,071 [INFO] Measure script: Entering state Moving (event: start_moving)
2023-06-27 16:49:02,844 [ERROR] Caught error (data_file.error): Unexpected response when querying value SP

  File "/home/alex/code/FINESSE/finesse/hardware/pubsub_decorators.py", line 30, in wrapped
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/data_file_writer.py", line 132, in write
    get_stepper_motor_instance().angle,
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/stepper_motor_base.py", line 101, in angle
    return self.step * 360.0 / self.steps_per_rotation
           ^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/st10_controller.py", line 311, in step
    step = self._request_value("SP")
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/st10_controller.py", line 437, in _request_value
    raise ST10ControllerError(f"Unexpected response when querying value {name}")
```

From looking at the logs, it seems that the response to `SP` is `*`, meaning "command queued", which is not what I expected! I assumed that it would just return the current number of steps, even if the motor was in motion. So seemingly we need a new way to figure out the current angle.

alexdewar:
It seems that you can't query the stepper motor's position while it's in motion. This isn't a problem for the experiments, as these will only be carried out when the motor is stationary, but as the motor position is continually written into the data file it was breaking things. Now if the motor is moving, a `nan` is written instead of the angle.

Fixes #250.

### DIFF
--- a/finesse/hardware/stepper_motor/dummy.py
+++ b/finesse/hardware/stepper_motor/dummy.py
@@ -47,8 +47,16 @@ class DummyStepperMotor(StepperMotorBase):
         return self._steps_per_rotation
 
     @property
-    def step(self) -> int:
+    def is_moving(self) -> bool:
+        """Whether the motor is currently moving."""
+        return self._move_end_timer.isActive()
+
+    @property
+    def step(self) -> int | None:
         """The number of steps that correspond to a full rotation."""
+        if self.is_moving:
+            return None
+
         return self._step
 
     @step.setter
--- a/finesse/hardware/stepper_motor/dummy.py
+++ b/finesse/hardware/stepper_motor/dummy.py
@@ -53,7 +53,11 @@ class DummyStepperMotor(StepperMotorBase):
 
     @property
     def step(self) -> int | None:
-        """The number of steps that correspond to a full rotation."""
+        """The number of steps that correspond to a full rotation.
+
+        As this can only be requested when the motor is stationary, if the motor is
+        moving then None will be returned.
+        """
         if self.is_moving:
             return None
 
--- a/finesse/hardware/stepper_motor/st10_controller.py
+++ b/finesse/hardware/stepper_motor/st10_controller.py
@@ -296,14 +296,37 @@ class ST10Controller(StepperMotorBase):
         self._write_check(f"FL{steps}")
 
     @property
-    def step(self) -> int:
+    def status_code(self) -> int:
+        """The status code of the device.
+
+        For a complete list of status codes and their meanings, consult the manual.
+        """
+        # SC is formatted as a hexadecimal string
+        return int(self._request_value("SC"), 16)
+
+    @property
+    def is_moving(self) -> bool:
+        """Whether the motor is moving.
+
+        This is done by checking whether the status code has the moving bit set.
+        """
+        return self.status_code & 0x0010 == 0x0010
+
+    @property
+    def step(self) -> int | None:
         """The current state of the device's step counter.
+
+        As this can only be requested when the motor is stationary, if the motor is
+        moving then None will be returned.
 
         Raises:
             SerialException: Error communicating with device
             SerialTimeoutException: Timed out waiting for response from device
             ST10ControllerError: Malformed message received from device
         """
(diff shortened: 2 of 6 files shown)


finesse/hardware/stepper_motor/dummy.py
```diff
@@ -47,8 +47,16 @@ def steps_per_rotation(self) -> int:
         return self._steps_per_rotation
 
     @property
-    def step(self) -> int:
+    def is_moving(self) -> bool:
+        """Whether the motor is currently moving."""
+        return self._move_end_timer.isActive()
+
+    @property
+    def step(self) -> int | None:
```
dalonsoa, line 55: According to the docstring, this one is doing something different than the equivalent in the real device. Is this outdated?

finesse/hardware/stepper_motor/st10_controller.py
```diff
@@ -296,14 +296,37 @@ def _relative_move(self, steps: int) -> None:
...
+        # SC is formatted as a hexadecimal string
+        return int(self._request_value("SC"), 16)
+
+    @property
+    def is_moving(self) -> bool:
+        """Whether the motor is moving.
+
+        This is done by checking whether the status code has the moving bit set.
+        """
+        return self.status_code & 0x0010 == 0x0010
```
dalonsoa, line 313: I don't think I get this. Isn't the part after & always true? And if the order of operations is the opposite, isn't the equality always false, since True or False are never equal to 0x0010? I think I might be missing something.

finesse/hardware/stepper_motor/stepper_motor_base.py
```diff
@@ -95,10 +95,23 @@ def notify_on_stopped(self) -> None:
...
-        """The current angle of the motor in degrees."""
-        return self.step * 360.0 / self.steps_per_rotation
+        """The current angle of the motor in degrees.
+
+        Returns:
+            The current angle or NaN if the stepper motor is moving
+        """
+        step = self.step
+        if step is None:
+            return float("nan")
```
dalonsoa, line 112: This is done like this so there's something to output to the file while moving, but what's the meaning of outputing nan as an angle? Why not just not outputing anything?

dalonsoa (approved):
I think it looks good - I'm glad you're picking all of these annoying small bits and pieces. I've just left a couple of comments/questions.

finesse/hardware/stepper_motor/dummy.py
alexdewar: Indeed it is. Good spot. I've updated it now.

finesse/hardware/stepper_motor/stepper_motor_base.py
alexdewar: You mean return `None`? I did consider doing that, but this was slightly less effort :wink:. The nice thing about returning `None` though is we get the extra type-checking from `mypy`... I'll do that and save the `nan` conversion for when it's being written to the data file.

finesse/hardware/stepper_motor/st10_controller.py
alexdewar: It's a bitwise and operation, not a logical one (i.e. it compares each of the bits of the inputs and returns a new integer). I'm just checking whether this particular bit is set or not.

More info here: https://www.geeksforgeeks.org/check-whether-k-th-bit-set-not/

PR #269, opened 2023-07-13
Make writing data files safer

### LINKED ISSUE #265: Ensure CSV file is regularly flushed to disk
From testing, it seems that sometimes if you exit the program without closing the data
file, you can end up with an empty CSV file, which is not good. We should be flushing to
disk regularly and probably should invoke fsync() too for good measure.

alexdewar:
While tinkering with things I found that if you exit the program while the data file is recording, it stops abruptly without flushing to disk, so data is lost. Not ideal.

This PR addresses the problem by:

1. Ensuring the write buffer is regularly flushed (and flushed to disk on close)
2. Ensuring that data recording is stopped properly on program exit

Fixes #265.

Note that I'm doing the flushing by using a private member of the `Writer` class, which isn't ideal. In the long run, we could consider changing `pycsvy` to better support this use case: https://github.com/ImperialCollegeLondon/pycsvy/issues/60

### DIFF
--- a/finesse/hardware/data_file_writer.py
+++ b/finesse/hardware/data_file_writer.py
@@ -1,5 +1,6 @@
 """Provides a class for writing sensor data to a CSV file."""
 import logging
+import os
 import platform
 from datetime import datetime
 from decimal import Decimal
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
Looks good to me!

PR #270, opened 2023-07-13
Close serial devices on window close

### LINKED ISSUE #248: Mirror isn't always moved to nadir on program close
(Or maybe ever?)  I'm seeing warnings about `QThread`s being destroyed, which makes me wonder if it's related to #116.

alexdewar:
The devices are currently not closed cleanly on program exit. This doesn't matter in general, except that the `close()` method for the stepper motor controller turns the mirror downwards first (to prevent dust buildup), so ideally we do want this to happen.

Fixes #248.


dalonsoa (approved):
LGTM

PR #288, opened 2023-07-19
Measure script dialog enhancements

### LINKED ISSUE #283: Don't cancel recording if measure script window closed
@jonemurray says he'd prefer for the EM27 to continue with its current measurement, if possible.

### LINKED ISSUE #284: Warn user if they attempt to close measure script dialog
Currently if the user does this, it will just abort the execution of the measure script.
Given that these scripts can sometimes be running for hours, it would be good to at
least confirm with the user that this is what they want.

alexdewar:
This PR implements a couple of changes relating to measure scripts: https://imperialcollegelondon.github.io/FINESSE/measure_scripts/

The first is that I removed a feature whereby recording from the interferometer would stop if the measure script dialog was closed (fixes #283), because the researcher said they might want this data.

The second is that I added a warning if users try to close the measure script dialog while it is running (fixes #284).

### DIFF
--- a/finesse/gui/measure_script/script.py
+++ b/finesse/gui/measure_script/script.py
@@ -196,9 +196,7 @@ class ScriptRunner(StateMachine):
     """Record another measurement at the same angle."""
     finish_waiting_for_measure = waiting_to_measure.to(measuring)
     """Stop waiting and start the next measurement."""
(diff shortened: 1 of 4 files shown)


cc-a (approved):
Wasn't able to test this locally but looks good.

PR #291, opened 2023-07-19
Data recording enhancements

### LINKED ISSUE #249: User is given warning about overwriting data file twice

### LINKED ISSUE #281: Continue data recording if USB connections drop
At present, if any of the sensors whose values are written into the CSV files suddenly
disconnect, writing to the CSV file will cease.

@jonemurray reports that it would be better to write NaN values for the missing sensors
so that data recording can continue.

### LINKED ISSUE #282: Add timestamps to CSV file names
Currently the user has to enter/choose the exact filename they want. @jonemurray says it
would be better to have timestamps built into the filename.

This would also help prevent overwriting existing data (although currently the user will
be prompted if they try to do this).

alexdewar:
This PR implements a couple of changes to the way we're recording data files.

Currently, you can only start a data recording when all the devices whose data will be used are connected (viz. the ST10, the hot BB's TC4820 and the DP9800). The problem is that @jonemurray reports he may want to record data without some of these devices present. Additionally, if the USB connection drops, currently the data recording stops, but it would be better if it carried on in some form. I've changed things to write NaNs for any sensor values that we don't have, which has required a bit of refactoring.

Secondly, @jonemurray asked that timestamps be included in the CSV file names, so I've reworked the GUI side a bit to accommodate this. Now instead of choosing the destination filename directly, the user chooses the destination folder and the filename prefix and the timestamp is appended.

Closes #281. Closes #282. Closes #249.

### DIFF
--- a/finesse/gui/data_file_view.py
+++ b/finesse/gui/data_file_view.py
@@ -1,17 +1,20 @@
 """Provides a panel which lets the user start and stop recording of data files."""
+from datetime import datetime
 from pathlib import Path
 
 from pubsub import pub
 from PySide6.QtWidgets import (
     QGroupBox,
     QHBoxLayout,
+    QLabel,
+    QLineEdit,
     QMessageBox,
     QPushButton,
     QSizePolicy,
 )
 
 from ..config import DEFAULT_DATA_FILE_PATH
-from .path_widget import SavePathWidget
+from .path_widget import OpenDirectoryWidget
 
 
 class DataFileControl(QGroupBox):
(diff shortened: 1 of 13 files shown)


cc-a (approved):
Wasn't really able to test this but makes sense to me. All I would say is that the test coverage doesn't look great. Anything that can be done to improve that or is it tricky to cover because of the hardware interaction aspect?

alexdewar:
Yeah, I know :disappointed:. Essentially the reason the coverage is poor is because we didn't write tests as we went along and although I went back and added some, we ran out of time and so prioritised getting new features in, even if that meant not including tests for them. The hardware integration itself isn't too hard to test: you just mock your `Serial` object or whatever. We just haven't done it consistently :stuck_out_tongue: 

There is a follow-up project to work on this code base (UNIRAS) and I'll add the missing tests at that point.

PR #292, opened 2023-07-20
Rearrange panels

alexdewar:
This PR rearranges the panels to a more sensible configuration. The only problem with the new layout, I think, is that the plot is perhaps a bit too stretched vertically. But it's probably good enough.

Old layout:

image

New layout:

image


cc-a (approved):
LGTM

Issue #323, opened 2023-10-02
Remember USB devices based on persistent attributes cf. COM port numbers


alexdewar:
Currently, FINESSE remembers the last COM port (or the Unix equivalent thereof) a device was connected to and selects this as the default option for that device in the GUI. The downside of this approach is that, in Windows, the COM port a device will be assigned depends on the port into which it is plugged, so you have to remember which port that was, or update the COM port used in the GUI. On Unix it's worse: ports are numbered depending on the order in which they were connected.

It would be better to select which port to connect to based on persistent device properties, e.g. vendor ID, product ID or serial number. These are provided for all USB devices in their [device descriptor](https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-device-descriptors#sample-device-descriptor). Sometimes I've seen devices without serial numbers, but that field would obviously be the best one to use.

Note that in our case the USB device descriptor describes the USB-to-serial adaptor, not whatever is wired up to it (e.g. a stepper motor controller). However, the wiring is not likely to change a whole lot, so this is still a better approach than the current one.

PR #336, opened 2023-10-06
Add plugin system for hardware and use for serial devices

### LINKED ISSUE #319: Make hardware used configurable
We need to be able to select between the hardware used for the different FINESSE and UNIRAS rigs. This configuration could possibly be done using a config file and some kind of plugin system as well as regular Python.

As we often want to use dummy devices in place of some or all of the sensors, this functionality may also be useful for letting the user choose which hardware they want.

alexdewar:
## Background

FINESSE connects to various devices: a stepper motor, two temperature controllers, a temperature monitor and the EM27 spectrometer. Except for the spectrometer, these devices are connected via USB serial. For each device, there is a "real" implementation for connecting to actual hardware as well as a dummy version for development. To use the dummy version of a serial device, you select "Dummy" as the COM port from the dropdown menu and for the EM27 you pass an extra command-line argument. Apart from being somewhat ugly, this strategy is also not particularly extensible, which will especially be a problem when we come to add new device types for UNIRAS. It would be better if the user could choose which device to use for a given type, whether that's the "real" one using USB serial, the dummy device or a new type of device added for UNIRAS. (This was suggested by @dalonsoa when the original code was merged, but we ran out of time to implement it.)

As different devices have different parameters (e.g. USB serial devices need a port and baudrate specified), the choices for parameters could be shown dynamically in the GUI based on what the currently selected device requires.

## Changes

This PR adds a plugin system for defining different types of hardware and refactors the existing serial devices to use this system. While the backend code has changed considerably, the basic design of the program is the same, i.e. users still have to manually connect to serial devices using a panel in the bottom left of the GUI. Eventually we will want the code to automatically connect to devices (#323) at which point this panel can be removed (and possibly repurposed into a separate configuration dialog). I have only made the serial devices plugins for now. The plan is to do the same thing for the EM27 code too eventually, but it will require some additional changes (e.g. there are EM27-specific GUI components which should also probably be part of the plugin).

## Implementation

There are two types of plugin, device base types and device types, both of which are represented with classes that inherit from `DeviceBase`. The base types are abstract classes which define the interface for a particular kind of hardware (e.g. a stepper motor). The device types are concrete classes which *must* inherit from one of the base types (i.e. they cannot inherit from `DeviceBase` directly). They must also be defined using one of the plugin decorators in order to be discoverable. Base types should use `@register_device_base_type` and device types can use `@register_device_type` or `@register_serial_device_type`. The decorators add some extra class variables (e.g. device parameters) and add the classes to common registries.

All plugins must live in `finesse.hardware.plugins`. The code obtains a list of device types by recursively importing the submodules from `finesse.hardware.plugins` and then checking the device registry. Note that this step is only necessary to obtain a list of all available device types, which we have to do to populate the GUI, but you can just import device classes directly to use them.

Each connected device is uniquely identified by a string representing its base type (e.g. "stepper_motor") and where there can be more than one instance of a base type, an additional name (e.g. "temperature_controller.hot_bb").

## Rough edges

* Currently there are no tests
* It's a bit ugly to pass `instance` into `"device.error.*"` messages
* I kept the current hack to report the temperature as NaNs when there is no temperature monitor connected (we need this for the data file recording) but there are definitely cleaner solutions

Fixes #319.


finesse/hardware/__init__.py
```diff
@@ -10,31 +11,102 @@
...
-)
+from datetime import datetime
 
-opus: OPUSInterface
+from finesse.config import NUM_TEMPERATURE_MONITOR_CHANNELS, TEMPERATURE_MONITOR_TOPIC
+from finesse.device_info import DeviceBaseTypeInfo, DeviceInstanceRef, DeviceTypeInfo
+
+from . import data_file_writer  # noqa: F401
+from .plugins import load_device_types
+from .plugins.temperature import get_temperature_monitor_instance
```
dalonsoa, line 21: Why is the temperature monitor treaded in a different way?

finesse/hardware/data_file_writer.py
dalonsoa, line 1: This file mixes stuff for the stepper motor and the temperature controller, which defeats a bit the modularity aims as someone adding a new device might need to edit this file to add its own stuff. If it makes sense for each hardware to have more than one module for some reason, maybe they can live in their own packages.

finesse/hardware/plugins/__init__.py
dalonsoa, line 1: Decorators are great for registering functionality because they are explicit. But when it comes to registering classes that need to be inheriting from a specific subclass, they are redundant: you can achieve identical functionality with less code simply by adding the registration step to the `__init_subclass__` method of the relevant base class. Not only it will be less code but also saves the user/developer to use both the decorator and the base class. 

Here you have an example illustrating the process:
```python
register = {}

class Base:

    def __init_subclass__(cls, long_name):
        cls.long_name = long_name
        if cls.__name__ not in register:
            register[cls.__name__] = cls

class Inherited(Base, long_name="A very long name"):
    pass

a = register["Inherited"]
print(a.long_name)
```

dalonsoa (commented):
I've just made a very lightweigth review and point to a couple of questions, suggestions. A much deeper one will come as soon as possible.

finesse/hardware/plugins/__init__.py
alexdewar: Ahh I didn't know you could do this. That's much nicer! I'll have a go at doing it this way

finesse/hardware/__init__.py
alexdewar: It's a hack -- I mentioned it briefly in the PR description.

The temperatures are requested from the temperature monitor every 2s and they are used to a) update the plot and b) save into the data file (if recording is happening). The problem is that at some point Jon decided that he wanted data files to carry on recording regardless of whether the temperature monitor is active. In general this is a better design, because it makes things more resilient in the case of the temperature monitor breaking, but the way I implemented it was to add a hack that NaNs are broadcast in place of real temperatures if the device isn't working. A better way to do it would be to have all the sensors whose data *can* be recorded into data files respond to a different, specific message used for this purpose only.

I've opened a new issue and copied this description into it: #337

finesse/hardware/data_file_writer.py
alexdewar: That's a good point. As I said in a previous comment, this is all basically a bit of a hack. It would be nicer if the data file module could figure out what columns should exist and what values they should be populated with dynamically. Maybe it could be resolved along with #337

finesse/hardware/plugins/__init__.py
alexdewar: I've had a go at this... let me know what you think. I had to do some nasty shenanigans in various `__init_subclass__` methods in order to get it to work the way I wanted, but I think the interface is at least a bit nicer now.

codecov.yml
```diff
@@ -6,4 +6,7 @@ coverage:
         informational: true
     patch:
       default:
-        informational: true
\ No newline at end of file
+        informational: true
+# HACK: Ignore hardware module for now to avoid spammy messages
+include:
```
dalonsoa, line 11: Maybe "exclude" instead of "include"?

finesse/hardware/device.py
```diff
@@ -0,0 +1,171 @@
...
+        # **HACK**: Allow callers to omit this arg in the case that they are
+        # non-concrete
+        if not description:
+            return
+
+        # Set device description for this class
+        cls._device_description = description
+
+        # Add the class to the registry of device types
+        _device_types.add(cls)
```
dalonsoa, line 88: I think I might be missing something. Is there any reason why you cannot do the following and skip the `DeviceBaseType` class below altogether?


```suggestion
    def __init_subclass__(cls, is_base_type: bool = False, **kwargs: Any) -> None:
        """Initialise a device type class.

        While the description argument has to be optional to allow for non-concrete
        classes to inherit from Device, it is mandatory to include it, else the device
        type class will not be added to the registry.

        Args:
            description: Human-readable name for this device type.
        """
        # If it is a base class, we initialise it as such
        if is_base_type:
            cls._init_base_type(**kwargs)
            return

        # If it is not, it should inherit from one.
        # Not sure how robust this will be to differences in the registration order...
        if not set(cls.__bases__).intersection(_base_types):
            raise ValueError(f"Class {cls.__name__} must be a device base type or inherit from one.")

        # And we initialise it as a concrete device class
        cls._init_device_type(**kwargs)

    @classmethod
    def _init_base_type(
        cls,
        name: str,
        description: str,
        names_short: Sequence[str] = (),
        names_long: Sequence[str] = (),
        **kwargs,
    ) -> None:
        super().__init_subclass__(**kwargs)
        # Store metadata about this base class
        cls._device_base_type_info = DeviceBaseTypeInfo(
            name, description, names_short, names_long
        )

        # Add the class to the registry of base types
        _base_types.add(cls)

    @classmethod
    def _init_device_type(
        cls,
        name: str,
        description: str,
        **kwargs,
    ) -> None:
        super().__init_subclass__(**kwargs)
        # Set device description for this class
        cls._device_description = description

        # Add the class to the registry of device types
        _device_types.add(cls)
```

dalonsoa (changes requested):
I've focused only on the `__init__subclass__` business. Will review the rest once that is out of the way. As I point out in one comment, I don't fuly understand the need of a `Device` class and a `DeviceBaseType`. I think they can be combined easily, but maybe I'm missing something?

finesse/hardware/device.py
dalonsoa: You can even go one step further and take the decision of a class being base or not depending on the class name instead of an input parameter (although that's more explicit): if it ends in `Base` it is a base class and not otherwise. `StepMotorBase` will be considered a base class, so will `TempControllerBase`, but not `TempController` or `MyFunnyClassWithOddName`.

codecov.yml
alexdewar: Ohh dear :laughing:. It turns out it should be `ignore` instead.

finesse/hardware/device.py
```diff
@@ -0,0 +1,171 @@
...
+        self.topic += f".{name}"
+
+    @abstractmethod
+    def close(self) -> None:
+        """Close the connection to the device."""
+
+    @classmethod
+    def get_device_base_type_info(cls) -> DeviceBaseTypeInfo:
+        """Get information about the base type for this device type."""
+        return cls._device_base_type_info
```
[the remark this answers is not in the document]
alexdewar: I guess it just felt a bit odd to have `get_device_type_info` as a `classmethod` and
then let users access this attribute directly. I take your point but I'm a bit on the
fence about this.

finesse/hardware/device.py
alexdewar: Interesting! I like the idea of specifying `is_base_type` explicitly. I'll have a tinker
with this.

The reason for having a separate `DeviceBaseType` class was partly because
`SerialDevice` inherits from `Device` but *not* `DeviceBaseType`, because it isn't a
concrete device implementation (it isn't a device base type either actually). I guess we
could either a) make an `AbstractDevice` parent class which includes just the bits actually
shared with `SerialDevice` or b) make `SerialDevice` not inherit from anything (not sure
if this will break something or not).

dalonsoa: So, how this goes? I think I'm missing the overall architecture of classes you are implementing. 

In the current implementation, I thought you would have a, lets say, `StepperMotorBase` inheriting from `DeviceBaseType`, and a `ModelXYZStepperMotor` as a concrete, usable device, inheriting from `StepperMotorBase` (because it is a stepper motor) and `SerialDevice` (because it is a serial device). Why does `SerialDevice` need to inherit from `Device` without being a `DeviceBaseType`? Or why it needs to inherit from `Device` at all?

It seems to me you might be mixing two orthogonal concepts here: device types (stepper motor, temp controller, etc.) with connection types (serial, GPIB, ethernet...). It makes sense to me that a concrete device needs to inherit from both (or use composition, instead), but I don't see any reason why these two concepts need to have a common ancestor to inherit from.

alexdewar: Right, so I've finally finished doing this.

I did end up making an `AbstractDevice` class as a separate base class. The reason is that the `SerialDevice` class needs to modify a class attribute (device parameters) which is shared with other device types.

finesse/hardware/serial_device.py
```diff
@@ -0,0 +1,64 @@
...
+        """Add serial-specific device parameters to the class."""
+        super().__init_subclass__(**kwargs)
+
+        # TODO: Allow for adding parameters elsewhere rather than clobbering them
+        cls._device_parameters = [
+            DeviceParameter("port", _get_usb_serial_ports()),
+            DeviceParameter(
+                "baudrate", list(map(str, BAUDRATES)), str(default_baudrate)
+            ),
+        ]
```
dalonsoa, line 56: You probably have though og this, but as it is now, if the concrete device class adds its own parameters (and depending on the order of the parent classes) these parameters will be overwritten. You are going to need to put in place some checks to make sure that parameters added in the multiple inheritance process can all live with each other.

dalonsoa (approved):
Looks good! I'm approving this PR, but I've left a comment that will need addressing.

finesse/hardware/serial_device.py
alexdewar: I thought about it and was like "ah I'll deal with it later", then promptly forgot to do so :stuck_out_tongue:. You're right, it does need fixing.

alexdewar:
Right, I think this is all done now!

@dc2917 Do you think you'll have a chance to have a look at this or shall I just go ahead and merge it?

dc2917:
> Right, I think this is all done now!
> 
> @dc2917 Do you think you'll have a chance to have a look at this or shall I just go ahead and merge it?

Yep, I'll take a look

finesse/device_info.py
```diff
@@ -0,0 +1,74 @@
...
+    """A human-readable name for the device."""
+    parameters: list[DeviceParameter]
+    """The device parameters."""
+    class_name: str
+    """The name of the device's class including the module name."""
+
+
+@dataclass(frozen=True)
+class DeviceBaseTypeInfo:
+    """A base device type (e.g. stepper motor)."""
```
dc2917, line 44:
```suggestion
    """A generic device type (e.g. stepper motor)."""
```
Maybe "generic" is a slightly better descriptor, being less self-referencing? Though maybe "base" works better for identifying the base class.

finesse/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -151,7 +157,9 @@ def read_sync(self, timeout: float | None = None) -> str:
         return response
 
 
-class ST10Controller(StepperMotorBase):
+class ST10Controller(
+    SerialDevice, StepperMotorBase, description="ST10 controller", default_baudrate=9600
```
dc2917, line 161: Would it perhaps make more sense to use the `__str__` or `__repr__` methods for setting the `description`?

dc2917 (approved):
I've only made a couple of super minor comments, otherwise this all looks really good to me. The new "Device control" panel on the GUI looks cool!

finesse/hardware/plugins/stepper_motor/st10_controller.py
alexdewar: In this case that wouldn't work, because we need the description to be a *class* variable. The plugin-related code uses it to get details about the `ST10Controller` plugin before an instance of it is created.

alexdewar:
Seeing as I'm going into the lab on Monday anyway I think I'll hold off on merging this until I've tested it (again) with the actual hardware.

alexdewar:
I've tested it on the real hardware and everything still seems to be working.

PR #351, opened 2023-10-12
User Guide Placeholder and Script

### LINKED ISSUE #344: Add placeholder file(s) for user guide
Before we can move onto actually doing anything with the user guide (e.g. auto-building it with the CI system) we need a placeholder file (or several) to be added to the repo for when we actually have content to put in there.

### LINKED ISSUE #345: Add script for building user guide
Once we have some placeholder content (#344), we will need to build it somehow, either into HTML or (possibly) a PDF, depending on what we want to do with it.

I suspect the easiest way to do this is to build it with `pandoc`. Probably this should be done in a Python script for cross-platform compatibility. Alternatively we *might* be able to integrate this with our existing `mkdocs`-based system for building developer documentation.

We only need to be able to do this locally for now; CI integration can wait.

CWestICL:
Added `user_guide.md` as a placeholder within `docs`. Also added `gen_user_guide.py` script that runs a `pandoc` command to produce a new `.html` file from `user_guide.md`.

A local version of `pandoc` will need to be installed to your local machine, with the `pandoc` command working within the project.

Closes #344 
Closes #345

### DIFF
--- a/docs/gen_user_guide.py
+++ b/docs/gen_user_guide.py
@@ -10,7 +10,7 @@ subprocess.run(
         "-t",
         "html5",
         "-o",
-        "user_guide.html",
+        f"{Path(__file__).parent}/user_guide.html",
         f"{Path(__file__).parent}/user_guide.md",
     ],
     check=True,
--- a/docs/gen_user_guide.py
+++ b/docs/gen_user_guide.py
@@ -1,0 +1,7 @@
+"""Generate the User Guide."""
+import os
+
+os.system(
+    r"""pandoc -f markdown -t html5 -o user_guide.html docs\user_guide.md -c style
+    .css"""
+)
--- a/docs/gen_user_guide.py
+++ b/docs/gen_user_guide.py
@@ -1,6 +1,17 @@
 """Generate the User Guide."""
 import subprocess
+from pathlib import Path
 
 subprocess.run(
-    "pandoc -f markdown -t html5 -o user_guide.html docs/user_guide.md", check=True
+    [
+        "pandoc",
+        "-f",
+        "markdown",
+        "-t",
+        "html5",
+        "-o",
+        "user_guide.html",
+        f"{Path(__file__).parent}/user_guide.md",
+    ],
+    check=True,
 )
--- a/docs/gen_user_guide.py
+++ b/docs/gen_user_guide.py
@@ -1,7 +1,6 @@
 """Generate the User Guide."""
-import os
+import subprocess
 
-os.system(
-    r"""pandoc -f markdown -t html5 -o user_guide.html docs\user_guide.md -c style
-    .css"""
+subprocess.run(
+    "pandoc -f markdown -t html5 -o user_guide.html docs/user_guide.md", check=True
 )
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,9 @@ instance/
 # Sphinx documentation
 docs/_build/
 
+# User Guide documentation
+docs/user_guide.html
+
 # PyBuilder
 target/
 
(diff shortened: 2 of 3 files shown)


docs/gen_user_guide.py
```diff
@@ -0,0 +1,7 @@
+"""Generate the User Guide."""
+import os
+
+os.system(
+    r"""pandoc -f markdown -t html5 -o user_guide.html docs\user_guide.md -c style
+    .css"""
+)
```
alexdewar, line 7: The general idea is correct here, but there are a few issues with this.

1. You're using a backslash as a path separator which only works on Windows. If you use a forward slash it'll work anywhere
2. Your editor has broken this string, so `.css` is on its own line and Python will try to run it as a separate command
3. I don't think you need the `-c style.css` bit anyway -- there is no `style.css` file!
4. This will only work if you run the script from the parent folder. If you do `cd docs; python gen_user_guide.py` it'll fail, because the path to `user_guide.md` will now be wrong. You can get the path to the folder the current script is in with `Path(__file__).parent`. I'd do this to make sure both the input and output path are pointing at the right place
5. I'd use `subprocess.run` to invoke pandoc rather than `os.system`, which is deprecated (see https://hackernoon.com/calling-shell-commands-from-python-ossystem-vs-subprocess-mc253z4f). You'll want the `check=True` option enabled

alexdewar (changes requested):
This is most of the way there, though I've got some minor suggestions. Currently the script doesn't work because your editor has mangled it.

alexdewar:
Btw, you also need to write "Closes #345" separately in your PR message. Oddly, if you write "Closes #x, #y", GitHub won't pick up that it closes #y.

alexdewar:
Two more things:

- I've marked this as ready for review, because I think it is!
- I've tagged Dan as a reviewer because he's also on the UNIRAS team. In general we do want to try to remember to tag everyone on the project so they're at least aware of what's going on

dc2917 (changes requested):
1. The `subprocess.run()` command didn't work for me. I think because you have spaces in the command you need to provide the string as a list of sub strings - see the examples [here](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module)
2. To resolve the path to the input and output files regardless of the user's current working directory, you'll need to use `pathlib`'s `Path` class as Alex suggested. The command currently only works if run from the FINESSE directory, which then places user_guide.html there (rather than in FINESSE/docs, which is presumably where we want it)

docs/gen_user_guide.py
```diff
@@ -0,0 +1,17 @@
...
+    [
+        "pandoc",
+        "-f",
+        "markdown",
+        "-t",
+        "html5",
+        "-o",
+        "user_guide.html",
+        f"{Path(__file__).parent}/user_guide.md",
+    ],
```
dc2917, line 15: I think we should probably ensure the user guide is written to the docs folder as well, so if we do something like the following we can ensure it's always written there, regardless of where `gen_user_guide.py` is run from.

```suggestion
docs_dir = Path(__file__).parent
subprocess.run(
    [
        "pandoc",
        "-f",
        "markdown",
        "-t",
        "html5",
        "-o",
        f"{docs_dir}/user_guide.html",
        f"{docs_dir}/user_guide.md",
    ],
```

dc2917 (changes requested):
One minor change and then this should be good to merge

docs/gen_user_guide.py
CWestICL: Cool looks good, I'll commit that change

dc2917 (approved):

PR #355, opened 2023-10-16
Express TC4820 power as a percentage of maximum

### LINKED ISSUE #289: Power for TC4820s should be expressed as a percentage
Currently it's displayed in the GUI as an integer between 0 and 511, which isn't v informative for the user.

dc2917:
This PR changes `TC4820.power()` to return the current power level as a percentage of maximum (`config.TC4820_MAX_POWER`), as it is only ever used (i.e. when written to file and displayed on GUI) as a percentage.
Essentially, the scaling is just performed in `TC4820.power()` rather than wherever it is used.

Note that the textbox next to the power meter on the GUI also shows the numerical value. In the original GUI, this instead showed an integer corresponding to the alarm status of the device, but I'm not sure whether that was useful (or even known, hence the assumption that this is instead the percentage power). It seems unnecessary to have this textbox since we could instead just use `setTextVisible(True)` and show this value on the power meter itself, but maybe there's a preference to have this next to, rather than on top of, the power meter - especially if no one cares about the alarm status value.

Closes #289

### DIFF
--- a/finesse/hardware/data_file_writer.py
+++ b/finesse/hardware/data_file_writer.py
@@ -79,7 +79,7 @@ def _get_hot_bb_power() -> float:
         return float("nan")
 
     try:
-        return hot_bb.power * 100 / config.TC4820_MAX_POWER
+        return hot_bb.power
     except Exception as error:
         pub.sendMessage(
             f"serial.{config.TEMPERATURE_CONTROLLER_TOPIC}.{hot_bb.name}.error",
--- a/finesse/hardware/temperature/tc4820.py
+++ b/finesse/hardware/temperature/tc4820.py
@@ -17,7 +17,6 @@ from decimal import Decimal
 
 from serial import Serial, SerialException
 
-from ...config import TC4820_MAX_POWER
 from .temperature_controller_base import TemperatureControllerBase
 
 
@@ -41,6 +40,7 @@ class TC4820(TemperatureControllerBase):
 
         self.serial = serial
         self.max_attempts = max_attempts
+        self.max_power = 511
 
         super().__init__(name)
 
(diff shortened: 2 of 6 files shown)


dalonsoa (approved):
LGTM

finesse/hardware/temperature/tc4820.py
```diff
@@ -155,8 +156,8 @@ def temperature(self) -> Decimal:
 
     @property
     def power(self) -> int:
-        """The current power output of the device."""
-        return self.request_int("020000")
+        """The current power output of the device, as a percentage of maximum."""
+        return int(self.request_int("020000") * 100 / TC4820_MAX_POWER)
```
alexdewar, line 160: I think we should just return a float instead -- you can `round()` it for the progress bar.

finesse/hardware/data_file_writer.py
```diff
@@ -79,7 +79,7 @@ def _get_hot_bb_power() -> float:
         return float("nan")
 
     try:
-        return hot_bb.power * 100 / config.TC4820_MAX_POWER
+        return hot_bb.power
```
alexdewar, line 82: Technically this is returning an `int`, not a `float` now. Not sure why `mypy` isn't complaining.

alexdewar (changes requested):
Looks good, except that I would use a `float` rather than an `int` to represent the power.

finesse/hardware/data_file_writer.py
alexdewar: Also, if `TC4820_MAX_POWER` isn't used anywhere outside of `tc4820.py` now, we should move it there.

finesse/hardware/temperature/tc4820.py
dc2917: Return a `Decimal` via
```python
return self.request_decimal("020000") * 100 / TC4820_MAX_POWER
```
or still request an int but return the value of the percentage conversion as a float?

finesse/hardware/temperature/tc4820.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, name: str, serial: Serial, max_attempts: int = 3) -> None:
 
         self.serial = serial
         self.max_attempts = max_attempts
+        self.max_power = 511
```
alexdewar, line 43: I'd probably leave this as a constant, but just define the value in this file rather than `config.py`.

alexdewar (approved):
One minor nit, but otherwise LGTM.

PR #359, opened 2023-10-24
Fix check for whether stepper motor is connected

### LINKED ISSUE #358: Buggy check for whether angle is `None`
See: https://github.com/ImperialCollegeLondon/FINESSE/blob/3bd14e6bac7a372166b1bdfc1a2963f0765719dc/finesse/hardware/data_file_writer.py#L63

An angle of 0 will also mean this expression is false, which is not what we want.

alexdewar:
This is only a small fix, but I thought I'd do it now before I forgot.

Fixes #358.


dc2917 (approved):
Seems logical

PR #360, opened 2023-10-26
Fix device error handling

alexdewar:
For now I've left #336 unmerged because I want to test it out with real hardware when I go into the lab on Monday. I have been using it for development though and noticed that I'd managed to break error handling for devices along the way, though I noticed some other error handling problems that existed before. As the changes I had to make to fix it have been fairly intrusive, I thought it deserved a separate review, so I've made this PR to be merged into the #336 branch.

The main problem was that I was sending device error messages in the wrong format so the frontend was no longer able to detect them. While I was fixing this I took the opportunity to consolidate some of the error handling code a bit more. In this PR I've added functionality to the `Device` class for subscribing to device-related topics and unsubscribing from them when `close()` is called (needed so that the object isn't accessed before the GC has the opportunity to delete it) along with some other helper functions.

I found another bug whereby if an error occurs in a handler for a `device.opened` message, then the buttons in the `DeviceControl` can remain disabled, preventing the user from trying again. This is because the buttons are also re-enabled with a `device.opened` message and depending on the order in which the listeners are invoked, this may not happen at the right point. To fix this I added a new message (`device.opening`) to be run before `device.opened`.


finesse/hardware/data_file_writer.py
```diff
@@ -65,11 +64,7 @@ def _get_stepper_motor_angle() -> tuple[float, bool]:
         else:
             return (float("nan"), True)
     except Exception as error:
-        pub.sendMessage(
-            f"device.error.{config.STEPPER_MOTOR_TOPIC}",
-            instance=DeviceInstanceRef(config.STEPPER_MOTOR_TOPIC),
-            error=error,
-        )
+        stepper.send_error_message(error)
```
dalonsoa, line 67: It feels a bit odd to handle _externally_ an error generated by doing something with a device by calling a method of the device itself. Would it make more sense to check that things are ok in the methods being called and, if not, call the `send_error_message` _internally_?

This applies to other devices using this same approach.

dalonsoa (approved):
This looks quite neat. I just have a comment, but not a big deal.

finesse/hardware/data_file_writer.py
alexdewar: I know what you mean -- it is a bit odd. It doesn't feel entirely right to catch the error inside the method either though, because we want to pass the error on to the caller. (The errors will be caught and retransmitted as a `pubsub` message if this value is requested via a `pubsub` message though.)

I feel in general this file needs another look at because I'm not convinced about the design. I tried rewriting it to use `pubsub` messages instead of importing these classes directly though and it ended up being up being horrible :cry:.

Issue #365, opened 2023-10-27
Allow for choosing a hardware "set" to use


alexdewar:
Currently the user can choose which devices to use along with their associated parameters from the GUI. There are a couple of mechanisms to help them select sensible values:

- For a given device base type (e.g. stepper motor), the device type and parameters for the last successful connection are remembered
- Default values corresponding to the correct baudrates for serial devices (well, correct for the FINESSE rig, anyway) are provided where no connection has been made

The problem is that we now need to be able to handle *at least* two different hardware configurations (for the separate FINESSE and UNIRAS rigs) and we don't want the user to have to manually select which hardware to use when changing between them. We could let the user select from several remembered configurations (i.e. as opposed to just remembering the last successful one). While technically we *might* be able to get away with just remembering the device *types*, it would be better to remember the device parameters too. If we do this, we should provide sensible default configurations for the FINESSE and UNIRAS rigs and make these read only.

This will probably require a new dialog class to allow users to manage these configurations.

As we are already using YAML for our measure script files, this is probably the most sensible format to use. Alternatively, you could use the `QSettings` module (which we're also using).

Bonus points if you allow for exporting the configuration files :wink:.

Issue #363, opened 2023-10-27
Modify OPUS (EM27) interface code to use plugin system


alexdewar:
See also #362. It probably makes sense to tackle #147 before this one.

Currently the plugin system only works with USB serial devices. The intent, though, is to use it to manage connections to all devices. The OPUS "device" is different from serial devices though:

- It uses an HTTP-based protocol, so connections are made and dropped for every request, rather than being made once at the start and dropped when the device is closed -- to know whether the spectrometer is "connected" or not you have to poll its status
- As for the EM27 scraper, we don't necessarily want errors to be fatal
- It has its own frontend GUI component

We may want to ping it regularly to check its status, though because it is slow (it takes ~2s per request), there is a limit to how often we can do it. (Bear in mind we also have to make other requests!)

Unlike the EM27 scraper, the frontend GUI component is quite device specific, so one option would be to make this GUI component a part of the plugin (which would require changes to the API). Alternatively, we could redesign it to be a bit more generic, in which case we could use it for the new spectrometer too (#322). As with #362, changes to the frontend could be implemented separately.

Frontend changes could be done alongside #190.

PR #375, opened 2023-10-31
Make EM27 sensors module into plugin

### LINKED ISSUE #362: Modify EM27 scraper code to use plugin system
Currently the plugin system only works with USB serial devices. The intent, though, is to use it to manage connections to all devices.

The EM27 scraper is different to serial devices in a number of respects though:

- Connections are only made when sensor readings are requested from the device and are then dropped again (cf. serial devices which are connected for the whole time they are alive)
- An error reading sensor values needn't be fatal: you can just try again when polling later
- The frontend includes a specific GUI component for this hardware

That said, I think it makes sense to keep the overall device API the same, but perhaps connect and disconnect could mean start/stop polling in this case. We should keep the error API the same for this device type too and maybe show the errors with a dialog/LED (see #246). Making errors non-fatal for this device will also require changing the plugin backend code (see `manage_devices.py`).

I propose that we leave the frontend component as is for now, as we also need a sensor panel for the aircraft sensors (#325) and it may be that we can reuse it there.

alexdewar:
This PR turns the EM27 sensors (web scraper) module into a plugin, like we have for the USB serial devices. This is part of the sprint goal to let the user choose between different hardware sets (#365); for a device to be part of a "set" it must be manageable in the same way as other devices (e.g. you can open and close it remotely with the relevant pubsub messages).

This required some changes to the way the device object works. Currently, requests for sensor data are sent from the frontend every 60s from the time the program starts; if an error occurs it is ignored. This is perfectly sensible, but we need users to be able to select the type of EM27 sensor device (i.e. real or dummy) after the program has started. Accordingly, I have refactored things so that the polling is done on the backend instead and the frontend is purely responsible for opening and closing the device. Errors are now handled by closing the device as we do for USB serial devices (though it may make sense to ignore errors as we did before, possibly notifying the user about it).

I've also put the OPUS-related code into the new `finesse.hardware.plugins.em27` module, but haven't converted this code to be a plugin yet.

Closes #362.


finesse/config.py
```diff
@@ -36,9 +36,15 @@
 EM27_URL = "http://10.10.0.1/diag_autom.htm"
 """The URL of the EM27 monitoring web server."""
 
-EM27_PROPERTY_POLL_INTERVAL = 60.0
+EM27_SENSORS_POLL_INTERVAL = 60.0
```
dalonsoa, line 39: Is this something we migth want configurable?

finesse/em27_info.py
dalonsoa, line 1: I'm not super fan of deeply nesting packages, but given that there's a package for `gui`, wouldn't it make sense to put all the hardware related stuff in a `hardware` package?

finesse/gui/measure_script/script_view.py
```diff
@@ -6,7 +6,7 @@
 from PySide6.QtWidgets import QFileDialog, QGridLayout, QGroupBox, QPushButton
 
 from ...config import DEFAULT_SCRIPT_PATH, STEPPER_MOTOR_TOPIC
-from ...em27_status import EM27Status
+from ...em27_info import EM27Status
 from ...event_counter import EventCounter
 from ...settings import settings
 from ..path_widget import OpenFileWidget
```
dalonsoa, line 12: We should have some consistency on importing internal stuff, either using relative imports like here or absolute ones, as you have done in other cases.

finesse/hardware/__init__.py
```diff
@@ -5,11 +5,9 @@
 from pubsub import pub
 
 if "--dummy-em27" in sys.argv:
-    from .dummy_em27_scraper import DummyEM27Scraper as EM27Scraper
-    from .opus.dummy import DummyOPUSInterface as OPUSInterface
+    from .plugins.em27.dummy_opus_interface import DummyOPUSInterface as OPUSInterface
```
dalonsoa, line 8: If the user is able (or will be able) to select the devices to use after starting the software, is this check still relevant. Indeed, is it relevant to read arguments from the command line anymore?

finesse/hardware/plugins/em27/em27_sensors.py
```diff
@@ -2,48 +2,15 @@
...
-
-        Returns:
-            str: The value and unit of a property in the format consistent with
-                 the previous FINESSE GUI.
-        """
-        return f"{self.value:.6f} {self.unit}"
+from finesse.config import EM27_SENSORS_POLL_INTERVAL, EM27_SENSORS_TOPIC, EM27_URL
+from finesse.em27_info import EM27Property
+from finesse.hardware.device import Device
+from finesse.hardware.http_requester import HTTPRequester
```
dalonsoa, line 13: See here? You are using absolute imports.

dalonsoa (approved):
Great work! I've a couple of stylistic suggestions, but looks good otherwise.

finesse/config.py
alexdewar: Yep! Atm you can only have provide a `list` of possible values for device params, but for this case it would be better to have a text box/spin box. I've added an issue for this (#377) and once that's done it'll be easier to implement.

finesse/em27_info.py
alexdewar: I'm not sure actually -- it would be good to know what you think. So atm I've been putting the various dataclasses that we pass between the frontend and backend that describe aspects of the hardware directly into the `finesse` module. Maybe I'm being too strict about the "no cross imports" rule?

finesse/gui/measure_script/script_view.py
alexdewar: True! I've opened an issue to do this project-wide (#378).

finesse/hardware/__init__.py
alexdewar: For now, we still need it, because this is how we choose between the real and dummy OPUS device. Once #363 is implemented we can remove it.

(Confusingly, there are two spectrometer-related "devices" in FINESSE: one to scrape sensor values and one to control recording with the spectrometer.)

dc2917 (approved):
Looks good to me

CWestICL (approved):
Looks all good on my end!

PR #376, opened 2023-10-31
Device frontend fixlets

### LINKED ISSUE #369: Baudrate argument to `Serial` constructor is passed as string
Device parameters are passed around in string format and seemingly I forgot to unstringify the baudrate before passing it to the `Serial` constructor.

While it seems to work anyway (on Linux, at least), we shouldn't rely on it.

### LINKED ISSUE #370: Device "open" button should be disabled if no USB ports available

alexdewar:
This PR fixes a few minor issues to do with the device management code. I started doing this as part of my work on #323, but figured there was enough here for a separate PR.

- Allow for having non-string device parameters (now used for baudrate, but we will also need it for serial device attributes)
- Refactor `DeviceTypeControl` to save relevant device-type info in the `QComboBox` along with the string label (I only just learned you can do this, but it makes things cleaner)
- Disable the "open" button if a device has any parameters for which there are no possible values given (this occurs if there are no COM ports on the current machine)

Fixes #369. Fixes #370.

### DIFF
--- a/finesse/device_info.py
+++ b/finesse/device_info.py
@@ -14,8 +14,10 @@ class DeviceParameter:
     possible_values: Sequence[Any]
     """Possible values the parameter can take."""
 
-    default_value: Any | None = None
-    """The default value for this parameter."""
+    default_value: Any = None
+    """The default value for this parameter.
+
+    A value of None indicates that there is no default value."""
 
     def __post_init__(self) -> None:
         """Check that default value is valid."""
--- a/finesse/device_info.py
+++ b/finesse/device_info.py
@@ -1,6 +1,7 @@
 """Provides common dataclasses about devices for using in backend and frontend."""
 from collections.abc import Sequence
 from dataclasses import dataclass
+from typing import Any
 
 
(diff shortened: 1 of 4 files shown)


finesse/device_info.py
```diff
@@ -10,10 +11,10 @@ class DeviceParameter:
     name: str
     """Name for the parameter."""
 
-    possible_values: list[str]
+    possible_values: Sequence[Any]
     """Possible values the parameter can take."""
 
-    default_value: str | None = None
+    default_value: Any | None = None
```
dalonsoa, line 17: This feels odd. Doesn't `Any` include `None`?

dalonsoa (approved):
Looks good to me!

finesse/device_info.py
alexdewar: True! I'll fix it up

dc2917 (approved):
Looks good, also never heard the term fixlet before 😄

alexdewar:
It's probs not the right term here either. This PR isn't particularly well named :laughing:.

PR #379, opened 2023-11-01
Add developer documentation for hardware framework

### LINKED ISSUE #366: Write developer documentation for core hardware code
The core hardware module is already somewhat complicated, but since #336 it has become more so (albeit a bit cleaner!).

As this is potentially a barrier to contributing, the developer documentation for the hardware code should be improved. This could be done either with longer module docstrings or with separate markdown files (or both).

It should include:

- Explaining concepts like device base type and device type
- Explaining how to make a new plugin (including making new base types and regular device types)
- Better API documentation for the current base types
- Descriptions of the `pubsub` messages that are used
- A better module description for `data_file_writer.py` (possibly, depending on whether we're going to refactor it or not)

alexdewar:
This potentially could be more detailed, but hopefully there's enough information in here for would-be contributors to get started.

@CWestICL: Would you mind reviewing this and checking that it makes sense to you? You're currently the target audience for it :wink:

Closes #366.

### DIFF
--- a/docs/hardware.md
+++ b/docs/hardware.md
@@ -79,7 +79,7 @@ Many messages for communicating with devices include a string indicating which d
 the communication is intended for. This is composed of the device base type's name and,
 if provided, the device's name. For example, this could be `stepper_motor` for the
 stepper motor and `temperature_controller.hot_bb` for the hot black body temperature
-monitor.
+controller.
 
 When the main program window has loaded, the plugins are dynamically loaded and
 information about each device type (grouped by base type) is sent to the frontend with
--- a/docs/hardware.md
+++ b/docs/hardware.md
@@ -52,38 +52,72 @@ for instances of the device, but this is currently only used for temperature con
 You can create a concrete implementation of `MyBaseType` like so:
 
 ```py
-class MyDevice(MyBaseType, description="An example device"):
+class MyDevice1(MyBaseType, description="An example device"):
     # ...
 ```
 
-Certain device base types require extra parameters. For example, to create a USB serial
-device type which also inherits from `MyBaseType`, you can do:
+You may optionally provide a `dict` specifying which parameters should be passed when
+the device object is constructed, along with a human-readable description. This provides
+a mechanism by which the frontend can know what parameters it can provide for a given
+device type, as well as information about their type and default value (if any). Here is
+an example:
 
 ```py
-class MySerialDevice(
-    SerialDevice, MyBaseType, description="An example device", default_baudrate=9600
+class MyDevice2(
+    MyBaseType,
+    description="An example device",
+    parameters={"my_param": "An example parameter"}
 ):
-    # ...
+    def __init__(self, my_param: int = 42) -> None:
+        # ...
 ```
 
-### Device parameters
+In this case, the frontend will be informed that `MyDevice2` has a parameter, `my_param`
+of type `int` with a default value of `42`. The user can then alter this value via a
+text box in the GUI. Note that if a default value were *not* provided for this
+parameter, the user would be forced to enter one. Subclasses inherit their parents'
+device parameters (but can add more of their own). As a result, they *must* also include
+these parameters for their constructors.
 
-Devices can also require additional parameters to be specified by the user, such as port
-and baudrate for serial devices. Details about these parameters are set in an
-`__init_subclass__()` method for the class. These parameters will be passed as arguments
-to the device type's constructor when it is opened.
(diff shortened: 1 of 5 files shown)


docs/hardware.md
```diff
@@ -0,0 +1,98 @@
...
+class MyDevice(MyBaseType, description="An example device"):
+    # ...
+```
+
+Certain device base types require extra parameters. For example, to create a USB serial
+device type which also inherits from `MyBaseType`, you can do:
+
+```py
+class MySerialDevice(
+    SerialDevice, MyBaseType, description="An example device", default_baudrate=9600
```
dalonsoa, line 64: Is the order in which the superclasses are specified important?

docs/hardware.md
```diff
@@ -0,0 +1,98 @@
...
+):
+    # ...
+```
+
+### Device parameters
+
+Devices can also require additional parameters to be specified by the user, such as port
+and baudrate for serial devices. Details about these parameters are set in an
+`__init_subclass__()` method for the class. These parameters will be passed as arguments
+to the device type's constructor when it is opened.
```
dalonsoa, line 74: This is very unclear - or at least ambiguous. Are these parameters passed to the `__init_subclass__` method, and therefore need to be included in the subclass definition, or are they pass when constructing the object? And example, as you have done above, will be useful.

docs/hardware.md
```diff
@@ -0,0 +1,98 @@
...
+which device type should be opened, along with any device parameters. If the connection
+is successful, a `device.opening.*` message is sent, followed by a `device.opened.*`
+one. If the connection fails, a `device.error.*` message is sent instead.
+(`device.error.*` messages can also be sent at any point during the device's lifetime to
+indicate that an error has occurred.) Similarly, the `device.close` method is used to
+close a connection to a device.
+
+Device types also need to define their own message types for communication. For example,
+the `StepperMotorBase` class allows for setting the current angle of the stepper motor
+with a `device.stepper_motor.move.begin` message.
```
dalonsoa, line 98: It will be useful, I think, to indicate somewhere the list of all possible messages a particular device type handles - either sending or receiving. In the end, pubsub topics are implicit function calls with their own parameters and they should be documented in detail as we do with any other function.

We have never done it before, but given the complexity of the messaging system of FINESSE, it might be worth considering doing something like this https://pypubsub.readthedocs.io/en/v4.0.3/usage/usage_advanced_maintain.html#specify-topic-tree-def, although as most topics are created implicitely based on the device names, I'm not sure how far we can go down this route...

dalonsoa (approved):
This is really good. I've pointed a couple of things that could be polished, but having this documentation will make life much easier to future developers, including ourselves!

CWestICL (approved):
Other than some of the points that Diego has brought up, it all makes sense to me! Don't think I have any other changes to add.

docs/hardware.md
alexdewar: Good point. In this specific case, yes, because `SerialDevice` has to come first (it defines a `close()` method which is  required by the abstract `Device` class). I'll update this

docs/hardware.md
alexdewar: Makes sense.

The doc you link to looks interesting. I'll see if we can incorporate that way of doing things :smile:

dc2917 (approved):
One tiny, single-word change but otherwise this looks good! Very informative and concise.

docs/hardware.md
```diff
@@ -0,0 +1,98 @@
...
+`__init_subclass__()` method for the class. These parameters will be passed as arguments
+to the device type's constructor when it is opened.
+
+### Communicating with devices via PyPubSub
+
+Many messages for communicating with devices include a string indicating which device
+the communication is intended for. This is composed of the device base type's name and,
+if provided, the device's name. For example, this could be `stepper_motor` for the
+stepper motor and `temperature_controller.hot_bb` for the hot black body temperature
+monitor.
```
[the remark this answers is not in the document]
alexdewar: Good spot!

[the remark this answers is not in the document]
alexdewar: My brain seems to swap "temperature monitor" and "temperature controller" seemingly at random :laughing:

docs/hardware.md
alexdewar: Done now

docs/hardware.md
alexdewar: This is a good idea, but I think we should do it later. I've opened an issue for it: #432

PR #381, opened 2023-11-02
Use persistent attributes to distinguish USB serial devices

### LINKED ISSUE #323: Remember USB devices based on persistent attributes cf. COM port numbers

alexdewar:
Currently, devices are distinguished based on their COM-port number (or the Unix equivalent), which is not guaranteed to be constitent between runs. For example, on Windows the COM-port number is fixed to specific USB ports for a given PC.

Change the USB serial code to use various persistent attributes of devices instead: viz. vendor ID, product ID and serial number (which is sometimes, but not always present). Devices which provide multiple serial ports (of which we are using one for FINESSE at present) will yield ports with the same attributes, so an additional "count" field is also used to distinguish devices in this case. It appears in this latter case that the serial ports will be in the same order for a given device, so this should be robust (e.g. if the stepper motor is the second port for this device, it will be always be given a higher COM-port number than the first).

I haven't tried this with real hardware yet (though I did hack the code so I could test parts of it without), but I wanted to get this reviewed before I try it out (hopefully on Monday).

Closes #323.

### DIFF
--- a/finesse/hardware/serial_device.py
+++ b/finesse/hardware/serial_device.py
@@ -1,9 +1,10 @@
 """Provides a base class for USB serial devices."""
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import Any
 
-from serial import Serial
+from serial import Serial, SerialException
 from serial.tools.list_ports import comports
 
 from finesse.config import BAUDRATES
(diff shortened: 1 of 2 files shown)


finesse/hardware/serial_device.py
```diff
@@ -1,20 +1,43 @@
...
+from serial import Serial, SerialException
 from serial.tools.list_ports import comports
 
 from finesse.config import BAUDRATES
 from finesse.device_info import DeviceParameter
 
 from .device import AbstractDevice
 
-_serial_ports: list[str] | None = None
+_serial_ports: dict[_USBSerialPortInfo, str] | None = None
```
dalonsoa, line 15: I cannot find when this is populated. You mention it is done once when opening the program.. but where?

dalonsoa (approved):
Looks good, but I think I'm missing something... Maybe it was already done and not in this PR, that's why I'm missing it :P

finesse/hardware/serial_device.py
```diff
@@ -1,20 +1,43 @@
...
+    def __str__(self) -> str:
+        out = f"{self.vendor_id:04x}:{self.product_id:04x}"
+        if self.serial_number:
+            out += f" {self.serial_number}"
+        if self.count > 0:
+            out += f" ({self.count+1})"
+        return out
+
+
+def _get_usb_serial_ports() -> dict[_USBSerialPortInfo, str]:
```
alexdewar, line 40: @dalonsoa This is where it's populated. The result is cached in `_serial_ports` so the ports are only scanned once.

dc2917 (approved):

PR #383, opened 2023-11-02
pyinstaller: Bundle all hardware plugins with executable

### LINKED ISSUE #382: `pyinstaller` not bundling hardware plugins
This makes the generated executables unusable as no hardware plugins will be loaded by FINESSE. We need to tell `pyinstaller` explicitly about all the hardware plugin modules as it cannot autodetect them given that they are loaded dynamically.

alexdewar:
It turns out that #336 broke the generated binaries for FINESSE. The problem is that the hardware modules are now loaded dynamically via the plugin system and so cannot be autodetected by `pyinstaller`. The solution is to add them to the `hiddenimports` in the spec file.

To implement this, I modified the `load_all_plugins()` function so that it returns a `list` of loaded plugins, so it can be used by the spec file.

Fixes #382.

### DIFF
--- a/FINESSE.spec
+++ b/FINESSE.spec
@@ -1,18 +1,21 @@
 # -*- mode: python ; coding: utf-8 -*-
 import docs.gen_user_guide as gen_guide
+from finesse.hardware.plugins import load_all_plugins
 
 block_cipher = None
 
 gen_guide.generate_html()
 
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
Neat!

dc2917 (approved):

PR #404, opened 2023-11-13
Make all imports absolute

### LINKED ISSUE #378: Convert all imports to be absolute
Currently we use a mix of absolute and relative imports and it's kinda untidy. Let's be consistent and make them all absolute.

There's a tool that might help with this: https://github.com/MarcoGorelli/absolufy-imports

alexdewar:
Convert all imports to be absolute. Closes #378.

I used this tool to do this: https://github.com/MarcoGorelli/absolufy-imports

It is also available as a pre-commit hook, so I've added it. The project is unfortunately archived but it still seems to work. We can drop the hook if it presents problems at some stage.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,3 +37,7 @@ repos:
     hooks:
       - id: markdownlint
         args: ["--disable", "MD013", "MD041", "--"]
+  - repo: https://github.com/MarcoGorelli/absolufy-imports
+    rev: v0.3.1
+    hooks:
+      - id: absolufy-imports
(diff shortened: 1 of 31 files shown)


dalonsoa (approved):
LGTM! Do you think there is any use case (thinking in general, not finesse specifically) where relative imports are more appropriate or even necessary?

alexdewar:
Hmm... I'm not sure. I feel like it's probably fine for imports in the same package (e.g. `from .subpackage import thing`), but it does get confusing if you start having `from ....package import other_thing`. It's maybe just easier to have a blanket rule for it though.

Issue #396, opened 2023-11-08
GUI should show EM27 status in more user-friendly form


alexdewar:
Currently, the only way to check what the device is doing and whether an error has occurred is to click the "Status" button and look at the (user-unfriendly) log.

It would be better:

1. To show the status visually (with a label, LED etc.)
2. To poll the device automatically rather than making the user do it manually.

PR #413, opened 2023-11-14
Manual device management dialog

### LINKED ISSUE #395: Move manual device configuration to separate dialog
Currently, this functionality is supplied by a panel in the main GUI (`DeviceTypeControl`). Once #392 is merged though, it will be possible to manage devices in bulk as "hardware sets". At this point, it will be possible to move the manual opening/closing/configuring of devices out of the main GUI into a separate dialog.

This will remove clutter from the main window as well as options (such as baudrate) which may confuse non-expert users. It will also mean that we don't need to unconditionally load all hardware plugins when the program opens (currently needed to populate `DeviceTypeControl`).

### LINKED ISSUE #397: `DeviceControl`: Open/close buttons stay disabled when hardware sets connect
Since #392, if the open/close button is disabled for a given device type (e.g. because it is a serial device and there are no serial devices connected), it will stay disabled when a hardware set using the type is connected, meaning the user cannot disconnect from this device manually (though the "Disconnect all" button can still be used).

Example:

- The DP9800 temperature monitor is selected in the combo box, but there are no serial devices connected, so the open/close button is disabled
- The user connects to a hardware set which uses the `DummyTemperatureMonitor` class instead
- The open/close button stays disabled, so the user cannot manually disconnect from the `DummyTemperatureMonitor` using this button

There are a couple of solutions to this:

1. A hack: Unconditionally enable the button when a hardware set connects
2. Change the selection in the combo box automatically when the user connects to the hardware set (and also the device parameters, ideally), which will also have the effect of enabling the button. This will also show the user visually which device types they have connected to

My preference should be obvious :wink:

### LINKED ISSUE #398: Add tests for `DeviceControl`

### LINKED ISSUE #405: Window does not fit on screen
The current layout and sizing of widgets prevents the entire window from fitting on my screen, and cannot be resized to make the window smaller due (presumably) to the size policies. Probably the matplotlib figure is responsible here.

The window cannot be maximised.

Also, minimizing the window and re-opening it causes a further change in sizing.

alexdewar:
This PR moves the panel used for opening/closing connections to devices into a separate dialog (#395) and fixes a small associated bug (#397). This makes the main window less cluttered (see #405) and also means that we no longer have to load all plugins unless the user is using the manual device management dialog. Otherwise, the plugins will just be loaded as needed.

As this is (probably) the end of my tinkering with the `DeviceControl` class, I've added tests for it (#398). The coverage for the `hardware_sets` module is at 100% which takes us to 80% coverage overall :partying_face: 

Closes #395. Closes #397. Closes #398. Closes #405.


dalonsoa (commented):
It looks good! Anything that declutters the main window is a step in the right direction. This looks a bit odd, though, and I'm not sure if its behaviour is the expected one.

The following screenshot is FINESSE just after opening it:
image

Note the odd dangling POLL Server light in the EM27 Monitor, with nothing else in that area. Then, I connect the (dummy) EM27 monitor in the brand new dialog, and I get a bunch of readings, displacing shuffling a bit the layout (specially the log, which get's squeeze):

image

Finally, I close the EM27 monitor in the dialog. The sensor readings remain, but greyed out. 

image

Is this behaviour intentional? To be honest, I'm not entirely sure what I was expecting, but maybe some sort of symmetric behavior when connecting/disconnecting, the POLL light flushed to the bottom of the area and a more... maybe "rigid" layout, so adding/removing widgets do not upset the GUI?

As I say, I'm not sure what I was expecting and therefore I cannot advise what to change.

alexdewar:
@dalonsoa It's less that it was intentional and more that that's just how it's ended up :stuck_out_tongue:.

I think there are three related problems here:

1. There are too many controls on the right cf. the left (maybe the data file controls could be moved to the left?)
2. The EM27 sensor panel has the issues that you mention. They've only really become apparent since I made it so that you have to connect to the EM27 sensors explicitly (#375). Maybe the poll light should also be invisible while the control is disabled?
3. The OPUS panel also needs reorganising. The log isn't particularly useful so the whole panel could just consistent of buttons, maybe with a much smaller control displaying the current status (#396)

See also: #405

How about I open separate issues for these and tackle them later?

dalonsoa:
That's totally fine with me :)

alexdewar:
We could also add to the list: 4. The temperature plot on the right is too big

dc2917:
Trying to position the POLL Server indicator & label sensibly was an absolute pain - it's definitely not positioned as intended! It bugs me every time I open the window but yeah, from what I recall it's not straightforward to fix.

finesse/gui/hardware_set/device_view.py
```diff
@@ -0,0 +1,317 @@
...
+    @property
+    def current_parameter_values(self) -> dict[str, Any]:
+        """Get all parameters and their current values."""
+        return {param: combo.currentData() for param, combo in self._combos.items()}
+
+
+class DeviceTypeControl(QGroupBox):
+    """A set of widgets for choosing a device and its params and connecting to it."""
+
+    def __init__(
```
dalonsoa, line 92: This constructor is quite long. It might make sense to break it down into smaller chunks, roughly matching your commented sections.

dalonsoa (approved):
It looks good - nice collection of tests! I just have a minor comment you can ignore. 

Having said that, the comments I made before on the layout persist. Did you mention you were going to open an issue about beautifying this?

alexdewar:
Thanks for the review :smile:. I've opened two more issues about the GUI's appearance: #419 and #420

finesse/gui/hardware_set/device_view.py
alexdewar: Good point. As I've since modified this code on newer branches, I'll open an issue for it and fix it later.

PR #416, opened 2023-11-15
TemperatureSequence alias for temperature monitors

### LINKED ISSUE #401: TemperatureSequence alias
Currently the API for `TemperatureMonitorBase` only accepts a `list[Decimal]` as the return type for the `get_temperatures()` abstract method. This should be changed to accept other data types such as numpy arrays of `np.float64`s.

For the specific temperature monitor device classes, all the `get_temperature()` f.unctions should have their return types changed to `Sequence[Decimal | np.float64]` or something similar. This could be defined as an alias `TemperatureSequence`.

CWestICL:
Added a typing alias called `TemperatureSequence` that is the type `list[Decimal | numpy.float64]`. This is now the return type for the `get_temperatures()` function in `temperature_monitor_base.py`, `senecak107.py` and `dp9800.py`.

Closes #401

### DIFF
--- a/finesse/hardware/plugins/temperature/dp9800.py
+++ b/finesse/hardware/plugins/temperature/dp9800.py
@@ -6,7 +6,7 @@ from serial import SerialException
 
 from finesse.hardware.serial_device import SerialDevice
 
-from .temperature_monitor_base import TemperatureMonitorBase
+from .temperature_monitor_base import TemperatureMonitorBase, TemperatureSequence
 
 
 def check_data(data: bytes) -> None:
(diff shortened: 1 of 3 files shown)


alexdewar (changes requested):
Looks good!

There's just one more change I've noticed that needs to be made on the frontend. Atm, the `_update_figure()` method in `temp_control.py` assumes its inputs are `Decimal`s, which is no longer true. Would you mind converting these arguments to be `float`s please? You'll also need to update the call site to cast the temperatures to `float`s too.

alexdewar (approved):
LGTM! :smile:

PR #418, opened 2023-11-16
Fix type errors in `manage_devices.py`

### LINKED ISSUE #417: Pre-commit broken: mypy issue in `manage_devices.py`

alexdewar:
The latest round of auto-updates to the `pre-commit` hooks and dependencies seemed to break things and unfortunately didn't trip the CI system until after they were all merged.

The problem is that we were applying the `|` operator to a `Mapping`, which isn't defined. I've reworked the code to avoid this.

I also spotted another typing mistake in the same function (`Device` => `type[Device]`), which I fixed while I was at it.

Fixes #417.

### DIFF
--- a/finesse/hardware/manage_devices.py
+++ b/finesse/hardware/manage_devices.py
@@ -51,13 +51,12 @@ def _open_device(
 
     # If this instance also has a name (e.g. "hot_bb") then we also need to pass this as
     # an argument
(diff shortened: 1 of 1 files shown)


dalonsoa (approved):
LGTM!

CWestICL (approved):
Looks good on my end!

PR #427, opened 2023-11-21
Use shortened module name for device classes

### LINKED ISSUE #403: Use terser version of class names for devices
Currently, the class names for devices (e.g. in `DeviceTypeInfo`) include the full module name (e.g. `finesse.hardware.plugins.whatever.MyDevice`). This makes config files unnecessarily verbose and means that if we move files around (e.g. because the project's name changes: #352) the config files will break. While at least part of the module name is needed (we import plugin modules lazily as needed), we could at least drop the `finesse.hardware.plugins` prefix.

alexdewar:
This makes config files less verbose and means if/when we change the name of the main package, things won't break.

Closes #403.

### DIFF
--- a/finesse/gui/hardware_set/finesse_dp9800.yaml
+++ b/finesse/gui/hardware_set/finesse_dp9800.yaml
@@ -1,24 +1,24 @@
 name: FINESSE (with DP9800)
 devices:
   stepper_motor:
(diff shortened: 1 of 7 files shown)


alexdewar:
Superseded by #428

Issue #335, opened 2023-10-04
Allow for refreshing list of COM ports


alexdewar:
Currently, the list of COM ports is requested from the OS once when the program starts and any changes to this list (i.e. connecting and disconnecting devices) will not be detected by FINESSE. Auto-detecting (dis)connection of devices will likely be difficult, but there could be a "Refresh" button for the user to click on. The alternative is that the whole program has to be restarted in order to detect newly connected devices.

alexdewar:
Since #472 this is less pressing.

PR #428, opened 2023-11-21
Add support for typed device parameters

### LINKED ISSUE #377: Allow for having possible values for device params that include types (e.g. `float` and `str`)
Currently, the possible values specified for device params must be a `list`; the frontend then puts these possible values in a dropdown box for the user to select. However, there are other aspects of device behaviour that the user might want to configure that don't fit this pattern: for example, destination IP address and poll interval.

At a minimum, we should add the option for the user to enter a device parameter in a text box, though we may also want the option to use other controls (e.g. a spin box for selecting a number).

### LINKED ISSUE #403: Use terser version of class names for devices

alexdewar:
This PR adds support for specifying arbitrary device parameters via a text box, rather than having to choose from a set list in a combo box (#377).

The changes to the frontend are fairly minimal. The device management dialog will now display text boxes or combo boxes for each of the parameters, depending on which is appropriate. While there are no labels saying what each parameter is, a description is provided in a tooltip so the user can find out by hovering.

The backend changes were a bit more invasive. I refactored the device plugin code, so now the user just has to specify the names of the parameters along with a description and, optionally, the list of possible values.

For example, the signature for the `EM27Sensors` class now looks like this:

```py
class EM27Sensors(
    EM27SensorsBase,
    description="EM27 sensors",
    parameters={
        "host": "The IP address or hostname of the EM27 device",
        "poll_interval": "How often to poll the device in seconds",
    },
):
```

(Both of these are typed device params, i.e. their values can be arbitrary.)

The types and default values are automatically worked out from the constructor's signature:

```py
    def __init__(
        self, host: str = EM27_HOST, poll_interval: float = EM27_SENSORS_POLL_INTERVAL
    ) -> None:
```

If a child class changes/adds a default value for these params in its own constructor, then these will be used instead. This feature was added with `SerialDevice`s in mind: it now means that devices can specify a default baudrate in their constructors, without having to provide an extra `default_baudrate` class argument, as it was before.

(I also implemented #403 at the same time.)

Closes #377. Closes #403.

### DIFF
--- a/finesse/hardware/serial_device.py
+++ b/finesse/hardware/serial_device.py
@@ -83,8 +83,8 @@ class SerialDevice(AbstractDevice):
 
         # Extra, serial-specific parameters
         cls.add_device_parameters(
-            DeviceParameter("port", list(_get_usb_serial_ports().keys())),
-            DeviceParameter("baudrate", BAUDRATES, default_baudrate),
+            port=DeviceParameter(tuple(_get_usb_serial_ports().keys())),
+            baudrate=DeviceParameter(BAUDRATES, default_baudrate),
         )
 
     def __init__(self, port: str, baudrate: int) -> None:
--- a/finesse/hardware/serial_device.py
+++ b/finesse/hardware/serial_device.py
@@ -1,13 +1,10 @@
 """Provides a base class for USB serial devices."""
 from __future__ import annotations
 
-from typing import Any
-
 from serial import Serial, SerialException
 from serial.tools.list_ports import comports
 
 from finesse.config import BAUDRATES
-from finesse.device_info import DeviceParameter
 from finesse.hardware.device import AbstractDevice
 
 _serial_ports: dict[str, str] | None = None
@@ -64,7 +61,16 @@ def _get_usb_serial_ports() -> dict[str, str]:
     return _serial_ports
 
 
-class SerialDevice(AbstractDevice):
+class SerialDevice(
+    AbstractDevice,
+    parameters={
+        "port": (
+            "USB port, including vendor ID, product ID and serial number",
+            tuple(_get_usb_serial_ports().keys()),
+        ),
+        "baudrate": ("Baud rate", BAUDRATES),
+    },
+):
     """A base class for USB serial devices.
 
     Note that it is not sufficient for a device type class to inherit from this class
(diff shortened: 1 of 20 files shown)


alexdewar:
@CWestICL This feature could be used for the `SenecaK107` class, to let the user set the voltage and temperature ranges, which would be handy for Jon. I'll open an issue for it once this PR is ready to go

finesse/hardware/serial_device.py
```diff
@@ -64,7 +61,16 @@ def _get_usb_serial_ports() -> dict[str, str]:
...
+class SerialDevice(
+    AbstractDevice,
+    parameters={
+        "port": (
+            "USB port, including vendor ID, product ID and serial number",
+            tuple(_get_usb_serial_ports().keys()),
+        ),
+        "baudrate": ("Baud rate", BAUDRATES),
+    },
+):
```
dalonsoa, line 73: I think you have mentioned this before, but I do not remember the answer:

This will retrieve the available USB ports that have "something" connected to them when this module is loaded, right? So, if the device of interest is off because we forgot to switch it on, then it will not be found. If we switch on the device, the collection of available USB ports will not be updated, so it will not be possible to connect to the device, anyway, the only solution being to restart the software. 

Is this correct or am I misunderstanding anything? Nor that you need to change anything, just checking what the protocol is in that, very common, situation.

dalonsoa (approved):
This is really neat! Just a question, but nothing to change.

finesse/hardware/serial_device.py
alexdewar: Correct! This is the same as the existing behaviour and there is already an open issue: #335 

I have an idea of how to fix it: the frontend could send a message to force a refresh of the COM ports before re-requesting the list of plugins. But it's more of a "could have" feature, so I haven't looked at it yet.

PR #445, opened 2023-11-29
Seneca Unit Tests

### LINKED ISSUE #399: Seneca K107 unit tests
Unit tests need to be written for the `SenecaK107` class temperature monitor device.

CWestICL:
Added `test_senecaK107.py` to `tests\hardware\plugins\temperature`.

Also made small change to return value of `SenecaK107.parse_data()` as it was previously trying to change a read-only numpy array.

Closes #399

### DIFF
--- a/finesse/hardware/plugins/temperature/senecak107.py
+++ b/finesse/hardware/plugins/temperature/senecak107.py
@@ -99,7 +99,7 @@ class SenecaK107(SerialDevice, TemperatureMonitorBase, description="Seneca K107"
         except Exception as e:
             raise SenecaK107Error(e)
 
-    def parse_data(self, data: bytes) -> list[numpy.float64]:
+    def parse_data(self, data: bytes) -> numpy.ndarray:
         """Parse temperature data read from the SenecaK107.
 
         The sequence of bytes is put through the conversion function and translated
@@ -117,7 +117,7 @@ class SenecaK107(SerialDevice, TemperatureMonitorBase, description="Seneca K107"
         # Converts incoming bytes into 16-bit ints
         ints = numpy.frombuffer(data, dt, 8, 3)
 
-        return [self.calc_temp(val) for val in ints]
+        return self.calc_temp(ints)
 
     def calc_temp(self, vals: numpy.ndarray) -> numpy.ndarray:
         """Convert data read from the SenecaK107 device into temperatures.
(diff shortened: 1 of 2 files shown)


finesse/hardware/plugins/temperature/senecak107.py
```diff
@@ -117,8 +117,7 @@ def parse_data(self, data: bytes) -> numpy.ndarray:
         # Converts incoming bytes into 16-bit ints
         ints = numpy.frombuffer(data, dt, 8, 3)
 
-        vals = self.calc_temp(ints)
-        return vals
+        return [self.calc_temp(val) for val in ints]
```
alexdewar, line 120: This isn't the best way to make a new data structure here. It'd be better to keep the data in a numpy array. Also, the type hints are now wrong for `calc_temp()` although it will still work -- not sure why `mypy` isn't flagging it.

Instead, I'd change `calc_temp()` to assign its result to a new variable, rather than overwriting `vals`.

alexdewar: I just noticed that the tests are failing and this is why.

tests/hardware/plugins/temperature/test_senecaK107.py
```diff
@@ -0,0 +1,89 @@
...
+    assert parsed == [
+        19.946250000000006,
+        20.085000000000008,
+        631.4406250000001,
+        631.4175,
+        20.14281249999999,
+        631.4406250000001,
+        19.946250000000006,
+        631.4406250000001,
+    ]
```
alexdewar, line 89: You have to be careful when comparing floating point values. Sometimes values don't exactly match because of small rounding errors. Annoyingly, you sometimes get subtly different results on different CPUs (e.g. Intel vs. AMD vs. ARM).

Here, I'd compare them using `numpy.allclose`.

alexdewar (changes requested):
There are a couple of things I'd change, but on the whole it's looking good :smile:

alexdewar (changes requested):
Looking good!

I found a couple of places where you're missing test coverage though. Would you mind having a go at writing tests for these bits too? Let me know if you get stuck.

Screenshot from 2023-11-29 at 16_22_58 013185334

Screenshot from 2023-11-29 at 16_23_12 506422890

CWestICL:
Weird, the `test_read_error()` test should be covering the `except` statement in the `.read()` function. Not sure why its not showing in the coverage.

alexdewar:
You've covered one of the `except` blocks but not the other. You need to test what happens when `read()` raises a `SerialException`

alexdewar (approved):
LGTM :smile:

PR #446, opened 2023-11-29
Show fallback html page if user guide has not been generated

### LINKED ISSUE #435: Uncaught error thrown if loading user guide when it hasn't been generated
This is *only* a problem when running FINESSE directly from the repo; the user guide is bundled with the `pyinstaller`-generated executable.

It only affects developers, so it isn't the most important problem, but it would be nicer to display an error message in a popup box instead. The error message should tell developers that they need to run the `docs/gen_user_guide.py` script to generate the user guide HTML file(s).

dc2917:
This PR adds a new html document, `docs/fallback.html` to be shown if `docs/user_guide.html` cannot be found - the most likely scenario being that it hasn't been generated.

I thought this might be a nicer approach than opening a popup dialog, but that's just me. Happy to go down the dialog route if others prefer.

Closes #435

### DIFF
--- a/FINESSE.spec
+++ b/FINESSE.spec
@@ -15,6 +15,7 @@ a = Analysis(
         ("finesse/gui/images/*.png", "finesse/gui/images"),
         ("finesse/hardware/plugins/em27/diag_autom.htm", "finesse/hardware/plugins/em27"),
         ("docs/user_guide.html", "docs"),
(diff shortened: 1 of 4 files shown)


dalonsoa (changes requested):
I agree with the idea, but not so much with the actual message being displayed.

dalonsoa (approved):

Issue #393, opened 2023-11-08
Make hardware sets configurable via dialog


alexdewar:
It would be useful to let users define their own hardware sets (possibly basing them on the default FINESSE/UNIRAS configuration settings). The problem with only providing built-in configurations is that if a single property changes (e.g. the user changes the type of temperature monitor used in the rig or wants to modify a device parameter), it will not be able to incorporate these changes into a hardware set without modifying FINESSE's source code.

Some of this code can likely be cannibalised from the existing `DeviceControl` implementation.

alexdewar:
NB: This task will probably take a while.

PR #447, opened 2023-11-30
Split combo box code out from `HardwareSetsControl`

alexdewar:
This PR contains some refactoring I did as a precursor to #393. Although we've decided to park that issue for now, I think this refactoring makes sense by itself and didn't want to lose the work already done on it.


alexdewar:
It was mostly just rearranging existing code. Sorry. I should have made that clearer.

PR #450, opened 2023-11-30
Unify opus message types

### LINKED ISSUE #147: Unify different OPUS error message types
Currently there are two ways the `OPUSInterface` has for communicating errors. When an exception is caught by the backend, it results in an `"opus.error"` message but when the device reports an error code in response to a request, this error information is instead just returned as part of the normal `"opus.response.{command}"` message.

It may be cleaner to report *all* errors through the `"opus.error"` message and only issue an `"opus.response.{command}"` message in case of success. Otherwise all handlers of OPUS responses also have to do error checking too.

alexdewar:
This PR removes the `error` arg from `opus.response` messages. Instead, all errors are now broadcast via the `opus.error` topic. This is a preliminary step towards making the spectrometer interface more generic.

I think the fact that this PR removes lines shows that this is a cleaner approach!

Closes #147.

### DIFF
--- a/finesse/gui/measure_script/script.py
+++ b/finesse/gui/measure_script/script.py
@@ -274,7 +274,7 @@ class ScriptRunner(StateMachine):
         )
 
         # EM27 messages
-        pub.unsubscribe(self._measuring_error, "opus.error")
+        pub.unsubscribe(self._on_em27_error, "opus.error")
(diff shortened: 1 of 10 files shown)


dalonsoa (approved):
Way cleaner, indeed!

PR #451, opened 2023-11-30
Convert OPUS devices to plugins

### LINKED ISSUE #363: Modify OPUS (EM27) interface code to use plugin system

alexdewar:
This completes the work of converting the device classes to plugins :partying_face:.

As I said in the meeting this morn, this PR just converts the existing OPUS (==EM27 spectrometer) code into a plugin, but there is more refactoring needed before we can get to a generic spectrometer interface (#459).

Closes #363.

### DIFF
--- a/finesse/hardware/plugins/em27/dummy_opus_interface.py
+++ b/finesse/hardware/plugins/em27/dummy_opus_interface.py
@@ -167,6 +167,8 @@ class DummyOPUSInterface(
 
         Args:
             command: The command to run
+        Raises:
+            OPUSError: If the device is in the wrong state for this command
         """
         if command == "status":
             if self.state_machine.current_state == OPUSStateMachine.idle:
--- a/finesse/hardware/plugins/em27/dummy_opus_interface.py
+++ b/finesse/hardware/plugins/em27/dummy_opus_interface.py
@@ -113,7 +113,11 @@ class OPUSStateMachine(StateMachine):
         logging.info(f"Current state: {target.name}")
 
 
-class DummyOPUSInterface(OPUSInterfaceBase):
+class DummyOPUSInterface(
+    OPUSInterfaceBase,
+    description="Dummy OPUS device",
+    parameters={"measure_duration": "Measurement duration in seconds"},
+):
     """A mock version of the OPUS API for testing purposes."""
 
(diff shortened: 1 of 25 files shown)


finesse/hardware/plugins/em27/dummy_opus_interface.py
```diff
@@ -173,11 +177,11 @@ def request_command(self, command: str) -> None:
             self.last_error = OPUSErrorInfo.UNKNOWN_COMMAND
 
         if errinfo := self.last_error.to_tuple():
-            self.error_occurred(OPUSError.from_response(*errinfo))
-        else:
-            # Broadcast the response for the command
-            state = self.state_machine.current_state
-            self.send_response(command, status=state.value, text=state.name)
+            raise OPUSError.from_response(*errinfo)
```
dc2917, line 180: Could add a `Raises:` to the docstring

dc2917 (approved):
This all looks good to me

finesse/hardware/plugins/em27/dummy_opus_interface.py
alexdewar: True!

dalonsoa (approved):
LGTM!!

PR #452, opened 2023-12-04
Clean up `Device` inheritance logic

alexdewar:
Currently, in order to create a device base type, the user passes `is_base_type=True` to the class definition. If this parameter is omitted, it is assumed that the class is a concrete device type, in which case, the class must inherit *directly* from a base type.

There are a couple of problems with this:

1. It doesn't allow for classes that inherit indirectly from a base type
2. The extra `is_base_type` parameter is a bit ugly

I've reworked the logic a bit to fix this. Now, if a class inherits directly from `Device` it is treated as a base type (no extra argument necessary) and otherwise it is treated as a device type *only* if it is not abstract (abstract classes are ignored).

I have a use case for this: I want to have an `OPUSInterfaceBase` class which inherits from a `SpectrometerBase` class. This class will be neither a device type or a base device type, but its children will be device types.

### DIFF
--- a/finesse/hardware/device.py
+++ b/finesse/hardware/device.py
@@ -12,7 +12,7 @@ import traceback
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping, Sequence
 from copy import deepcopy
-from inspect import signature
+from inspect import isabstract, signature
 from typing import Any, get_type_hints
 
 from decorator import decorate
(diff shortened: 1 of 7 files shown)


finesse/hardware/device.py
```diff
@@ -170,25 +170,16 @@ class Device(AbstractDevice):
...
-        """Initialise a device type class.
-
-        Args:
-            is_base_type: Whether this class represents a device base type
-        """
-        # If it is a base class, we initialise it as such
-        if is_base_type:
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        """Initialise a device type class."""
+        if _base_types.isdisjoint(cls.__mro__):
```
dalonsoa, line 175: I've tried to figure out what this `__mro__` attribute is doing but could not figure out from the obscure explanations I've found. Could you give a one-liner explanation of what this is for?

dalonsoa (approved):
It seems like a cleaner approach and saves the user from providing more input parameters. I think that explanation that you give in the PR description should be in the documentation, though (or a similar explanation) to describe the three possibilities.

finesse/hardware/device.py
alexdewar: It contains *all* of the ancestor classes of `cls`. Before, we were just looking at the class's immediate parents, but now we check if any of the ancestors is a device base type.

alexdewar: (I found the explanations online confusing too :laughing:)

PR #453, opened 2023-12-04
Fix stepper motor port for FINESSE

alexdewar:
I think I must have had the port wrong before, but it's a bit annoying that I didn't spot it, as now it is a bug in v1.2. I was a little worried that I *hadn't* made a mistake, but that the supposedly persistent USB strings I'm using aren't actually persistent. It doesn't seem like that's the case though: the ports are in the same order on two different machines (running Windows + Linux).

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/hardware_set/finesse_dp9800.yaml
+++ b/finesse/gui/hardware_set/finesse_dp9800.yaml
@@ -3,7 +3,7 @@ devices:
   stepper_motor:
     class_name: stepper_motor.st10_controller.ST10Controller
     params:
-      port: "0403:6011 FT1NMSVR (4)"
+      port: "0403:6011 FT1NMSVR"
       baudrate: 9600
   temperature_controller.hot_bb:
     class_name: temperature.tc4820.TC4820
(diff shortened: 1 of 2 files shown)


finesse/gui/hardware_set/finesse_dp9800.yaml
```diff
@@ -3,7 +3,7 @@ devices:
   stepper_motor:
     class_name: stepper_motor.st10_controller.ST10Controller
     params:
-      port: "0403:6011 FT1NMSVR (4)"
```
dalonsoa, line 6: So, where does this `(4)` came from?

dalonsoa (approved):

finesse/gui/hardware_set/finesse_dp9800.yaml
alexdewar: The extra number at the end is to distinguish ports which have the same vendor ID, product ID and serial number. It sounds odd, but you see this with USB serial adapters that have multiple ports (Jon is using one). The ports will always be in the same order for a given device.

PR #462, opened 2023-12-07
Drop use of USB serial numbers

### LINKED ISSUE #458: USB device serial numbers are reported incorrectly on Windows
USB devices' serial numbers are being reported incorrectly on Windows: an extra
character (letter or number) is appended. (It seemed like it was always one extra
character and that the character was always the same, but I only tinkered with it
briefly.) Unfortunately this bug has made it into the v1.2 release.

It may be related to this upstream issue:
https://github.com/pyserial/pyserial/issues/516

As we are using these serial numbers to distinguish devices, it means that the hardware
set configs will currently not work on Windows.

There are a couple of possible workarounds:

1. Discard the final character of the serial numbers on Windows
2. Stop using serial numbers (possibly temporarily)

The problem with 1 is that will only work provided that the bug is present on all
systems in the same way and will break things again once it is fixed upstream. Option 2
is also a bit unpalatable because it means that it will not be possible to have more
than one device with the same vendor and product IDs connected, which is a realistic
possibility (though the current FINESSE rig thankfully uses two different models of USB
serial adaptors).

My suggestion is to drop the use of serial numbers temporarily and open an issue to
revert this change once it is fixed upstream.

alexdewar:
# Description

Owing to #458, we cannot rely on USB serial numbers on Windows (at least until the upstream bug is fixed). In the interim, drop the use of serial numbers to distinguish devices and just use the vendor and product IDs. I have confirmed with @jonemurray that both the FINESSE and UNIRAS rigs will use serial adaptors of different makes/models, so these IDs will be different.

Once the upstream bug is fixed, we should revert this change, otherwise users will not be able to use two USB serial adaptors of the same make and model at the same time (or, at least, they won't be able to use the hardware set functionality).

Closes #458.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/hardware_sets.md
+++ b/docs/hardware_sets.md
@@ -15,17 +15,17 @@ devices:
   stepper_motor:
     class_name: stepper_motor.st10_controller.ST10Controller
     params:
-      port: "0403:6011 FT1NMSVR (4)"
+      port: "0403:6011 (4)"
       baudrate: 9600
   temperature_controller.hot_bb:
     class_name: temperature.tc4820.TC4820
     params:
-      port: "0403:6011 FT1NMSVR (2)"
+      port: "0403:6011 (2)"
       baudrate: 115200
   temperature_controller.cold_bb:
     class_name: temperature.tc4820.TC4820
     params:
-      port: "0403:6011 FT1NMSVR (3)"
+      port: "0403:6011 (3)"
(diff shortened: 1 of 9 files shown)


dalonsoa (approved):
If it works in practice, then its all good!

PR #464, opened 2023-12-07
Automatically generate state machine diagrams for documentation

alexdewar:
Currently there is a diagram for the `ScriptRunner` state machine in the documentation, but it was statically generated and does not update automatically when the code changes. In addition, there is another state machine in the code (`OPUSStateMachine`) which does not have a diagram.

Add a mkdocs plugin to automatically generate these diagrams.


finesse/hardware/plugins/em27/dummy_opus_interface.py
```diff
@@ -65,6 +65,10 @@ def __init__(
     ) -> None:
         """Create a new OPUSStateMachine.
 
+        The state diagram looks like this:
+
+        ![](../../../../../OPUSStateMachine.png)
```
dalonsoa, line 70: These sort of 5-levels-deep relative imports look awful. I was thinking that maybe we could use some sort of macro for the base path of the documentation, but apparently that is not possible :( 

https://github.com/fralau/mkdocs_macros_plugin/discussions/174

dalonsoa (approved):
LGTM!

PR #466, opened 2023-12-07
Add some more `pre-commit` hooks

alexdewar:
There are a few useful hooks that I've been using elsewhere and haven't enabled for FINESSE yet, so I thought I'd add them now.

The proximate reason for this PR is that I've been getting annoyed at VS Code auto-formatting large chunks of (admittedly inconsistently formatted) YAML files :laughing:, so I thought I'd add a formatter as a `pre-commit` hook.


alexdewar:
I think they're all fairly useful, so happy to open a PR for it :smiley:

Ruff looks cool. I'll open an issue for it.

PR #469, opened 2023-12-08
Fix: Use `closeEvent()` cf. `hideEvent()` to identify when window closed

alexdewar:
# Description

I think this bug must have been around for a while and it's rather embarrassing. It turns out we've been using the main window's `hideEvent` method to check for when the window is closed. While it *is* run when the window is closed, it also turns out it's invoked when the window is minimised (on Windows at least). This means that currently if you minimise the window, all of the connected devices will disconnect, as they are supposed to do when the window closes.

I'm not sure whether `closeEvent` wasn't working for me on Linux before (it seems fine now, although admittedly I am using a weird tiling window manager), or if I went with `hideEvent` for the symmetry with the `showEvent` handling we have next to it. In any case, I didn't think it would be invoked when the window was minimised.

Oops.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/main_window.py
+++ b/finesse/gui/main_window.py
@@ -1,7 +1,7 @@
 """Code for FINESSE's main GUI window."""
 
 from pubsub import pub
-from PySide6.QtGui import QHideEvent, QShowEvent
+from PySide6.QtGui import QCloseEvent, QShowEvent
 from PySide6.QtWidgets import (
     QGridLayout,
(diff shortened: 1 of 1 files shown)


dalonsoa (approved):
Curious. This just proves we have not tried to minimise FINESSE window ever :P

alexdewar:
Indeed :facepalm:

PR #470, opened 2023-12-08
Fix: Move device error message code to HardwareSetsControl

alexdewar:
# Description

Currently, the code responsible for showing an error message when a device error occurs lives in the DeviceControl class, where it always has. The problem is that nowadays a DeviceControl is only created when the user clicks "Manage devices", meaning that otherwise no error dialogs will be shown for device errors.

Fix by moving the relevant code to the HardwareSetsControl class instead.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/hardware_set/hardware_sets_view.py
+++ b/finesse/gui/hardware_set/hardware_sets_view.py
@@ -75,6 +75,7 @@ class HardwareSetsControl(QGroupBox):
         self._connected_devices: set[OpenDeviceArgs] = set()
         pub.subscribe(self._on_device_opened, "device.opening")
         pub.subscribe(self._on_device_closed, "device.closed")
+        pub.subscribe(self._on_device_error, "device.error")
 
         self._combo = HardwareSetsComboBox()
         """A combo box for the different hardware sets."""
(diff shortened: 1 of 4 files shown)


finesse/gui/hardware_set/hardware_sets_view.py
```diff
@@ -75,6 +75,7 @@ def __init__(self) -> None:
         self._connected_devices: set[OpenDeviceArgs] = set()
         pub.subscribe(self._on_device_opened, "device.opening")
         pub.subscribe(self._on_device_closed, "device.closed")
+        pub.subscribe(self._on_device_error, "device.error")
```
dalonsoa, line 78: The original version used `f"device.error.{instance!s}"` as topic, and not just `f"device.error"`. Is this change on purpose?

dalonsoa (approved):
LGTM!!

finesse/gui/hardware_set/hardware_sets_view.py
alexdewar: Yep. This handler will catch errors for *all* devices, as opposed to just the one specific type. `instance` is also passed as an argument, as well as part of the topic name, so we can distinguish devices that way.

PR #472, opened 2023-12-11
Refresh COM ports if device not found

alexdewar:
# Description

THIS SHOULD NOT BE MERGED BEFORE v1.3

If the user starts FINESSE with one or more of the USB devices not connected (or powered off) then they may be unable to use it, even if it is connected later. The only remedy is to restart the program. The reason for this is because the list of serial ports is retrieved once when the SerialDevice class is created and the result is cached.

An easy workaround for this is to force-refresh the list of serial ports if a given port isn't found. This is a small QoL improvement for users.

This relates to #335, but doesn't fix it (the list of serial ports in the GUI won't be updated when they are refreshed).

Someone needs to test this before we merge it. To test:

- Start FINESSE with *at least one USB device disconnected*
- Try to connect to a FINESSE hardware set; you should get an error
- Connect the USB cables
- Try to connect again; it should succeed

You don't need a working spectrometer to test this.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/serial_device.py
+++ b/finesse/hardware/serial_device.py
@@ -36,13 +36,18 @@ def _get_port_number(port: str) -> int:
     return int(match.group(1))
 
 
-def _get_usb_serial_ports() -> dict[str, str]:
+def _get_usb_serial_ports(refresh: bool = False) -> dict[str, str]:
     """Get the ports for connected USB serial devices.
 
-    The list of ports is only requested from the OS once and the result is cached.
+    The list of ports is only requested from the OS once and the result is cached,
+    unless the refresh argument is set to true.
+
+    Args:
+        refresh: Refresh the list of serial ports even if they have already been
+                 requested
     """
     global _serial_ports
-    if _serial_ports is not None:
+    if _serial_ports is not None and not refresh:
         return _serial_ports
 
     # Keep track of ports with the same vendor and product ID and assign them an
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
LGTM! 

So, what's missing for this PR to be able to close the issue you mention?

jamesturner246 (approved):
Looks good. Is it possible that there is the converse problem, where a device is later unplugged, but the COM port is still registered?

alexdewar:
@jamesturner246 You're quite right!

So this problem could still happen:

- FINESSE starts with a device connected
- The user triggers loading the list of COM ports (by connecting to something or opening the manage devices dialog)
- They disconnect a device
- They then reconnect the device to a different port
- They try to connect again

Then the connection will fail because it will be trying to use the old COM port for the device. Unfortunately in that case the list of COM ports won't be refreshed, although that would fix things.

I've had another go at implementing it, where the list of COM ports will be refreshed if the connection fails for any reason, rather than just that the port name is missing from the list. That should catch the problem you mention too.

This is obvs completely untested, so it would be great if someone could check that it works before we merge it.

jamesturner246:
That sounds sensible. Happy to test with @dc2917 on the Monday.

jamesturner246 (approved):
LGTM!

dalonsoa (approved):
LGTM!

jamesturner246:
I've run the tests in lab, and confirm that they are working.

Issue #459, opened 2023-12-05
Make a generic spectrometer device base type


alexdewar:
This class will be used as a base for the existing OPUS plugins as well as the future plugin for the ABB spectrometer (#322).

It should accept messages to invoke different commands for the spectrometer (viz. start measuring, stop measuring, etc.) and send a message every time that the device's status changes (e.g. from connected to measuring). Note that in order to notice when the status changes on the EM27 and ABB devices, it will be necessary to poll them.

PR #482, opened 2023-12-12
OPUS: Poll status on backend

alexdewar:
# Description

This PR simplifies the OPUS API for reporting the device's status. It's the last big PR on the road to #459 -- after this there are just some smaller frontend changes and then a very small amount of refactoring.

The OPUS device itself reports its status in response to a command or when it is explicitly requested, which is reflected in our current API. This means there are two ways the frontend can end up being notified of a status change, whereas it makes more sense to have a single "status changed" message. It also doesn't make much sense to have the frontend be responsible for polling the backend -- if just changes in status are communicated it simplifies much of the frontend logic. There's no particular reason why a spectrometer needs to have a polling-based interface (though the ABB spectrometer does too) so I think it's more sensible to keep this encapsulated in the specific device classes. (Maybe someday someone will make a spectrometer which can do proper push notifications.)

The hardest part of implementing this was changing the state machines for running measure scripts and the `DummyOPUSInterface`. Accordingly, I had to fix and replace a whole lot of tests, which means the diff looks rather large. The changes to the actual code are much smaller though.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -68,6 +68,14 @@ TEMPERATURE_CONTROLLER_TOPIC = "temperature_controller"
 OPUS_IP = "10.10.0.2"
 """The IP address of the machine running the OPUS software."""
 
+OPUS_POLLING_INTERVAL = 1.0
+"""How long to wait between polls of the EM27's status.
+
+Note that in reality the minimum poll interval is ~2s, because that's how long the
+device takes to reply. This value then determines how often we wait before even starting
+a request.
+"""
+
 SPECTROMETER_TOPIC = "spectrometer"
 """The topic name to use for spectrometer-related messages."""
 
--- a/finesse/gui/measure_script/script.py
+++ b/finesse/gui/measure_script/script.py
@@ -13,7 +13,6 @@ from typing import Any
 
 import yaml
 from pubsub import pub
-from PySide6.QtCore import QTimer
 from PySide6.QtWidgets import QWidget
 from schema import And, Or, Schema, SchemaError
 from statemachine import State, StateMachine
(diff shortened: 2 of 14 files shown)


finesse/config.py
```diff
@@ -68,6 +68,14 @@
 OPUS_IP = "10.10.0.2"
 """The IP address of the machine running the OPUS software."""
 
+OPUS_POLLING_INTERVAL = 1.0
+"""How long to wait between polls of the EM27's status.
+
+Note that in reality the minimum poll interval is ~2s, because that's how long the
+device takes to reply. This value then determines how often we wait before even starting
+a request.
```
dalonsoa, line 76: I really don't understand this. Does it mean that the fastest we can get a response is every 3s?

dalonsoa (approved):
Ok, _I think_ I follow the evolution of the states in the state machine and the cross-fire of pubsub messages. The tests have been sueful to understand this. It looks sensible, but the real proof would be if this works in the lab as expected!

alexdewar:
@dalonsoa Thanks for the review!

It has been tested with the real hardware, so it should be fine. I'll get Jon to retest everything once it's all merged anyway.

finesse/config.py
alexdewar: Sorry, this isn't v well worded. The spectrometer itself seems to handle incoming requests on a single thread and only responds to messages once every ~2s, so that's the max amount of time you'll be waiting.

We *could* just send a new status request as soon as we've received the last response and that would mean we'd still get an update every 2s, but it seemed sensible to limit how often we send requests from our side too (i.e. no more than one per sec), even if the result is the same in this case. If there was a new model that responded instantly to requests and we didn't have a limit then it would mean that we would spam it constantly.

Does that make sense?

PR #483, opened 2023-12-12
Replace plethora of tools with `ruff`

alexdewar:
I've been looking at whether to swap our exiting tooling in `poetry_template_2` for `ruff` and thought I'd try it out on the FINESSE codebase. It found a number of small stylistic issues with the current codebase and fixed most of them automatically.

I've used it to replace:

- autoflake
- black
- isort
- flake8 (and flake8-docstrings)
- pyupgrade

In addition, I also enabled ruff's own rule set.

I was concerned that it might not be so hot with `pyupgrade`'s checks, so went back in the git history to before I enabled `pyupgrade` and ran `ruff` over the codebase. It seemed to do a pretty decent job, though there was one check that `pyupgrade` could auto-fix that `ruff` couldn't. In the grand scheme of things though, it seems nicer to just be using a single tool.

Although it found various small issues all over the place, there are no merge conflicts with other things I've been working on with FINESSE, so I thought I should get this merged.

### DIFF
--- a/finesse/gui/hardware_set/device_view.py
+++ b/finesse/gui/hardware_set/device_view.py
@@ -2,8 +2,8 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Mapping, Sequence
-from typing import AbstractSet, Any, cast
+from collections.abc import Mapping, Sequence, Set
+from typing import Any, cast
 
 from pubsub import pub
 from PySide6.QtWidgets import (
@@ -319,7 +319,7 @@ class DeviceTypeControl(QGroupBox):
 class DeviceControl(QGroupBox):
     """Allows for viewing and connecting to devices."""
 
-    def __init__(self, connected_devices: AbstractSet[OpenDeviceArgs]) -> None:
+    def __init__(self, connected_devices: Set[OpenDeviceArgs]) -> None:
         """Create a new DeviceControl."""
         super().__init__("Device control")
         self.setSizePolicy(QSizePolicy.Policy.Minimum, QSizePolicy.Policy.Fixed)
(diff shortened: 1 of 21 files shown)


jamesturner246 (approved):
Nice too just have one tool. No chance of them battling each other. LGTM.

finesse/gui/hardware_set/device_view.py
```diff
@@ -319,7 +319,7 @@ def _on_open_close_clicked(self) -> None:
 class DeviceControl(QGroupBox):
     """Allows for viewing and connecting to devices."""
 
-    def __init__(self, connected_devices: AbstractSet[OpenDeviceArgs]) -> None:
+    def __init__(self, connected_devices: Set[OpenDeviceArgs]) -> None:
```
dalonsoa, line 322: Out of curiosity, is this something you have decided to fix here simply because of convenience, or has it been a sort of recommendation/fix done by ruff? Either way, why the change?

finesse/hardware/device.py
```diff
@@ -65,11 +65,11 @@ def get_device_types() -> dict[DeviceBaseTypeInfo, list[DeviceTypeInfo]]:
 class AbstractDevice(ABC):
     """An abstract base class for devices."""
 
-    _device_base_type_info: DeviceBaseTypeInfo
+    _device_base_type_info: ClassVar[DeviceBaseTypeInfo]
```
dalonsoa, line 68: What's `ClassVar` for?

dalonsoa (approved):
All changes look good, although I must admit I don't know the reason for some of them (the type ones, specially).

finesse/gui/hardware_set/device_view.py
alexdewar: It was a recommendation from ruff. Apparently using `typing.AbstractSet` is deprecated and we should be using `collections.abc.Set` instead.

finesse/hardware/device.py
alexdewar: It indicates that the variable is a class variable so it should not be being changed by individual instances of `AbstractDevice`. (I just tried doing this to see what happened and `mypy` duly told me off.)

Ruff suggested this one too.

PR #484, opened 2023-12-13
Simplify frontend spectrometer widgets

### LINKED ISSUE #190: OPUS buttons should be enabled/disabled as needed
Currently all of the buttons for interacting with OPUS are enabled all of the time. It would make for a better UX if we could enabled/disable them selectively (e.g. only enable the "Stop" and "Cancel" buttons when a measurement is in progress).

### LINKED ISSUE #396: GUI should show EM27 status in more user-friendly form

### LINKED ISSUE #461: `ScriptControl`: Disable "Run" button while spectrometer is measuring
Users shouldn't be starting a measure script while the spectrometer is already recording a measurement, otherwise an error will occur when the script running code tries to initiate a measurement later.

alexdewar:
# Description

The main purpose of this PR is to simplify the spectrometer panel, following the necessary backend changes. Now it is just composed of a few buttons and a label showing the current status:

image

I've fixed the enabling/disabling logic for the script run button (#461) and the buttons in the spectrometer panel are now also enabled/disabled as needed (#190).

Fixes #190. Fixes #461. Fixes #396.

## Type of change

## Key checklist

## Further checks


finesse/gui/event_counter.py
```diff
@@ -40,16 +42,20 @@ def __init__(
...
             self._on_target_reached()
 
-    def decrement(self, **kwargs) -> None:
+    def decrement(self) -> None:
         """Decrease the counter by one and run callback if count drops below target."""
         self._count -= 1
         if self._count == self._target_count - 1:
             self._on_below_target()
+
+    def _on_device_closed(self, instance: DeviceInstanceRef) -> None:
```
dalonsoa, line 59: I seem to remember we had removed this before as it seemed redundant with `decrement` (`instance` is not used here). Why is it back?

dalonsoa (approved):
LGTM - everything that is about simplifying code is a step in the right direction. The answer to my only comment I think it is "because `pypubsub` is not happy, otherwise", but I leave it there for you to confirm.

finesse/gui/event_counter.py
alexdewar: > "because pypubsub is not happy, otherwise"

This ^^ :wink:

As you suggested before, you can specify all your pubsub topics + args in .py files and that might also work. That's a job for another day though.

Issue #471, opened 2023-12-08
Amalgamate "stop" and "cancel" commands for spectrometer


alexdewar:
While the OPUS API supports both stopping and cancelling the current measurement, cancelling it is the only operation that actually seems to do anything. In any case, cancelling does what we need it to do (stops the measurement immediately), so we should just drop the current "stop" operation and replace it with "cancel" (while still keeping the name as "stop").

PR #485, opened 2023-12-14
Add a generic `SpectrometerBase` class

### LINKED ISSUE #459: Make a generic spectrometer device base type

alexdewar:
# Description

Add a generic spectrometer base class and use it as the parent of `OPUSInterfaceBase`. This is the final change I want to make before we release v1.3 :partying_face: 

I do have some other changes that I'd like to see merged (in particular I'd like to implement #471 which should make it easier to develop the new spectrometer driver), but unfortunately Jon won't be able to test until next week, so I think we should be conservative about introducing new changes at this point. I'll open PRs for the remaining bits and they can be merged once someone has an opportunity to test them in the lab.

Fixes #459.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/plugins/spectrometer/opus_interface_base.py
+++ b/finesse/hardware/plugins/spectrometer/opus_interface_base.py
@@ -3,9 +3,7 @@ from __future__ import annotations
 
 from abc import abstractmethod
 
-from finesse.config import SPECTROMETER_TOPIC
-from finesse.hardware.device import Device
-from finesse.spectrometer_status import SpectrometerStatus
+from finesse.hardware.plugins.spectrometer.spectrometer_base import SpectrometerBase
 
 
 class OPUSError(Exception):
@@ -17,13 +15,24 @@ class OPUSError(Exception):
         return cls(f"Error {errcode}: {errtext}")
 
 
-class OPUSInterfaceBase(Device, name=SPECTROMETER_TOPIC, description="OPUS device"):
+class OPUSInterfaceBase(SpectrometerBase):
     """Base class providing an interface to the OPUS program."""
 
-    def __init__(self) -> None:
-        """Create a new OPUSInterfaceBase."""
-        super().__init__()
-        self.subscribe(self.request_command, "request")
+    def connect(self) -> None:
+        """Connect to the spectrometer."""
+        self.request_command("connect")
+
+    def start_measuring(self) -> None:
+        """Start a new measurement."""
+        self.request_command("start")
+
+    def stop_measuring(self) -> None:
+        """Stop the current measurement when finished."""
+        self.request_command("stop")
+
+    def cancel_measuring(self) -> None:
+        """Cancel the current measurement immediately."""
+        self.request_command("cancel")
 
     @abstractmethod
     def request_command(self, command: str) -> None:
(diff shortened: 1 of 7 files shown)


jamesturner246:
This is the abstract class which the FINESSE and UNIRAS spectrometers will eventually be concrete classes of?

alexdewar:
Yep! I actually made an `OPUSInterfaceBase` class for the dummy and real EM27 spectrometer classes to inherit from and it's this class which inherits from `SpectrometerBase`. The ABB spectrometer class should probably inherit from `SpectrometerBase` directly.

jamesturner246:
Ah, I see. How come the abstract spectrometer followed by abstract opus followed by spectrometer? Are there many different opus devices that need their own interface?

alexdewar:
Not a silly question. There are two in our case: one for connecting to the real EM27 and another "dummy" driver that just emulates the spectrometer.

jamesturner246:
Ah, right, and there might be other spectrometer models you might need both a dummy and 'real' class for.

alexdewar:
Yeah, though I don't think we'll need a dummy device for the ABB spectrometer, because a) we already have one dummy spectrometer to work with and b) we can test the ABB spectrometer code with their emulator; there's no emulator for OPUS.

finesse/hardware/plugins/spectrometer/opus_interface_base.py
```diff
@@ -17,13 +15,24 @@ def from_response(cls, errcode: int, errtext: str) -> OPUSError:
         return cls(f"Error {errcode}: {errtext}")
 
 
-class OPUSInterfaceBase(Device, name=SPECTROMETER_TOPIC, description="OPUS device"):
+class OPUSInterfaceBase(SpectrometerBase):
```
dalonsoa, line 18: I had not relealised of this, but I have two questions:
1. Is it possible (i.e. does it work as intended) having an `abstractmethod` (`request_command`) in a non-abstract class? 
2. What is the point of this intermediate class now that we have the `SpectrometerBase`?  All of these methods could go in the `OpusInterface`, deriving directly from `SpectrometerBase`, and then have the `DummyOPUSInterface` inheriting from `OpusInterface`. To be honest, that is something we could have done before, but I just thought about it now.

dalonsoa (approved):
I have a, potentially, major suggestion but considering we do not want to break anything before a new release, I leave up to you the decision of making any changes on that front or not.

finesse/hardware/plugins/spectrometer/opus_interface_base.py
alexdewar: 1. Yep! It's still an abstract class (i.e. you can't create an instance of it).
2. This class doesn't do *much*, just sends the appropriate command to the OPUS device. The `DummyOPUSInterface` is pretty similar to `OPUSInterface` in that it accepts the same commands, has the same states and throws the same errors. We could drop the `OPUSInterfaceBase` class, but then we'd basically just end up moving the boilerplate from it to `OPUSInterface` and `DummyOPUSInterface` instead. Not sure.

alexdewar: I'm going to merge this for now, if that's ok. V happy to revisit this discussion, but I'd like to make the release for FINESSE. We can always change this later.

dalonsoa: Not a problem!

PR #487, opened 2023-12-14
Amalgamate "stop" and "cancel" commands for spectrometers

### LINKED ISSUE #471: Amalgamate "stop" and "cancel" commands for spectrometer

alexdewar:
# Description

DO NOT MERGE BEFORE v1.3

The OPUS API allows for sending both a "cancel" and a "stop" command during a measurement. "Cancel" stops the measurement immediately but it's not apparent what "stop" does (maybe it stops after the current series of measurements?). Either way, we don't need the separate stop command, so let's just drop the stop command and replace it with what we're currently calling cancel.

This means we can simplify `SpectrometerBase` slightly and drop a button from the GUI.

This code needs to be tested on real hardware before being merged. You will need the spectrometer for this.

A reasonable test would be:

1. Choose the "FINESSE (with DP9800)" hardware set from the dropdown box in the top left and click "Connect"
2. In the spectrometer panel, click the "Connect" button
3. Try starting a measurement by clicking the "Start" button and confirm that it has started in OPUS
4. Try cancelling this measurement partway through by clicking the "Stop" button and confirm that it has stopped in OPUS
5. Try running a measure script (I've attached an example one, but any should do)

Fixes #471.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/plugins/spectrometer/dummy_opus_interface.py
+++ b/finesse/hardware/plugins/spectrometer/dummy_opus_interface.py
@@ -72,7 +72,7 @@ class OPUSStateMachine(StateMachine):
         """Timer signalling the end of a measurement."""
         self.measure_timer.setInterval(round(measure_duration * 1000))
         self.measure_timer.setSingleShot(True)
-        self.measure_timer.timeout.connect(self.stop)
+        self.measure_timer.timeout.connect(self._on_measure_finished)
 
         super().__init__()
 
@@ -81,13 +81,13 @@ class OPUSStateMachine(StateMachine):
         self._start_connecting()
         self._finish_connecting()
 
-    def cancel(self) -> None:
-        """Cancel the current measurement."""
+    def stop(self) -> None:
+        """Stop the current measurement."""
         self._cancel_measuring()
         self._reset_after_cancelling()
         logging.info("Cancelling current measurement")
 
-    def stop(self) -> None:
+    def _on_measure_finished(self) -> None:
         """Finish measurement successfully."""
         self._start_finishing_measuring()
         self._finish_measuring()
--- a/finesse/gui/spectrometer_view.py
+++ b/finesse/gui/spectrometer_view.py
@@ -19,7 +19,6 @@ class SpectrometerControl(DevicePanel):
             "connect": "Connect",
             "start_measuring": "Start",
             "stop_measuring": "Stop",
-            "cancel_measuring": "Cancel",
         }
     )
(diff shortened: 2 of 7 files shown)


alexdewar:
Something like that, though honestly I think the concepts ended up a bit confused in my head too :stuck_out_tongue: .

For the dummy device, I was treating `stop()` as meaning that the measurement had completed successfully and `cancel()` as meaning that it had been aborted, but that did mean that users could in theory prematurely trigger a successful end to the measurement, which is a bit weird. `_on_measure_finished` does the same thing as `stop` did (it just transitions to the "finishing" then "connected" states), but users now can't trigger it directly.

jamesturner246 (approved):
Ah, okay, I'm with you. Looks good, then!

finesse/hardware/plugins/spectrometer/dummy_opus_interface.py
```diff
@@ -81,13 +81,13 @@ def connect(self) -> None:
...
+        """Stop the current measurement."""
         self._cancel_measuring()
         self._reset_after_cancelling()
         logging.info("Cancelling current measurement")
 
-    def stop(self) -> None:
+    def _on_measure_finished(self) -> None:
         """Finish measurement successfully."""
         self._start_finishing_measuring()
         self._finish_measuring()
```
dalonsoa, line 93: As you are discussing, this is very confusing and I'm not sure I follow the use case for these two paths now that the "Cancel" button is gone. We have the following paths in the state machine:

- `Stop` path:
`measuring -> cancelling -> connected`
- `on_measure_finish` path:
`measuring -> finishing -> connected`

Why are they different? When each should be used? Regardless of being triggered by the user or not, it seems to me that the second one is to be followed under normal operating conditions, while the first one is followed in exceptional circumstances. If that's right, why if there is a timeout (line 75), we are triggering the normal operating path rather than the exceptional one?- What other actions need to be revisited based on this?

dc2917 (approved):
This looks good to me, though I am also slightly confused by the issue Diego has raised. `stop()` still exists, and serves the same purpose as before, so I'm not sure why the timeout has been set for `_on_measure_finished()`.

jamesturner246:
I've run the tests in lab, and confirm that they are working.

PR #504, opened 2024-01-17
Added Seneca device parameters

### LINKED ISSUE #433: Make arguments to `SenecaK107` into device parameters
The temperature and voltage ranges should be configurable by the user. The temperature ranges of Seneca devices can be changed -- and the researchers have indicated that they may want to do this -- and this will require corresponding changes on the software side.

There is more information [in the docs](https://imperialcollegelondon.github.io/FINESSE/hardware/). A good example of using device parameters is given in `plugins/em27/em27_sensors.py`.

As part of this, we should also make sure that errors are raised if incorrect arguments are passed to `SenecaK107` (e.g. if the minimum voltage is higher than the maximum).

CWestICL:
Added device parameters to the `SenecaK107` device class.

Closes #433

### DIFF
--- a/finesse/hardware/plugins/temperature/senecak107.py
+++ b/finesse/hardware/plugins/temperature/senecak107.py
@@ -20,7 +20,17 @@ class SenecaK107Error(Exception):
     """Indicates that an error occurred while communicating with the device."""
 
 
-class SenecaK107(SerialDevice, TemperatureMonitorBase, description="Seneca K107"):
+class SenecaK107(
+    SerialDevice,
+    TemperatureMonitorBase,
+    description="Seneca K107",
+    parameters={
+        "min_temp": "The minimum temperature limit of the device",
+        "max_temp": "The maximum temperature limit of the device",
+        "min_millivolt": "The minimum voltage output (millivolts) of the device",
+        "max_millivolt": "The maximum voltage output (millivolts) of the device",
+    },
+):
     """An interface for the Seneca K107USB serial converter.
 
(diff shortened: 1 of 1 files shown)


jamesturner246 (approved):
Looks good! Will these parameters be used for the temperature sensor limit bug?

CWestICL:
No I was going to tackle that in a separate PR

finesse/hardware/plugins/temperature/senecak107.py
```diff
@@ -20,7 +20,17 @@ class SenecaK107Error(Exception):
...
+class SenecaK107(
+    SerialDevice,
+    TemperatureMonitorBase,
+    description="Seneca K107",
+    parameters={
+        "min_temp": "The minimum temperature limit of the device",
+        "max_temp": "The maximum temperature limit of the device",
+        "min_millivolt": "The minimum voltage output (millivolts) of the device",
+        "max_millivolt": "The maximum voltage output (millivolts) of the device",
+    },
```
dc2917, line 32: Do we need to set values for these? cf. the equivalent for the DP9800 or TC4820, where the default baud rates are set.

CWestICL: The default values are being defined in the `.__init__()` which is how the params are being defined in `em27_sensors.py` (the example Alex gave in #433).

dc2917: Ah yes, thanks for clarifying. I should have read the docs 😅

PR #505, opened 2024-01-18
Seneca out-of-range temperature detection

### LINKED ISSUE #490: `SenecaK107`: Replace out-of-range temperatures with NaN
Currently, if one of the sensors is not attached to anything, its temperature will be given as >600°C. It would be better to catch these out-of-range values and replace them with NaNs so they don't skew the plots or data analysis.

Once UNIRAS is finished, it is unlikely that we will see this problem, but *if* a sensor wire were to become disconnected, it would be more useful to have the output be a NaN (possibly also with a warning in the program log).

CWestICL:
Added logic to `calc_temp()` that replaces temperatures outside the minimum and maximum temperature ranges with NaNs. If the resulting numpy array contains NaNs, a warning is raised.

Tests have also been altered to expect the NaN values when parsing the data.

Closes #490

### DIFF
--- a/finesse/hardware/plugins/temperature/senecak107.py
+++ b/finesse/hardware/plugins/temperature/senecak107.py
@@ -151,10 +151,8 @@ class SenecaK107(
         # Adjusts for minimum temperature limit
         calc += self.MIN_TEMP
 
-        nan = float("nan")
-
-        calc[calc > self.MAX_TEMP] = nan
-        calc[calc < self.MIN_TEMP] = nan
+        calc[calc > self.MAX_TEMP] = numpy.nan
+        calc[calc < self.MIN_TEMP] = numpy.nan
 
         if numpy.isnan(calc).any():
             logging.warn(f"Out-of-range temperature(s) detected: {calc}")
--- a/finesse/hardware/plugins/temperature/senecak107.py
+++ b/finesse/hardware/plugins/temperature/senecak107.py
@@ -1,4 +1,5 @@
 """This module provides an interface to Seneca temperature readers."""
+import logging
 from collections.abc import Sequence
 
 import numpy
@@ -132,6 +133,9 @@ class SenecaK107(
     def calc_temp(self, vals: numpy.ndarray) -> numpy.ndarray:
         """Convert data read from the SenecaK107 device into temperatures.
 
+        Any readings outside the minimum and maximum temperature values will be changed
+        to NaNs and a warning will be raised in the logs.
+
         Args:
             vals: The numpy array described by the data received from the device.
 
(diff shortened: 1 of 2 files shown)


tests/hardware/plugins/temperature/test_senecaK107.py
```diff
@@ -87,12 +87,12 @@ def test_parse_data(dev: SenecaK107, data: bytes) -> None:
...
-        631.4406250000001,
-        631.4175,
+        float("nan"),
+        float("nan"),
         20.14281249999999,
-        631.4406250000001,
+        float("nan"),
         19.946250000000006,
-        631.4406250000001,
+        float("nan"),
```
jamesturner246, line 95: Might also want to test to see if `MIN_TEMP` is respected here.

jamesturner246 (approved):
LGTM! See the note about `MIN_TEMP` below.

tests/hardware/plugins/temperature/test_senecaK107.py
CWestICL: Yeah this is definitely something we should write tests for, but I imagine it'll require having examples of bytestring data containing temperatures below the minimum limit. I don't know enough about bytestrings to fabricate test data like this so might be worth making a separate issue.

dc2917 (changes requested):
It likely won't make any practical difference but I think we should use `numpy.nan` rather than `float("nan")`. As the values are all in a NumPy array anyway, I think it makes sense to use their NaN implementation.

finesse/hardware/plugins/temperature/senecak107.py
```diff
@@ -146,6 +150,13 @@ def calc_temp(self, vals: numpy.ndarray) -> numpy.ndarray:
         calc *= self.SCALING_FACTOR
         # Adjusts for minimum temperature limit
         calc += self.MIN_TEMP
+
+        calc[calc > self.MAX_TEMP] = numpy.nan
+        calc[calc < self.MIN_TEMP] = numpy.nan
+
+        if numpy.isnan(calc).any():
+            logging.warn(f"Out-of-range temperature(s) detected: {calc}")
```
dc2917, line 158: `logging.warn` seems to be undocumented so I'm assuming it's deprecated.
```suggestion
            logging.warning(f"Out-of-range temperature(s) detected: {calc}")
```

Also, I'm fairly sure calling `logging` to do the logging directly is advised against - we should be using `logging.getLogger()`. However, it seems everywhere else in the code `logging` is called directly. @dalonsoa do you have any insight on this? I know you used `logging.getLogger()` in your `materials` module in PyECN.

dalonsoa: It is not advised again - [Python documentation uses it with no problem in the first example](https://docs.python.org/3/library/logging.html) - but might be not what you want.

So, `logging.warning`, `logging.error`, etc. all use the `root` logger. If we use `logging.getLogger()`, with no arguments, we also get the `root` logger, therefore `logging.warning("something")` is the same that  `logging.getLogger().warning("something")`.

The difference comes when you have different loggers configured for different parts of the program or want to configure a specific behavior depending on where the logging happen, or what the log is about. In that case you could do `logging.getLogger("gui").warning("something")`, for example, and have a handler configured that send all the logging events sent to the `gui` logger to a pop-up window or something like that.

In summary, in our case, it does not matter. In general, it might.

dc2917: Ah okay, thanks for clarifying!

dc2917 (approved):

PR #512, opened 2024-02-09
DECADES sensor queries over LAN

### LINKED ISSUE #325: Add support for interfacing with aircraft sensors over LAN
After speaking to FAAM, it transpires that they have a RESTful API for accessing aircraft sensors in real time (their system is called DECADES). Apparently there is a new MQTT-based system in the works (DECADES-2), but it won't be operational until the end of 2025, so it won't be useful to us for now.

They stressed that this data should not be used for data analysis as it is of relatively low quality: the temporal resolution is only ~1s and there are no corrections applied. They say they can provide higher quality data (in netCDF format, IIRC) post flight. The lower quality data should still be good enough for our purposes: we just need to show it in the GUI as a guide to the user about what's going on.

It's debatable whether this data should be included in the CSV files at all, though if it is we should probably flag up that it's of low quality (maybe include "don't use" in the header name or something).

FAAM have a website for this, but there doesn't seem to be much useful documentation for us: https://www.faam.ac.uk/data-centre/live-data/

I've marked this as "Needs info" because we're still waiting to hear from FAAM about the details of their API. They also indicated that they should be able to provide us with an emulator (either on the web or as a Python script), which is probably necessary for us to be able to implement this feature.

jamesturner246:
# Description

This PR introduces code to talk to the DECADES sensor API. 

Currently only three sensors are queried, as configured in `./finesse/config.py`, and nothing is actually done with the data after the pubsub messages. Poling interval and URL is also configured there. The code reuses the existing `HTTPRequester` class, as used by the spectrometer, so is similar in many respects.

Tests are added, and the code is tested and working with the DECADES emulator.

Fixes #325 

## Type of change

## Key checklist

## Further checks


finesse/hardware/plugins/decades/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Code for interfacing with the DECADES server.
+
+This module contains code for polling sensor values using the DECADES API.
```
dc2917, line 3:
```suggestion
This module contains code for polling aircraft parameter values using the DECADES API.
```

finesse/config.py
```diff
@@ -53,6 +53,25 @@
 EM27_SENSORS_TOPIC = "em27_sensors"
 """The topic name to use for EM27 sensor-related messages."""
 
+DECADES_HOST = "localhost"
+"""The IP address or hostname of the DECADES server."""
+
+DECADES_SENSORS_URL = "http://{host}/decades/livedata"
+"""The URL of the DECADES sensor data endpoint."""
+
+DECADES_SENSORS_POLL_INTERVAL = 60.0
```
dc2917, line 62: We might want a higher poll rate than this, especially when it comes to plotting the air temperature over time, but we can probably leave it as this for now.

finesse/config.py
```diff
@@ -53,6 +53,25 @@
 EM27_SENSORS_TOPIC = "em27_sensors"
 """The topic name to use for EM27 sensor-related messages."""
 
+DECADES_HOST = "localhost"
+"""The IP address or hostname of the DECADES server."""
+
+DECADES_SENSORS_URL = "http://{host}/decades/livedata"
```
dc2917, line 59: I think we should avoid the term sensors in reference to DECADES. Could swap SENSORS -> SERVER, or just drop SENSORS from each of these names and it would make sense.

dc2917 (approved):
This looks good, I've just made a suggestion that we drop the term "sensors" from DECADES things, so as to avoid confusion between the internal spectrometer sensors. It's probably fine either way, so if you feel keeping "sensors" makes sense I don't mind 😅

Presumably displaying them on the GUI will be a separate issue/PR?

finesse/config.py
```diff
@@ -53,6 +53,25 @@
...
+"""The URL of the DECADES sensor data endpoint."""
+
+DECADES_SENSORS_POLL_INTERVAL = 60.0
+"""Poll rate for DECADES sensors."""
+
+DECADES_SENSORS_TOPIC = "decades_sensors"
+"""The topic name to use for DECADES sensor-related messages."""
+
+DECADES_SENSORS_QUERY_LIST = [
```
dalonsoa, line 68: Shall we also query the `time` if that is available, so it can be later correlated with other measurements?

finesse/hardware/plugins/decades/decades_sensors.py
```diff
@@ -0,0 +1,128 @@
...
+        self.send_data()
+
+    def send_data(self) -> None:
+        """Request the sensor data from the DECADES server.
+
+        The HTTP request is made on a background thread.
+        """
+        epoch_time = str(int(time.time()))
+        url = QUrlQuery(self._url + "?")
+        url.addQueryItem("frm", epoch_time)
```
dalonsoa, line 89: Just checking this is actually `frm` and not `from`.

finesse/hardware/plugins/decades/decades_sensors.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Poll device once on open.
+        # TODO: Run this synchronously so we can check that things work before the
+        # device.opened message is sent
+        self.send_data()
+
+    def send_data(self) -> None:
+        """Request the sensor data from the DECADES server.
+
+        The HTTP request is made on a background thread.
```
dalonsoa, line 85: Maybe it's some of the !T magic, but where is it set that the request is done in a background thread?

finesse/config.py
jamesturner246: No worries. I'll leave it for now.

finesse/config.py
jamesturner246: Done.

finesse/config.py
```diff
@@ -53,6 +53,25 @@
 EM27_SENSORS_TOPIC = "em27_sensors"
 """The topic name to use for EM27 sensor-related messages."""
 
+DECADES_HOST = "localhost"
```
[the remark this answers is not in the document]
jamesturner246: Good question. Should probably make this more easily configurable.

finesse/hardware/plugins/decades/decades_sensors.py
jamesturner246: Not actually sure. This part was copied from `em27_sensors.py`. Should probably have checked,

finesse/hardware/plugins/decades/decades_sensors.py
jamesturner246: No idea why, but yes. `frm`, not `from`.

finesse/config.py
jamesturner246: Yeah, agree. Not sure how best to do this yet. Timestamps are returned from decades with each query anyway:

```
{
    "deiced_true_air_temp_c": [
        20.672006730159012
    ],
    "nondeiced_true_air_temp_c": [
        20.790049884423468
    ],
    "utc_time": [
        1708360706.0
    ]
}
```

PR #517, opened 2024-02-16
Layout change - Data file view moved to left side

### LINKED ISSUE #420: Odd placement and sizing of controls
Since #413, the controls in the window look a little odd:

image

There are several related issues (#405, #419, #396), but one other problem may be that there are too many controls on the right compared to the left of the window. One tweak that could be made is to move the data file control to the left-hand side.

Another possibility is that the temperature plot is too large.

CWestICL:
GUI layout change to move Data file view from right side to left side.

I think this change makes sense for the layout as it creates a divide between the temperature monitoring views on the right and the config/other sensors on the left.
image

However now that the Data file view is on the left, its width is shortened and the directory/filename fields are a bit stubby. I suggest changing the layout of the Data file view so that these two fields are vertically aligned. I've attempted to do this with the `QGridLayout` but I still need to play around to get the elements aligned properly.
image

Opening in draft to get feedback on this layout change before it's refined.

Closes #420

### DIFF
--- a/finesse/gui/data_file_view.py
+++ b/finesse/gui/data_file_view.py
@@ -5,8 +5,8 @@ from typing import cast
 
 from pubsub import pub
 from PySide6.QtWidgets import (
+    QGridLayout,
     QGroupBox,
-    QHBoxLayout,
     QLabel,
     QLineEdit,
     QMessageBox,
@@ -35,7 +35,7 @@ class DataFileControl(QGroupBox):
         """Create a new DataFileControl."""
         super().__init__("Data file")
 
-        layout = QHBoxLayout()
+        layout = QGridLayout()
 
         self.open_dir_widget = OpenDirectoryWidget(
             parent=self,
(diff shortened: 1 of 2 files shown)


dc2917:
I think moving the Data File panel to the left makes sense now that the left-hand side is less busy. If you want to align the components, you can stick each of them in a `QGridLayout` - an example of this is in temperature_monitor_view.py, but actually, most of the panels contain QGridLayouts so there should be a few examples to refer to.

jamesturner246 (approved):
Looks good and tidy -- agree with Dan.

CWestICL:
This is the best I've been able to align the elements with the `QGridLayout`. Still not perfect, not sure if anyone has some ideas on getting it a bit tidier.
image

jamesturner246 (approved):
Whilst I'm not super familiar with GUI design, the buttons do fill the area more neatly this time. It's an approval from me, but maybe someone with more QT experience knows differently.

CWestICL:
Ah perfect, that's done it. Looks much cleaner now!
image

dc2917 (approved):
Yep, looks good, nice one!

PR #520, opened 2024-02-22
Add plugin for time source devices

### LINKED ISSUE #488: Synchronise timestamps between different sensors
For UNIRAS, there are likely to be three kinds of data files produced which use different time sources:

1. The CSV files produced by the FINESSE program (using the laptop's system clock)
2. Spectrographic data from ABB's interferometer (using its own clock)
3. Aircraft sensor data, e.g. GPS coordinates etc. (using the aircraft's onboard clock, which is likely to be highly accurate)

Happily, FAAM report that they have an onboard NTP server, which should yield the same timestamps as used by the aircraft data, and ABB report that their interferometer can be configured to use an NTP server as a time source.

(Side note: I couldn't find an option for setting an NTP server in FTSW500 itself, so presumably it will have to be set on the actual interferometer. I've no idea how that will work. Someone will need to chase up ABB about this nearer the time.)

That leaves FINESSE. There are a couple of options here:

1. Configure Windows on the laptop to use the aircraft's NTP server to set the system clock
2. Use a Python library to retrieve the time correction from the NTP server directly (e.g. [ntplib](https://pypi.org/project/ntplib/))

I'm somewhat leaning towards 2 as the better option, because if we go with 1:

- FINESSE won't know whether or not Windows' time settings have been correctly configured
- FINESSE won't know whether synchronisation has happened yet or not
- We will need to provide careful instructions for researchers about how to configure Windows before they run FINESSE
- I'd worry that a Windows update might break something

If we decide to go with 2, we could perhaps have time sources as another kind of plugin. Then there'd be one for just using the system time (as would be used by the FINESSE rig) and one for NTP (as used by UNIRAS).

This will involve some refactoring and you might want to tackle #337 first.

Remember we are using UTC for timestamps!

Relevant xkcd: https://xkcd.com/2867

jamesturner246:
# Description

This change adds a new plugin class for time source devices. Consumers can choose between a machine-local time source, or an NTP server over the network.

I decided to break this feature request into a couple of PRs. This one merely adds the plugins for reading UNIX epoch time (timezone agnostic) from a choice of the local system time or an NTP server. The server host/IP can be configured in `finesse.config`. I will add the code to actually *use* the timestamp to correct everything in a separate PR.

the new classes are `HostTime` and `NTPTime`. Unit tests are added for both.

Fixes #488 

## Type of change

## Key checklist

## Further checks


finesse/hardware/plugins/time/ntp_time.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Args:
+            ntp_host: The IP address or hostname of the NTP server
+        """
+        super().__init__()
+        self._client = NTPClient()
+        self._ntp_host = ntp_host
+        self._ntp_version = 3
+        self._ntp_port = "ntp"
+        self._ntp_timeout = 5
```
dc2917, line 32: Do we want this to be configurable, as we've done with the timeouts for other devices?

dc2917 (approved):
Nice! Looks good to me

finesse/hardware/plugins/time/ntp_time.py
jamesturner246: That's a good question. Also for the other fields. Maybe I'll just add them all and be done with it.

finesse/hardware/plugins/time/host_time.py
```diff
@@ -0,0 +1,15 @@
...
+from finesse.hardware.plugins.time.time_base import TimeBase
+
+
+class HostTime(TimeBase, description="Host time"):
+    """A time source that queries the system time."""
+
+    def close(self) -> None:
+        """Close the connection to the device."""
+
+    def get_time(self) -> float:
```
dalonsoa, line 13: What the actual output (a float) means? Is it time from epoch or what? Maybe add some comment on that.

finesse/hardware/plugins/time/ntp_time.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def close(self) -> None:
+        """Close the connection to the device."""
+
+    def get_time(self) -> float:
```
dalonsoa, line 37: As above, what the actual output (a float) means? Is it time from epoch or what? Maybe add some comments.

dalonsoa (approved):
This looks good!

I've some concerns on the timezone, as usual. What actual time is any of this servers returning?

finesse/hardware/plugins/time/host_time.py
jamesturner246: Yeah, seconds since the epoch. I'll update the comments.

dc2917 (approved):

PR #521, opened 2024-02-22
Add an interface to FTSW500

### LINKED ISSUE #322: Add support for interfacing with ABB's spectrometer software
ABB's software is available in the RDS along with some documentation and a program for generating simulated spectrometer signals. It can be controlled via a simple TCP-based protocol. In addition to being able to start and stop recording via this protocol, it seems that a fairly large proportion of the program's functionality is also exposed this way (more so than the Bruker program used before).

dc2917:
# Description

This PR adds an interface to ABB's FTSW500 software.

A new spectrometer interface base device, `FTSW500InterfaceBase`, is added, which inherits from the abstract `SpectrometerBase` class. This provides the core `connect`, `start_measuring` and `stop_measuring` commands, which are executed when the corresponding buttons on the GUI are pressed.

The `FTSW500Interface` provides the underlying functionality for sending instructions to FTSW500, as well as receiving and handling any responses. The status of FTSW500 is translated into a `SpectrometerStatus` which largely provides a compatible set of states. The main difference is that once FTSW500 connects to an instrument, it begins acquiring data, but requires further instruction to write acquired data to file. This is in contrast to the interface to the EM27, which always writes acquired data to file, but does not begin acquiring until told to. (We could, of course, mimic this behaviour with FTSW500 by effectively making `connect` redundant, then having `start_measuring` perform `connect` + `start measuring`.)

I'm sure some of this can be tidied up but I thought I'd get some feedback first.

Fixes #322 

## Type of change

## Key checklist

- [X] The documentation builds without warnings (`mkdocs build -s`)

## Further checks

### DIFF
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -85,7 +85,7 @@ FTSW500_PORT = 7778
 FTSW500_POLLING_INTERVAL = 1.0
 """How long to wait between polls of the EM27's status."""
 
-FTSW500_TIMEOUT = 1.0
+FTSW500_TIMEOUT = 5.0
 """How long to wait for a response from FTSW500 for."""
 
 SPECTROMETER_TOPIC = "spectrometer"
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -76,6 +76,12 @@ device takes to reply. This value then determines how often we wait before even 
 a request.
 """
 
+FTSW500_HOST = "127.0.0.1"
+"""The IP address or hostname of the machine running the FTSW500 software."""
+
+FTSW500_PORT = 7778
+"""The port on which the TCP server of FTSW500 is listening."""
+
 SPECTROMETER_TOPIC = "spectrometer"
 """The topic name to use for spectrometer-related messages."""
 
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -83,7 +83,7 @@ FTSW500_PORT = 7778
 """The port on which the TCP server of FTSW500 is listening."""
 
 FTSW500_POLLING_INTERVAL = 1.0
-"""How long to wait between polls of the EM27's status."""
+"""How long to wait between polls of FTSW500's status."""
 
 FTSW500_TIMEOUT = 5.0
 """How long to wait for a response from FTSW500 for."""
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -82,6 +82,12 @@ FTSW500_HOST = "127.0.0.1"
 FTSW500_PORT = 7778
 """The port on which the TCP server of FTSW500 is listening."""
 
+FTSW500_POLLING_INTERVAL = 1.0
+"""How long to wait between polls of the EM27's status."""
+
+FTSW500_TIMEOUT = 1.0
+"""How long to wait for a response from FTSW500 for."""
+
 SPECTROMETER_TOPIC = "spectrometer"
 """The topic name to use for spectrometer-related messages."""
 
--- a/finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
+++ b/finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
@@ -1,0 +1,1 @@
+"""Provides a dummy FTSW500 interferometer device for interfacing with."""
--- a/finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
+++ b/finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
@@ -1,1 +1,0 @@
-"""Provides a dummy FTSW500 interferometer device for interfacing with."""
--- a/finesse/hardware/plugins/spectrometer/ftsw500_interface.py
+++ b/finesse/hardware/plugins/spectrometer/ftsw500_interface.py
@@ -49,7 +49,7 @@ def parse_response(response: bytes) -> SpectrometerStatus | None:
         SpectrometerStatus: the generic spectrometer device state of FTSW500
     """
     msg = response.decode()
-    if "NAK" in msg:
+    if msg.startswith("NAK"):
         if "&" in msg:
             try:
                 status = int(msg.split("&")[1])
@@ -58,7 +58,7 @@ def parse_response(response: bytes) -> SpectrometerStatus | None:
                 return None
         else:
             return None
-    elif "ACK" in msg:
+    elif msg.startswith("ACK"):
         if "&" in msg:
             try:
                 status = int(msg.split("&")[1])
--- a/finesse/hardware/plugins/spectrometer/ftsw500_interface.py
+++ b/finesse/hardware/plugins/spectrometer/ftsw500_interface.py
@@ -1,1 +1,194 @@
-"""Module containing code for sending commands to FTSW500 for the ABB spectrometer."""
+"""Module containing code for sending commands to FTSW500 for the ABB spectrometer.
+
+Communication is via TCP.
+
+The FTSW500 program must be running on the computer at FTSW500_HOST for the commands to
+work.
+"""
+
+import logging
+from socket import AF_INET, SOCK_STREAM, socket
+
+from PySide6.QtCore import QTimer
+
+from finesse.config import (
+    FTSW500_HOST,
+    FTSW500_POLLING_INTERVAL,
+    FTSW500_PORT,
+    FTSW500_TIMEOUT,
+)
+from finesse.hardware.plugins.spectrometer.ftsw500_interface_base import (
+    FTSW500Error,
+    FTSW500InterfaceBase,
+)
+from finesse.spectrometer_status import SpectrometerStatus
+
+
+def parse_response(response: bytes) -> SpectrometerStatus | None:
+    r"""Parse FTSW500's response.
+
+    The server always returns a response after having received a command. The response
+    will start with the tag "ACK" in case of success, or "NAK" otherwise, and is
+    terminated by the end-of-line character "\n". An answer can also contain a string
+    value or message after the tag, which will be preceded by "&". For example, querying
+    whether the instrument is currently measuring data would yield a response
+    "ACK&true\n" or "ACK&false\n".
+
+    Querying the FTSW500 state yields one of the following values:
+    0: when disconnected
+    1: when in the process of connecting to an instrument
+    2: when acquiring data without saving it
+    3: when acquiring and saving data
+    -1: when in an intermediate state that should normally not last for a long time
+        (less than 500 ms) or when the FTSW500_SDK object is not well initialized
+
+    Args:
+        response: the byte sequence received from FTSW500
+
+    Returns:
+        SpectrometerStatus: the generic spectrometer device state of FTSW500
+    """
+    msg = response.decode()
+    if "NAK" in msg:
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
+            except ValueError:
+                logging.error(f"{msg.split('&')[1][:-1]}")
+                return None
+        else:
+            return None
+    elif "ACK" in msg:
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
+            except ValueError:
+                logging.info(f"{msg.split('&')[1][:-1]}")
+                return None
+        else:
+            return None
+    else:
+        raise FTSW500Error("Unrecognised response")
+
+    if status == -1:
+        return SpectrometerStatus(1)
+    elif status in (0, 1, 2, 3):
+        return SpectrometerStatus(status)
+    else:
+        raise FTSW500Error("Unable to parse response")
+
+
+class FTSW500Interface(FTSW500InterfaceBase, description="FTSW500 spectrometer"):
+    """Interface for communicating with the FTSW500 program."""
+
+    def __init__(self) -> None:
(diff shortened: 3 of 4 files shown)


finesse/config.py
```diff
@@ -76,6 +76,18 @@
...
 """
 
+FTSW500_HOST = "127.0.0.1"
+"""The IP address or hostname of the machine running the FTSW500 software."""
+
+FTSW500_PORT = 7778
+"""The port on which the TCP server of FTSW500 is listening."""
+
+FTSW500_POLLING_INTERVAL = 1.0
+"""How long to wait between polls of the EM27's status."""
```
jamesturner246, line 86: Is this related to `EM27_SENSORS_POLL_INTERVAL` above? Could you just use that instead? Or maybe is `EM27` a typo?

finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
jamesturner246, line 1: Is the dummy interface coming in another PR?

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+                return None
+        else:
+            return None
+    elif "ACK" in msg:
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
+            except ValueError:
+                logging.info(f"{msg.split('&')[1][:-1]}")
+                return None
```
jamesturner246, line 67: Is it possible that multiple string arguments are returned? This code wouldn't pick up any after the first.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+        (less than 500 ms) or when the FTSW500_SDK object is not well initialized
+
+    Args:
+        response: the byte sequence received from FTSW500
+
+    Returns:
+        SpectrometerStatus: the generic spectrometer device state of FTSW500
+    """
+    msg = response.decode()
+    if "NAK" in msg:
```
jamesturner246, line 52: I think (personally) the following makes it clearer what a message looks like. Feel free to ignore -- kinda subjective.
```suggestion
    if msg[:3] == "NAK"
```

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
+            except ValueError:
+                logging.error(f"{msg.split('&')[1][:-1]}")
+                return None
+        else:
+            return None
+    elif "ACK" in msg:
```
jamesturner246, line 61: Same here:
```suggestion
    elif msg[:3] == "ACK":
```

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+            except ValueError:
+                logging.info(f"{msg.split('&')[1][:-1]}")
+                return None
+        else:
+            return None
+    else:
+        raise FTSW500Error("Unrecognised response")
+
+    if status == -1:
+        return SpectrometerStatus(1)
```
jamesturner246, line 74: How come -1 changed to 1? I guess an intermediate state is kind of like connecting.

jamesturner246 (changes requested):
Looking good. 

I put some questions below, and a kinda subjective suggestion that I'm not worried about being ignored.

Would be good to have some unit tests for this class though. Presumably they'd be very similar to the other spectrometer class.

finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
dc2917: I'm not sure what to do about this. I'm not sure it's needed, as the dummy OPUS is generic enough that it's essentially just a dummy spectrometer.

finesse/config.py
dc2917: Good spot 😅

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dc2917: Yeah, good idea

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
+            except ValueError:
+                logging.error(f"{msg.split('&')[1][:-1]}")
+                return None
+        else:
+            return None
+    elif "ACK" in msg:
+        if "&" in msg:
```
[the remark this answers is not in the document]
dc2917: It shouldn't appear anywhere else, no. Perhaps it's safer to look for `ACK&` and `NAK&`?

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+        response: the byte sequence received from FTSW500
+
+    Returns:
+        SpectrometerStatus: the generic spectrometer device state of FTSW500
+    """
+    msg = response.decode()
+    if "NAK" in msg:
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
```
jamesturner246, line 55: If it's parsing `msg.split("&")[1] == 'false'` or `true`, wouldn't the integer cast fail?

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dc2917: This is because `SpectrometerStatus` is based on EM27's statuses, so this just converts a FTSW500 status to a(n apparently) generic spectrometer status

finesse/hardware/plugins/spectrometer/dummy_ftsw500_interface.py
jamesturner246: Ah, fair enough. Can probably just remove this then, right?

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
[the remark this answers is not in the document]
jamesturner246: Ah yes, just in case `msg` contains `ACK` or `NAK`, right? true.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
jamesturner246: I see. okay.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dc2917: Yeah the `except ValueError:` is there to catch non-integer messages. This should only be used for obtaining the status, so if there's a non-integer message something must have gone awry, so it's sent to the logger

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dalonsoa: What about:
```suggestion
    if msg.startswith("NAK"):
```

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dalonsoa: Indeed...
```suggestion
    elif msg.startswith("ACK"):
```

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
jamesturner246: Aahhh, I see... Maybe add a comment or something to that effect.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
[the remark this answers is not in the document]
dc2917: `msg` will always contain "ACK" or "NAK", and the "&" we want to search for is a delimiter which should always come after the "ACK" or "NAK". So by instead searching for "ACK&" and "NAK&" we can be sure that we're getting the delimiting "&".

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dc2917: Yeah, it felt a bit weird to do. I felt the base spectrometer device framework a bit restrictive but thought I'd see what I could do without changing it.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
jamesturner246: There really is a Python method for everything!

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+    Returns:
+        SpectrometerStatus: the generic spectrometer device state of FTSW500
+    """
+    msg = response.decode()
+    if msg.startswith("NAK"):
+        if "&" in msg:
+            try:
+                status = int(msg.split("&")[1])
+            except ValueError:
+                logging.error(f"{msg.split('&')[1][:-1]}")
```
dalonsoa, line 57: Maybe add something more explicit like: 
```suggestion
                logging.error(f"Error while reading from FTSW500: {msg.split('&')[1][:-1]}")
```
so the user knows where the error is coming from.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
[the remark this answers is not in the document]
dalonsoa: If the code should only do something special (besides returning `None`) if there is `ACK&` or `NAK&`, then I would suggest to just look for that string (in particular using the `startswith` method)

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -0,0 +1,194 @@
...
+            self._requester.recv(1024)
+
+    def _regurgitate_nonmodal_dialog_message(self) -> None:
+        """Obtain the last non-modal message displayed on FTSW500 and log it."""
+        if self._check_is_nonmodal_dialog_open():
+            self._requester.sendall(b"getLastNonModalMessageDisplayed\n")
+            data = self._requester.recv(1024)
+            logging.info(f"FTSW500: {data.decode().split('&')[1][:-1]}")
+            self._requester.sendall(b"closeNonModalDialogMessage\n")
+            self._requester.recv(1024)
```
dalonsoa, line 142: While the messages sent are different, the logic in both cases seems to be the same. I'd suggest you combine both and use wrappers for each of them. Something like:

```suggestion
    def _is_dialog_open(self, check_modal: bool) -> bool:
        """Query whether FTSW500 has a dialog open."""
        msg = "" if check_modal else "Non"
        self._requester.sendall(f"is{msg}ModalMessageDisplayed\n".encode("utf-8"))
        data = self._requester.recv(1024)
        if data != b"" and data.decode().split("&")[1] == "true\n":
            return True
        return False
        
    def _regurgitate_dialog_message(self, is_modal: bool) -> None:
        """Obtain the last message displayed on FTSW500 and log it."""
        msg = "" if is_modal else "Non"
        self._requester.sendall(f"getLast{msg}ModalMessageDisplayed\n".encode("utf-8"))
        data = self._requester.recv(1024)
        logging.info(f"FTSW500: {data.decode().split('&')[1][:-1]}")
        self._requester.sendall(f"close{msg}ModalDialogMessage\n".encode("utf-8"))
        self._requester.recv(1024)

    def _regurgitate_modal_dialog_message(self) -> None:
        """Obtain the last modal message displayed on FTSW500 and log it."""
        if self._is_dialog_open(check_modal=True)
            self._regurgitate_dialog_message(is_modal=True)

    def _regurgitate_non_modal_dialog_message(self) -> None:
        """Obtain the last non modal message displayed on FTSW500 and log it."""
        if self._is_dialog_open(check_modal=False)
            self._regurgitate_dialog_message(is_modal=False)
```

dalonsoa (approved):
Keeping aside some tests, this looks good!

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dc2917: No, there should only ever be one string. Could potentially mess things up if the string contains a `'&'`, but if we instead search for "ACK&" and "NAK&" as suggested [above](https://github.com/ImperialCollegeLondon/FINESSE/pull/521#discussion_r1500560751) we should be able to account for that.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
dc2917: I'm tempted to remove the handling of modal dialog messages from our interface, as these generally mean user input is required in FTSW500 directly.

jamesturner246 (approved):
Looks good! Are you planning on adding tests in another PR?

PR #524, opened 2024-02-28
Log menu

### LINKED ISSUE #434: Add option to open log file (and logs dir?)
The old program provided an option to open the program's log file, which is handy for debugging.

FINESSE does save its log to disk, but this folder isn't somewhere that would be obvious to an end-user. It would be useful to have an option to open this log file somehow. While we could show it in a custom dialog within FINESSE, it is probably better to open it with the OS's default program (e.g. notepad), so that the user can save it elsewhere etc.

As a bonus, we could have an option to open the folder with the log files in it.

CWestICL:
`Logs` added to the main window toolbar menu with two options:
- `Open log` - Opens the current log file in the default text editor
- `Open file location` - Opens the log file dir in File Explorer

Closes #434

### DIFF
--- a/finesse/gui/logs_view.py
+++ b/finesse/gui/logs_view.py
@@ -4,7 +4,7 @@ import os
 from PySide6.QtCore import QObject, QUrl
 from PySide6.QtGui import QAction, QDesktopServices
 
-from finesse.logger import get_log_path, log_file
+from finesse.logger import get_log_path
 
 
 class LogOpen(QAction):
@@ -21,6 +21,8 @@ class LogOpen(QAction):
 
     def open_log(self) -> None:
         """Opens the current log file."""
+        from finesse.logger import log_file
+
         QDesktopServices.openUrl(QUrl.fromLocalFile(log_file))
 
 
--- a/finesse/gui/logs_view.py
+++ b/finesse/gui/logs_view.py
@@ -1,11 +1,12 @@
 """Code for FINESSE's documentation viewer window."""
 import os
-import webbrowser
+import platform
+import subprocess
 
 from PySide6.QtCore import QObject, QUrl
 from PySide6.QtGui import QAction, QDesktopServices
 
-from finesse.logger import get_log_path
+from finesse.logger import filename, get_log_path
 
 
 class LogOpen(QAction):
@@ -22,12 +23,12 @@ class LogOpen(QAction):
 
     def open_log(self) -> None:
         """Opens the current log file."""
-        self.path = os.path.realpath(get_log_path())
-        self.files = [
-            f"{self.path}/{x}" for x in os.listdir(self.path) if x.endswith(".log")
-        ]
-        self.log = max(self.files, key=os.path.getctime)
-        webbrowser.open(self.log)
+        if platform.system() == "Darwin":
+            subprocess.call(("open", filename))
+        elif platform.system() == "Windows":
+            os.startfile(filename)
+        else:
+            subprocess.call(("xdg-open", filename))
 
 
 class LogLocationOpen(QAction):
--- a/finesse/gui/logs_view.py
+++ b/finesse/gui/logs_view.py
@@ -1,4 +1,4 @@
-"""Code for FINESSE's documentation viewer window."""
+"""Code for FINESSE's logs menu."""
 import os
 
 from PySide6.QtCore import QObject, QUrl
--- a/finesse/gui/logs_view.py
+++ b/finesse/gui/logs_view.py
@@ -24,7 +24,7 @@ class LogOpen(QAction):
         """Opens the current log file."""
         self.path = os.path.realpath(get_log_path())
         self.files = [
-            f"{self.path}\\{x}" for x in os.listdir(self.path) if x.endswith(".log")
+            f"{self.path}/{x}" for x in os.listdir(self.path) if x.endswith(".log")
         ]
         self.log = max(self.files, key=os.path.getctime)
         webbrowser.open(self.log)
--- a/finesse/gui/logs_view.py
+++ b/finesse/gui/logs_view.py
@@ -1,0 +1,48 @@
+"""Code for FINESSE's documentation viewer window."""
+import os
+import webbrowser
+
+from PySide6.QtCore import QObject, QUrl
+from PySide6.QtGui import QAction, QDesktopServices
+
+from finesse.logger import get_log_path
+
+
+class LogOpen(QAction):
+    """A menu option for opening the current log file."""
+
+    def __init__(self, parent: QObject) -> None:
+        """Create a menu item for opening the current log file.
+
+        Args:
+            parent: the menu on which to place the menu item
+        """
+        super().__init__("Open log", parent)
+        self.triggered.connect(self.open_log)
+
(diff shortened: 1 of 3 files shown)


CWestICL:
I'm assuming the functionality for opening files/folders is OS agnostic but would be good to check! Works as expected on Windows.

jamesturner246:
Hi @CWestICL. Just quickly before I review, just had an error loading the log file on Linux:

```
  File "/home/jpturner/workspace/FINESSE/FINESSE/finesse/gui/logs_view.py", line 29, in open_log
    self.log = max(self.files, key=os.path.getctime)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen genericpath>", line 65, in getctime
FileNotFoundError: [Errno 2] No such file or directory: '/home/jpturner/.local/state/FINESSE/log\\20240215_16-17-26.log'
```

Looks like a Windows backslash file delimiter after `FINESSE/log`. Works if I change back to the OS's file delimiter.

CWestICL:
Ah ok, I've replaced the backslash. Could you check if it's working now?

finesse/gui/logs_view.py
```diff
@@ -0,0 +1,48 @@
...
+        self.triggered.connect(self.open_log)
+
+    def open_log(self) -> None:
+        """Opens the current log file."""
+        self.path = os.path.realpath(get_log_path())
+        self.files = [
+            f"{self.path}/{x}" for x in os.listdir(self.path) if x.endswith(".log")
+        ]
+        self.log = max(self.files, key=os.path.getctime)
+        webbrowser.open(self.log)
```
dc2917, line 30: I think `Pathlib` is the way to go here to ensure cross-platform compatibility. 

However, I think we could do something cleaner than searching for the latest file. If we stored `filename` (defined in `initialise_logging()`) globally in `finesse.logger`, we could just import `finesse.logger.filename` and then we know it'll be the current session's log, and it'll be a `Path`, so will be OS-independent. You'll still need to pass it as a string to `webbrowser.open()`, but still. Anyone else (@dalonsoa @jamesturner246) have an opinion on this?

CWestICL:
When I was researching how to open a file in the OS's text editor, I found a StackExchange post saying `webbrowser.open()` would work. And it does open NotePad on my Windows PC.

However I've just seen this in the Python docs:
`
Note that on some platforms, trying to open a filename using this function, may work and start the operating system’s associated program. However, this is neither supported nor portable.
`
So probably worth a check if it works on other OSs.

dc2917:
> When I was researching how to open a file in the OS's text editor, I found a StackExchange post saying `webbrowser.open()` would work. And it does open NotePad on my Windows PC.
> 
> However I've just seen this in the Python docs: `Note that on some platforms, trying to open a filename using this function, may work and start the operating system’s associated program. However, this is neither supported nor portable.` So probably worth a check if it works on other OSs.

It opens in my web browser. Seems weird that a something called `webbrowser.open()` opens a file in a text editor 😅 Does it at least open as read-only?

CWestICL:
It does not 😅 I might have to go back and find another way of doing this

finesse/gui/logs_view.py
```diff
@@ -0,0 +1,49 @@
...
+        self.triggered.connect(self.open_log)
+
+    def open_log(self) -> None:
+        """Opens the current log file."""
+        if platform.system() == "Darwin":
+            subprocess.call(("open", filename))
+        elif platform.system() == "Windows":
+            os.startfile(filename)
+        else:
+            subprocess.call(("xdg-open", filename))
```
dalonsoa, line 31: You should be able to use here exactly the same you use below:

```suggestion
        QDesktopServices.openUrl(QUrl.fromLocalFile(filename))
```
It does work for Windows. Worth to check in other platforms.

finesse/gui/logs_view.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Args:
+            parent: the menu on which to place the menu item
+        """
+        super().__init__("Open file location", parent)
+        self.triggered.connect(self.open_file_location)
+
+    def open_file_location(self) -> None:
+        """Opens the log file folder location."""
+        self.path = os.path.realpath(get_log_path())
```
dalonsoa, line 48: Is there a reason why it needs to be an attribute?
```suggestion
        path = os.path.realpath(get_log_path())
```

finesse/logger.py
```diff
@@ -8,12 +8,19 @@
...
+def get_log_path():
+    """Return the user log path."""
     log_path = user_log_path(config.APP_NAME, config.APP_AUTHOR)
     log_path.mkdir(parents=True, exist_ok=True)
-    filename = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
+    return log_path
+
+
+log_path = get_log_path()
+filename = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
```
dalonsoa, line 19: We don't need `log_path` to be a global variable and, now that is used in other places, we need to give `filename` a more descriptive name:
```suggestion
log_file = get_log_path() / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
```

dalonsoa (changes requested):

finesse/logger.py
dc2917: I had in mind something like:

```py
"""in finesse/logger.py"""
...

from pathlib import Path

...

log_file: Path

...

def initialise_logging() -> None:
    """Configure the program's logger."""
    log_path = get_log_path()

    global filename
    log_file = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
...
```

That way the log filename is still defined when logging is initialised rather than when the module is imported

dc2917 (changes requested):

finesse/gui/logs_view.py
jamesturner246: Agree. `pathlib` is preferable, and covert to string only at the last moment.

dalonsoa (approved):
Looks good to me!

finesse/gui/logs_view.py
```diff
@@ -0,0 +1,44 @@
+"""Code for FINESSE's documentation viewer window."""
```
jamesturner246, line 1: Update docstring here.

finesse/logger.py
```diff
@@ -2,17 +2,28 @@
...
-    filename = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
+    return log_path
+
+
+def initialise_logging() -> None:
+    """Configure the program's logger."""
+    log_path = get_log_path()
+
+    global log_file
+    log_file = log_path / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
```
jamesturner246, line 26:
```suggestion
    global log_file
    log_file = get_log_path() / f"{datetime.now().strftime('%Y%m%d_%H-%M-%S')}.log"
```

jamesturner246 (approved):
A couple of trivial things, but looks good!

dc2917 (approved):

PR #535, opened 2024-03-21
Sync timestamps in data file

### LINKED ISSUE #522: Apply correct timestamps to data

jamesturner246:
# Description

This change ensures that timestamps for temperature readings in the data file are synchronised using the time source (host or NTP server). The temperature monitor pubsub send call is updated to retrieve time from the time source before publishing.

Fixes #522. 

## Type of change

## Key checklist

## Further checks


finesse/hardware/data_file_writer.py
```diff
@@ -156,7 +164,7 @@ def close(self) -> None:
     def write(self, time: datetime, temperatures: list[Decimal]) -> None:
         """Write temperature readings to the CSV file."""
         # Also include timestamp as seconds since midnight
-        midnight = datetime(time.year, time.month, time.day)
+        midnight = datetime(time.year, time.month, time.day, tzinfo=UTC)
         secs_since_midnight = floor((time - midnight).total_seconds())
```
dalonsoa, line 168: This will produce a warning if `time` does not have timezone information since it is comparing it with a datetime object that do have such information. Where is `time` coming from? Should it be the time taken from the time device, as well?

finesse/hardware/plugins/time/__init__.py
```diff
@@ -1 +1,9 @@
 """This module contains interfaces for querying time sources."""
+
+from finesse.hardware.manage_devices import get_device_instance
+from finesse.hardware.plugins.time.time_base import TimeBase
+
+
+def get_time_instance() -> TimeBase | None:
+    """Get the instance of the current time source if connected or None."""
+    return get_device_instance(TimeBase)
```
dalonsoa, line 9: Justo to confirm I get this right, this call will return the appropriate time device based on the configuration (the host time or the NTP server) regardless of not having any indication here of what device in particular should be provided, right?

dalonsoa (approved):
The code itself looks ok, but I have a couple of questions that will be useful to clarify.

finesse/hardware/data_file_writer.py
jamesturner246: Fair point. `time` comes as a pubsub message from `finesse.hardware._send_temperatures`, which I have set to UTC time there, but maybe some other function will try to use it with a timezone-less `datetime`. Since it's all UTC time anyway, maybe it's best to just use timezone-less datetimes everywhere.

finesse/hardware/plugins/time/__init__.py
jamesturner246: Right. It's the (arbitrary) time device set in manage devices, or the 'hardware set' yaml files. If a time device is not connected (with 'connect' button or device manager), then other parts if the code default to epoch zero time. Not sure of a more sensible default.

finesse/hardware/__init__.py
```diff
@@ -53,11 +53,10 @@ def _send_temperatures() -> None:
...
 
     # Get time from the time source.
-    epoch_time = _try_get_time()
-    if epoch_time is None:
+    time = _try_get_time()
+    if time is None:
         # On failure, set time to the UNIX epoch.
-        epoch_time = 0.0
-    time = datetime.fromtimestamp(epoch_time)
+        time = datetime.fromtimestamp(0.0)
```
dalonsoa, line 59: Add this to `try_get_time` so it always returns a `datetime` object.

finesse/hardware/data_file_writer.py
```diff
@@ -35,7 +35,7 @@ def _get_metadata(filename: str) -> dict[str, Any]:
         logging.warning("No time device connected. Using system time.")
         time = datetime.now()
     else:
-        time = datetime.fromtimestamp(time_device.get_time())
```
dalonsoa: This is a bit confusing. Above, if there is a problem getting the time, the epoch is returned. But here, if there's a problem, `now` is returned. This inconsistency might cause problem. I am not sure what the answer should be, but I would choose one and stick to it consistently across the code.

dalonsoa (changes requested):
I'm sorry for requesting changes again, but the more I think about it, the  more I think that:
- There must be always a time device. If not set explicitly, it will be set to the `host_time` one (with the appropriate mention to the logs).
- If the chosen time device fails to provide a time for whatever reason, it should revert to get the time from the host time, logging the error. But there must always be a time!
- A time device will always return a time, being this `now` or epoch, but we need to be consistent on what is returned. I think that `now` makes more sense otherwise, it will be the same time all the time, which is not useful and cannot be converted back to the right time retrospectively.

finesse/hardware/__init__.py
alexdewar: +1. Then you can rename it to just `_get_time()`.

finesse/gui/temperature_plot.py
```diff
@@ -80,7 +80,7 @@ def _create_figure(self) -> None:
         hot_bb_temp: list[float | None] = [None] * self._figure_num_pts
         cold_bb_temp: list[float | None] = [None] * self._figure_num_pts
 
-        t[0] = datetime.now().timestamp()
+        t[0] = 0.0
```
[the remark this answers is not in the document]
alexdewar: It *might* work to set it to `None`.

alexdewar:
Sorry to jump in here unbidden :wink:... but just wanted to add that I agree with @dalonsoa. On reflection, I'm not sure it makes sense to have a separate `HostTime` device type, because you always need some time source, regardless of whether there's a time device connected. Maybe it would be better to have time devices return the time *offset* instead? Then you could drop the `HostTime` class and just have the offset default to 0.

On a related note, there is another issue with how things are being done currently: the NTP server is being queried whenever the time is requested (around every 2 secs). Normally you only resynchronise with an NTP server occasionally (e.g. every 10 mins) and assume that your clock hasn't drifted too much in the meantime.

I think we should probs close this PR and address these issues with the time devices first. I'm not sure what your plans are for the rest of the sprint @jamesturner246, but I'm at a bit of a loose end this week so happy to take this on.

jamesturner246:
I've removed the `HostTime` class, and finesse time now defaults to local time.

I've also changed `NTPTime` so that it queries periodically rather than every time it's needed.

Tests have been updated.

Looks like there's some conflicts. I'll sort them out when tests pass.

finesse/gui/temperature_plot.py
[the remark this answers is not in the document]
alexdewar: @jamesturner246 This needs looking at... unfortunately it won't work on Windows.

finesse/hardware/data_file_writer.py
```diff
@@ -33,7 +34,7 @@ def _get_metadata(filename: str) -> dict[str, Any]:
     return {
         "encoding": "utf-8",
         "name": filename,
-        "datetime": datetime.now().astimezone().isoformat(),  # include timezone
+        "datetime": get_time().isoformat(),
```
alexdewar, line 37: I'm not sure about this change. I know we're using UTC time everywhere else, but I deliberately wanted to save the localtime in the metadata (i.e. with timezone included) because I think that's useful information. Jon does sometimes take the equipment abroad; not sure if he changes the timezone when he does though.

While I can see the argument for using the same timesource here as for the actual data, I'd be inclined to leave it as is.

finesse/hardware/plugins/time/__init__.py
```diff
@@ -1 +1,23 @@
...
+def get_time() -> datetime:
+    """Read the current time, with optional time offset.
+
+    If the time device is not connected or the operation fails, local time is returned.
+    """
+    timestamp = datetime.now().timestamp()
+    if dev := get_time_instance():
+        timestamp += dev.get_time_offset()
+
+    return datetime.fromtimestamp(timestamp)
```
alexdewar, line 23: Instead of converting back and forth between epoch, you could do:

```suggestion
    time = datetime.now()
    if dev := get_time_instance():
        return time + timedelta(seconds=dev.get_time_offset())

    return time
```

(You need `from datetime import timedelta`)

finesse/hardware/plugins/time/time_base.py
```diff
@@ -10,9 +10,9 @@ class TimeBase(Device, name=TIME_TOPIC, description="Time source"):
     """The base class for time source devices or mock devices."""
 
     @abstractmethod
-    def get_time(self) -> float:
-        """Get the current time.
+    def get_time_offset(self) -> float:
+        """Get the current time offset.
```
alexdewar, line 14:
```suggestion
        """Get the current time offset in seconds.
```

finesse/hardware/plugins/time/__init__.py
```diff
@@ -1 +1,23 @@
...
+from finesse.hardware.manage_devices import get_device_instance
+from finesse.hardware.plugins.time.time_base import TimeBase
+
+
+def get_time_instance() -> TimeBase | None:
+    """Get the instance of the current time source if connected or None."""
+    return get_device_instance(TimeBase)
+
+
+def get_time() -> datetime:
```
alexdewar, line 14: This is a bit pedantic, but maybe this function should be called `get_current_time()`? Otherwise the meaning may not always be clear from context.

finesse/hardware/plugins/time/ntp_time.py
```diff
@@ -49,23 +54,33 @@ def __init__(
...
+        self.poll_time_offset()
+        self._poll_timer = QTimer()
+        self._poll_timer.timeout.connect(self.poll_time_offset)
+        self._poll_timer.start(int(ntp_poll_interval * 1000))
 
-        Returns:
-            The current time in seconds since the epoch.
-        """
+    def poll_time_offset(self) -> None:
+        """Query the NTP server for the current time offset."""
```
alexdewar, line 64: This looks good, but I feel like this function could do with some logging. It would be useful to know when the NTP server is being polled and, if successful, what the offset is. Given that we won't be polling too often, this shouldn't create much log spam.

finesse/hardware/plugins/time/ntp_time.py
```diff
@@ -49,23 +54,33 @@ def __init__(
...
         try:
-            response = self._client.request(
+            self._response = self._client.request(
                 self._ntp_host,
                 version=self._ntp_version,
                 port=self._ntp_port,
```
alexdewar, line 69: I've only just noticed that this is a string, which seems a bit weird. I'm guessing the string `"ntp"` is just an alias for the default port (123). If so, I'd just use the number. Also, the typing does matter here, because the frontend casts device parameters to whatever type it thinks they should be and I'm guessing that a string value of e.g. `"1234"` won't work.

finesse/hardware/plugins/time/ntp_time.py
```diff
@@ -49,23 +54,33 @@ def __init__(
...
 
-        return response.dest_time + response.offset
+        Returns:
+            A float representing the current time offset.
+        """
+        return self._response.offset
+
+    def close(self) -> None:
+        """Close the device."""
+        self._poll_timer.stop()
```
alexdewar, line 86: You also need to explicitly call the parent class's `close()` because there is some `Device`-specific cleanup that needs to happen:

```suggestion
        self._poll_timer.stop()
        super().close()
```

tests/hardware/plugins/time/test_ntp_time.py
```diff
@@ -2,41 +2,74 @@
...
+
+    assert ntp_time._ntp_host == "test.org"
+    assert ntp_time._ntp_version == 2
+    assert ntp_time._ntp_port == "test_port"
+    assert ntp_time._ntp_timeout == 123.0
+
+    # Interval is in milliseconds, not seconds.
+    assert ntp_time._poll_timer.interval() == 456000
+
+    request_mock.assert_called_once()
```
alexdewar, line 27: I'd go with `assert_called_once_with()` and give the expected args.

alexdewar (changes requested):
This is looking much better. Good work! I've got some minor suggestions, but I think the design is cleaner now.

One other question: why is the default NTP protocol version 3 rather than 4?

I also found a small bug. If you try to connect to an NTP server that doesn't exist, the device still *appears* to connect successfully and things only fall over when the timer triggers a request, at which point you start repeatedly getting uncaught exceptions that look like this:

```
  File "/home/alex/code/FINESSE/finesse/gui/temperature_monitor_view.py", line 94, in _poll_device
    pub.sendMessage(f"device.{TEMPERATURE_MONITOR_TOPIC}.data.request")
  File "/home/alex/code/FINESSE/.venv/lib/python3.11/site-packages/pubsub/core/publisher.py", line 216, in sendMessage
    topicObj.publish(**msgData)
  File "/home/alex/code/FINESSE/.venv/lib/python3.11/site-packages/pubsub/core/topicobj.py", line 452, in publish
    self.__sendMessage(msgData, topicObj, msgDataSubset)
  File "/home/alex/code/FINESSE/.venv/lib/python3.11/site-packages/pubsub/core/topicobj.py", line 482, in __sendMessage
    listener(data, self, allData)
  File "/home/alex/code/FINESSE/.venv/lib/python3.11/site-packages/pubsub/core/listener.py", line 237, in __call__
    cb(**kwargs)
  File "/home/alex/code/FINESSE/finesse/hardware/__init__.py", line 40, in _send_temperatures
    time = get_time()
           ^^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/time/__init__.py", line 21, in get_time
    timestamp += dev.get_time_offset()
                 ^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/time/ntp_time.py", line 87, in get_time_offset
    return self._response.offset
           ^^^^^^^^^^^^^^
AttributeError: 'NTPTime' object has no attribute '_response'
```

The trick is to call `poll_time_offset()` once in `__init__()` before you start the timer. If an exception is raised at this point, then the user will get an appropriate error message and the device won't be marked as open.

dalonsoa (approved):
This looks good to me, but I'm unsure if we are producing the time information unambiguously.

Is the timezone included somehow in the time information saved in the data files and the logs? If so, do we want to produce that data in whatever is the local timezone or always in UTC?

finesse/gui/temperature_plot.py
```diff
@@ -81,7 +81,7 @@ def _create_figure(self) -> None:
         hot_bb_temp: list[float | None] = [None] * self._figure_num_pts
         cold_bb_temp: list[float | None] = [None] * self._figure_num_pts
 
-        t[0] = datetime.now().timestamp()
+        t[0] = None
         hot_bb_temp[0] = 25.0
         cold_bb_temp[0] = 25.0
```
alexdewar, line 87: I *think* these lines can just be deleted altogether:

```suggestion
```

@dc2917 is there a reason we need an initial temperature set here? It seems a little odd. I've tried deleting these lines locally and things still seem to work.

alexdewar (approved):
This is great! I've tested it locally and it all seems to work fine.

I've got one query (more for @dc2917 than you really), but other than that I think we're good to merge.

alexdewar:
@dalonsoa We had this discussion with Jon a while back. He wants everything to be in UTC (as it is is for their current software, I believe).

There is one place where we record the timezone: in the metadata header for the CSV files. I figured having localtime was more informative in that context (e.g. if you're trying to remember when you recorded a dataset). The timestamps in the actual CSV part of the file are still UTC though.

finesse/gui/temperature_plot.py
dc2917: Yeah, I think this used to be required but the `pop` and `append` in `_update_figure` presumably now take care of whatever the problem was.

dc2917 (approved):

finesse/gui/temperature_plot.py
dc2917: It would be nice if a Windows-user could check they don't have any issues though (i.e. as in https://github.com/ImperialCollegeLondon/FINESSE/issues/220).

dalonsoa: I guess the answer is "it does not work on Windows". I get the following when trying to launch FINESSE:

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\__main__.py", line 6, in <module>
    run()
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\__init__.py", line 24, in run
    window = MainWindow()
             ^^^^^^^^^^^^
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\gui\main_window.py", line 89, in __init__
    bb_monitor: QGroupBox = TemperaturePlot()
                            ^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\gui\temperature_plot.py", line 34, in __init__
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\gui\temperature_plot.py", line 62, in _create_controls
    self._create_figure()
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\gui\temperature_plot.py", line 104, in _create_figure
    self._canvas.draw()
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\backends\backend_agg.py", line 388, in draw
    self.figure.draw(self.renderer)
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\artist.py", line 95, in draw_wrapper
    result = draw(artist, renderer, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\artist.py", line 72, in draw_wrapper
    return draw(artist, renderer)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\figure.py", line 3148, in draw
    self.get_layout_engine().execute(self)
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\layout_engine.py", line 273, in execute
    return do_constrained_layout(fig, w_pad=w_pad, h_pad=h_pad,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\_constrained_layout.py", line 116, in do_constrained_layout
    make_layout_margins(layoutgrids, fig, renderer, h_pad=h_pad,
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\_constrained_layout.py", line 381, in make_layout_margins
    pos, bbox = get_pos_and_bbox(ax, renderer)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\_constrained_layout.py", line 634, in get_pos_and_bbox
    tightbbox = martist._get_tightbbox_for_layout_only(ax, renderer)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\artist.py", line 1411, in _get_tightbbox_for_layout_only
    return obj.get_tightbbox(*args, **{**kwargs, "for_layout_only": True})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\_api\deprecation.py", line 454, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\axes\_base.py", line 4395, in get_tightbbox
    ba = martist._get_tightbbox_for_layout_only(axis, renderer)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\artist.py", line 1411, in _get_tightbbox_for_layout_only
    return obj.get_tightbbox(*args, **{**kwargs, "for_layout_only": True})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\axis.py", line 1334, in get_tightbbox
    ticks_to_draw = self._update_ticks()
                    ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\axis.py", line 1276, in _update_ticks
    major_labels = self.major.formatter.format_ticks(major_locs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\ticker.py", line 216, in format_ticks
    return [self(value, i) for i, value in enumerate(values)]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\ticker.py", line 216, in <listcomp>
    return [self(value, i) for i, value in enumerate(values)]
            ^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\AppData\Local\pypoetry\Cache\virtualenvs\finesse-VOhfgRFV-py3.11\Lib\site-packages\matplotlib\ticker.py", line 323, in __call__
    return self.func(x, pos)
           ^^^^^^^^^^^^^^^^^
  File "C:\Users\dalonsoa\Projects\FINESSE\finesse\gui\temperature_plot.py", line 160, in _xtick_format_fcn
    return datetime.fromtimestamp(val).strftime("%H:%M:%S")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 22] Invalid argument

```

dc2917: So yeah, issue #220 then. 

As I mentioned in that issue, we can set t[0] to whatever we want, as long as it doesn't cause any of the x tick values to be less than 0. I just chose `datetime.now()` at the time, but we could just do midday or something.

alexdewar: Good catch @dc2917! I don't think midday would work if you were recording *before*
midday though... Did you mean midnight?

dc2917: It only causes a problem if the argument to `datetime.fromtimestamp()` is a negative float. `_xtick_format_fcn` turns a (float) timestamp into a string via `datetime.fromtimestamp(val).strftime("%H:%M:%S")`.

If you create a plot with empty x and y data, matplotlib centres the plot on (0,0). So if we don't set `t[0]`, the x axis will be centred on 0, so x ticks left of centre will be negative (e.g. `datetime.fromtimestamp(-0.1).strftime("%H:%M:%S")`), raising the error.

Midday any day on or after 1st Jan 1970 would work. Midnight any day after (but not on) 1st Jan 1970 would work. `datetime.now()` seems the most logical to me, but it could literally be any timestamp after 00:03:00 on 1st Jan 1970 (matplotlib creates x-ticks at 00:00:00, 00:01:00, 00:02:00, 00:03:00, ..., 00:06:00 if `t[0]` corresponds to 00:03:00, so anything before 00:03:00 would make the first x tick before 00:00:00 on 1st Jan 1970, i.e. a negative float).

The value is never recorded anywhere, it's just for matplotlib to be able to initialise the figure. Once the connection to the temperature monitors takes place, `_update_figure` is called and these initial x and y values are `pop`ped from their respective lists.

alexdewar: I seeeeeee... Thanks for the explanation!

Would it work to do:

```py
        return datetime.fromtimestamp(max(val, 0)).strftime("%H:%M:%S")
```

dc2917: Yeah, I guess that would just set all the ticks strings to 00:00:00, and we could remove setting the initial values.

jamesturner246: Okay, so I'll change the return value of `_xtick_format_fcn` to `datetime.fromtimestamp(max(val, 0)).strftime("%H:%M:%S")` and remove the initialisation in `_create_figure` and I think we're good.

dalonsoa:
Works well on Windows, now.

Issue #534, opened 2024-03-13
Tell FTSW500 desired output file location


dc2917:

alexdewar:
I've just been looking at this. Unfortunately, there's a bug in FTSW500 which means if you try to set the output path prefix to an absolute *nix path (i.e. starting with /) then you get an error message about the path being invalid. Presumably the fix on their end is to just use whatever the proper path handling functions are in Java rather than some open-coded hackery that only works on Windows. (In theory they're only targetting Windows, but that's still no excuse :wink:.)

So let's put this on ice for now and see if we can get ABB to fix up FTSW500 for us.

PR #548, opened 2024-04-09
Tidy-up of `FTSW500Interface` code

### LINKED ISSUE #539: Uncaught exceptions when connection to FTSW closes unexpectedly
To reproduce:

1. Start FTSW500
2. Open connection from FINESSE using "Manage devices" dialog
3. Close FTSW500 program window

Result:

- A variety of uncaught exceptions

Workaround:

- Don't do this :upside_down_face:. In the normal course of events, users shouldn't be closing FTSW500 while FINESSE is interacting with it, *however* we should handle this case cleanly, partly for UX reasons, but also for the sake of robustness: e.g. we don't want data recording to stop because a connection is lost to a device (not sure if that would actually happen here).

The most common error seems to be this:

```
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 196, in _request_status
    self.request_command(b"getFTSW500State\n")
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 204, in request_command
    self._make_request(command)
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 191, in _make_request
    self._requester.sendall(command)
BrokenPipeError: [Errno 32] Broken pipe
```

But I also sometimes see:

```
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 196, in _request_status
    self.request_command(b"getFTSW500State\n")
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 204, in request_command
    self._make_request(command)
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 192, in _make_request
    self._on_reply_received(self._requester.recv(1024))
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 186, in _on_reply_received
    self._regurgitate_nonmodal_dialog_message()
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 147, in _regurgitate_nonmodal_dialog_message
    if self._check_is_nonmodal_dialog_open():
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/code/FINESSE/finesse/hardware/plugins/spectrometer/ftsw500_interface.py", line 131, in _check_is_nonmodal_dialog_open
    if data.decode().split("&")[1] == "true\n":
       ~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

`BrokenPipeError`s should be caught with the error propagated to the fronted via a `pubsub` message (e.g. by using the `pubsub_errors()` decorator). In the second error I mention though, the code just needs to be written a bit more defensively. The send and receive operations for `FTSW500Interface` *could* possibly be encapsulated in their own functions with error handling taking place there (plus, having `self._requester.recv(1024)` dotted all over the place smells a bit wrong to me).

Tagging @dc2917

### LINKED ISSUE #540: FTSW code doesn't handle status of -1 correctly
From the Java API documentation:

> -1 = when the state is an intermediate state that should normally not last for a long time (less than 500 ms) or when the FTSW500_SDK object is not well initialized
> 
> Note: If the actual state is a possible internal intermediate states that should not last for a long time, a sleep of 500 ms is done before to get the following state.

At present our code just assumes that -1 means that FTSW is connecting, which according to the docs is not correct:

```py
    if status == -1:
        return SpectrometerStatus(1)
    elif status in (0, 1, 2, 3):
        return SpectrometerStatus(status)
    else:
        raise FTSW500Error("Unable to parse response")
```

It seems like the proper way to handle this is to retry the request after a short delay up to a maximum of 500 ms.

That said, this might not actually create problems in practice. @dc2917 have you seen a status of -1 returned in the wild?

alexdewar:
# Description

I've been playing around with ABB's FTSW500 program over the last few days and in the process came across some rough edges in our code to interface with it.

One problem is that there is currently no error handling so any exceptions caused by a closed socket (or whatever) will just lead to the "uncaught exception" dialog displaying (e.g. see #539). The way this should be handled is to send a device error message via `pubsub`, which leads to a) a more user-friendly error message on the frontend and b) the device being closed properly (FINESSE currently treats all device errors as fatal). Otherwise both FINESSE and the device will likely limp on in an inconsistent state with unpredictable results. (I should probably have documented somewhere that this is the right way to handle device errors -- sorry!)

Additionally, there were a few places where the code could have been better modularised with better separation of concerns. For example, the `parse_response()` function assumes that the return value accompanying an `ACK` response always represents a `SpectrometerStatus` and returns `None` if parsing this fails; the return value can be anything though and it is only a `SpectrometerStatus` for the `getFTSW500State` command. Perhaps because of this, there were lots of places where functions were using the socket object directly to send/receive data, but without checking the response, meaning no error checking was happening (i.e. a `NAK` response would be ignored). The FTSW500's protocol is actually a pretty conventional RPC system: you tell it what command to run (with optional arguments) and it gives a response indicating either success (optionally with extra info) or failure, so it makes sense to have a single function to send requests and parse the responses. (I think part of the problem here is that the existing OPUS code was probably used as a base and the OPUS API is a horrid mess that was clearly designed by sociopaths :wink:.)

There's one last thing I thought about changing but have left as is for now: currently the code reads the contents of any open dialog boxes in FTSW500 (you can request this info via the API) and writes it to FINESSE's program log. The problem is that after the messages have been read, the dialog boxes are all closed, meaning that the user won't be aware of what they said unless they dig into FINESSE's logs, which I don't think is ideal. I'm not entirely sure why this was being done in the first place -- was it a way of recording error messages? If so, I'd be inclined to remove it as any command that fails should yield a `NAK` response that FINESSE will intercept anyway.

Finally, I've removed a bunch of functions that aren't currently being used, partly because they don't have proper error checking and partly so that they don't clutter up the code base. If we need them at a later stage, we can add the functionality back; the functions should only be a couple of lines long now.

Once everyone's happy with the code in this PR I'll add some tests too.

Fixes #540. Fixes #539.

## Type of change

## Key checklist

## Further checks


finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -70,135 +60,141 @@ def __init__(self) -> None:
...
+        The status is requested after sending the command so that we are immediately
+        notified if it has changed in response to the command (e.g. recording has
+        started). Contents of dialogs in FTSW500 are saved to the program log.
+
         Args:
             command: Name of command to run
         """
-        self._make_request(command)
+
+        def internal():
```
dalonsoa, line 189: Why making this an internal function? That makes it directly untestable.

dalonsoa (approved):
LGTM! Just made a comment on the internal function.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
alexdewar: Good point! I'll fix that up

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -70,135 +60,141 @@ def __init__(self) -> None:
...
-        """Set the output prefix for FTSW500 data."""
-        self._requester.sendall(f"setOutputPrefixOnlyField&{output_prefix}\n".encode())
-        self._requester.recv(1024)
-
-    def set_data_description(self, data_description: str) -> None:
-        """Set the description for FTSW500 data."""
-        self._requester.sendall(
-            f"setDataDescriptionField&{data_description}\n".encode()
-        )
-        self._requester.recv(1024)
```
dc2917, line 104: These will be required for #534

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -70,135 +60,141 @@ def __init__(self) -> None:
...
-        """Set the number of measurements for FTSW500 to record."""
-        self._requester.sendall(
-            f"setNumberMeasurementField&{num_measurements}\n".encode()
-        )
-        self._requester.recv(1024)
-
-    def set_temperature(self, temperature: float) -> None:
-        """Set the temperature for FTSW500 data."""
-        self._requester.sendall(f"setTemperatureField&{temperature}\n".encode())
-        self._requester.recv(1024)
```
dc2917, line 116: While this is currently unused, Jon mentioned it would be useful to be able to send the black-body temperatures to FTSW500 to compute spectra. Exactly how and how often we do this isn't clear, so we should discuss this with Jon.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
```diff
@@ -70,135 +60,141 @@ def __init__(self) -> None:
...
+        """Request the current status from FTSW500.
 
-        Args:
-            reply: the byte sequence received from the FTSW500 program
+        From the Java API documentation:
+
+        Querying the FTSW500 state yields one of the following values:
+            0: when disconnected
+            1: when in the process of connecting to an instrument
+            2: when acquiring data without saving it 3: when acquiring and saving data
```
dc2917, line 118:
```suggestion
            2: when acquiring data without saving it
            3: when acquiring and saving data
```

dc2917 (approved):
Looks great, this is a huge improvement haha. I've made a couple of comments about why there are currently unused functions there. I'm not sure if it's worth removing them entirely via this PR, but as you say, they're only short so can be added back when the associated features are implemented.

> ... currently the code reads the contents of any open dialog boxes in FTSW500 (you can request this info via the API) and writes it to FINESSE's program log. ... I'm not entirely sure why this was being done in the first place -- was it a way of recording error messages?

My intention was that we could use our GUI to control FTSW500 as much as possible, so that the user doesn't have to keep switching between programs. I therefore wanted these messages to be presented to the user via our program, rather than by FTSW500. However, this was a complete oversight on my part - I assumed the user would see the log output in the terminal, as developers do, but if they're running the executable that won't be the case 😅 so yes, your confusion is understandable! Maybe I intended to raise dialog windows in our program but didn't get round to it haha.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
alexdewar: Seeing as they needed tweaking anyway, I thought I'd just remove them for now and add them back in when we were going to use them.

finesse/hardware/plugins/spectrometer/ftsw500_interface.py
alexdewar: As above, I thought we could wait till we needed it.

alexdewar:
> I've made a couple of comments about why there are currently unused functions there. I'm not sure if it's worth removing them entirely via this PR, but as you say, they're only short so can be added back when the associated features are implemented.

In general, I'd personally avoid adding unused functions etc. to a code base when you could just add them in a later PR along with code that uses them, so the new functionality can be reviewed together. (Plus, if a function isn't actually called anywhere, it probably hasn't been tested all that thoroughly so might need later changes anyway.) As I said though, given that I would have had to have changed them to use `_make_request()` anyway, I thought I'd just leave them out for now.
 
> My intention was that we could use our GUI to control FTSW500 as much as possible, so that the user doesn't have to keep switching between programs. I therefore wanted these messages to be presented to the user via our program, rather than by FTSW500. However, this was a complete oversight on my part - I assumed the user would see the log output in the terminal, as developers do, but if they're running the executable that won't be the case 😅 so yes, your confusion is understandable! Maybe I intended to raise dialog windows in our program but didn't get round to it haha.

Jon will definitely have to switch between the programs anyway, e.g. to look at the graphs, so we can't do everything from FINESSE anyway (sadly :cry:). And given that the dialog windows should appear over FINESSE, I'm not sure what benefit we gain by trying to replicate this functionality. For the sake of simplicity, then, I'll just remove that logging for now.

PR #549, opened 2024-04-10
Data file UX improvements

### LINKED ISSUE #526: logfile prompt
I was running some test yesterday and I realised that it is quite easy to not start the logfile running. The button is to the lower right and non-descript. This was also true with the original logfile button and setup. I missed starting this logging on a couple of occassions.

It would be worthwhile to have a pop up window appear, particularly when a script file is running to advise the user if their have not started logging.

alexdewar:
# Description

@jonemurray has pointed out that it isn't currently obvious whether or not recording is taking place, so it is easy to forget to start recording before e.g. running a measure script.

I've tried to remedy this in two ways:

1. Make it more obvious when recording is happening by flashing a red LED and displaying some red text in a `QLabel`
2. Prompt the user to start data recording if they try to run a measure script when recording is not happening

Widget when recording is stopped:

Screenshot from 2024-04-10 at 12_30_48 080227005

Widget when recording is running (with flashing LED):

Screenshot from 2024-04-10 at 12_31_10 774311837

Suggestions on how to improve the appearance of the widget are welcome! (In particular, it looks a little odd the way the controls move around when the user starts/stops recording.) That said, I think it's perfectly usable as is :smile: 

Fixes #526.

I've broken tests (hence why this is a draft PR), but I'll fix them up if everyone is happy with this solution.

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
It looks perfectly fine to me - and to be honest, a very useful feature from the usability perspective. I guess things could be "frozen" in place playing around with alignment flags and that sort of things, but nothing comes immediately to my mind and I don't think it is worth investing too much effort in that.

dc2917 (commented):
This is a nice addition. I think some aspects of it are slightly counter-intuitive though. I feel like a red flashing light would typically be interpreted as a negative thing, so having it flash red when recording seems odd to me. Also, I don't think having it flash makes sense, especially with a different frequency to that which the data are written to file. Maybe it makes more sense to have a solid red light when not recording, and a solid green light when recording?

alexdewar:
A red light is often used to mean "recording in progress" though, e.g. on camcorders, so that was my logic. It's also more eye-catching than green IMO. That's the same reason I made the light blink, to draw the user's attention to it. I take your point that it might make more sense to sync the blinking with when data is actually recorded, though.

dc2917:
Ah yes, fair point!

alexdewar:
I've finished adding all the tests now, so we have 100% coverage for the changed files.

I also followed Dan's suggestion of flashing the LED when the data is actually being written.

PR #559, opened 2024-04-16
Make spectrometers configurable

### LINKED ISSUE #547: Expose some device parameters for `FTSW500Interface`
There are some parameters which could be configurable, namely:

- Hostname and port (this is likely to be localhost and 7778, respectively, but might as well make it configurable)
- The polling interval for the status timer

### LINKED ISSUE #557: Expose device parameters for `OPUSInterface`
Similar to #547, the device parameters for `OPUSInterface` are currently not configurable. We should fix this because it's easy enough to do and because having hardcoded IP addresses seems like unnecessary ugliness.

alexdewar:
# Description

The two spectrometer types (`FTSW500Interface` and `OPUSInterface`) have various parameters that users may want to change, e.g. the hostname to connect to. Currently these parameters are all hardcoded, but it's easy enough to make them into "device parameters" like we have for other device classes, so they can be changed manually from the GUI. Let's do this so that we don't have a bunch of unnecessarily hardcoded parameters.

While I haven't verified that `OPUSInterface` still works with the real hardware, as I don't have it, we have unit tests to ensure that the URLs requested are unchanged.

Fixes #547. Fixes #557.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/config.py
+++ b/finesse/config.py
@@ -85,8 +85,8 @@ TEMPERATURE_MONITOR_TOPIC = "temperature_monitor"
 TEMPERATURE_CONTROLLER_TOPIC = "temperature_controller"
 """The topic name to use for temperature controller-related messages."""
 
-DEFAULT_OPUS_HOST = "10.10.0.2"
-"""The IP address of the machine running the OPUS software."""
+DEFAULT_OPUS_HOST = "localhost"
+"""The IP address or hostname of the machine running the OPUS software."""
 
 DEFAULT_OPUS_PORT = 80
 """The port for OPUS HTTP requests."""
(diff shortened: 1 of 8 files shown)


jamesturner246 (approved):
Looks sensible enough. Will it be a problem if the device manager screen starts getting to cluttered?

dalonsoa (approved):
Looks good to me!

alexdewar:
@jamesturner246: Shouldn't be a problem. The extra GUI components for these devices will only appear when you select them from the dropdown boxes. Plus, while there are a fair few options for devices overall, that's kind of the point of that dialog -- to let users tinker with things. In the ordinary course of events they should just connect to a hardware set from the main window, which will select the appropriate device types and parameters for them.

PR #562, opened 2024-04-17
Fix GUI inconsistencies when opening data file fails

### LINKED ISSUE #561: If data file fails to open, GUI still shows that recording is taking place
The best fix is probably to do the same thing that we do for `Device`s: have one pubsub topic to request that the device/file be opened and a separate one for if/when that request succeeds. Otherwise the only way for the frontend to stay consistent is to check for errors itself and manually undo whatever operations are done when the `data_file.open` message is received.

alexdewar:
# Description

Currently if the user attempts to start recording a data file, but the file creation fails for some reason, the GUI is still updated as if it was successful (e.g. the `RecordingWidget` shows "RECORDING IN PROGRESS").

The problem is caused by the fact that there is only one pubsub message to request that recording starts (`"data_file.open"`), but there is no way for the frontend to know whether or not this was successful (it is implied to have failed if a `"data_file.error"` message is sent, but it would be finicky to make use of this information). A better approach is to do what we do for `Device`s and have a separate message when opening is successful, at which point the GUI can update itself. Implement this with a new `"data_file.opened"` message.

Fixes #561.

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Makes sense and the changes look good to me, specially because you were not using the path provided in many of the callbacks, because that was not necessary.

Issue #457, opened 2023-12-05
Make EM27 sensors code generic


alexdewar:
For the UNIRAS Project, we will be reading sensor values from the aircraft, instead of
the EM27. Given that the interfaces are likely to be similar, both the backend and
frontend code could be made more generic with little effort.

dalonsoa:
This is about interfacing with the Decades software (see email sent in December).

dalonsoa:
@jonemurray , are there any sensor readings to be taken from the spectrometer itself - not the aircraft - like voltages, temperatures of other things like that?

dalonsoa:
It is unclear how to pull these from FTSW500, but they are displayed there. So, let's just use that software and not extracting them into FINESSE.

alexdewar:
Reopening for further discussion. See #527

PR #566, opened 2024-04-19
Misc device fixes and tidy-ups

### LINKED ISSUE #563: Allow for creating classes which inherit indirectly from `Device` but are not registered as device types
Currently we use the following heuristic to determine whether a device plugin class should be registered as a base type or a concrete device type:

 * If it inherits directly from `Device` it is registered as a base type
 * If it has no abstract methods (i.e. it is a concrete class) then it is registered as a device type
 * Else it is assumed to be "intermediate" and is not registered as either

I have a use case where I want to create a class which is neither a base type nor a device type, but has no abstract methods, which is currently not possible. It would be nice to be able to explicitly override the above behaviour when needed.

### LINKED ISSUE #564: `AbstractDevice` should remove from `_device_parameters` any which don't appear in child class's constructor
It is possible that child classes may not have a device parameter which their parent does, in which case that parameter will not be an argument in the child's `__init__`. This currently doesn't work, however, because the code assumes that children always have the same device parameters as their parents (plus, potentially, some more). We should check for this condition and remove the relevant device params when creating child classes.

alexdewar:
# Description

In the course of my failed refactor and while working on #457 I've found various small problems and I thought they probably deserved their own PR, so that the one for #457 doesn't end up too big and convoluted.

Commit summary:

- Fix a couple of small issues with `Device`'s `__init_subclass__` magic (they aren't breaking anything at present, but they are a blocker for the code I've written for #457)
- Various devices' implementations of `close()` don't call `super().close()`, which could break things
- `FTSW500Interface`: Remove unnecessary error handling code for `request_command()` (because we subscribe to messages with `self.subscribe()`, error handling is taken care of)
- Add some missing tests and remove a duplicate

Fixes #563. Fixes #564.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/device.py
+++ b/finesse/hardware/device.py
@@ -87,11 +87,6 @@ class AbstractDevice(ABC):
         """
         super().__init_subclass__()
 
-        # Every time we create a new class, create a new _device_parameters attribute,
-        # so that _add_parameters() and _update_parameter_defaults() don't clobber
-        # values for the parent class.
-        cls._device_parameters = deepcopy(cls._device_parameters)
-
         cls._add_parameters(parameters)
         cls._update_parameter_defaults()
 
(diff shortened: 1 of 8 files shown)


dalonsoa (approved):
LGTM!

alexdewar:
I've replaced the set operations I used when copying device parameters, because I want
to keep the order of the keys in the dict.

Thanks for the review!

Issue #567, opened 2024-04-22
Port DECADES module to use generic sensor interface


alexdewar:
Once the generic sensors interface has been finished (#457), we need to port the DECADES code over to use it. Once this is done, it should fix #513 automatically, as the sensor values will be displayed in the existing panel.

PR #568, opened 2024-04-22
Add `SensorsBase` class for generic interface and port EM27 sensor code to use it

### LINKED ISSUE #457: Make EM27 sensors code generic

alexdewar:
# Description

This PR adds the generic `SensorsBase` class and modifies the frontend and backend EM27 sensor code to use it. I've left off doing the same for the DECADES code for now (#567).

Fixes #457.

## Type of change

## Key checklist

## Further checks


finesse/sensor_reading.py
```diff
@@ -5,16 +5,16 @@
 
 
 @dataclass
-class EM27Property:
+class SensorReading:
     """Class for representing EM27 monitored properties.
```
dc2917, line 9: Something like

```suggestion
    """Class for representing physical quantities monitored by sensors.
```

finesse/sensor_reading.py
```diff
@@ -5,16 +5,16 @@
...
 
     Args:
-        name: name of the physical quantity
+        description: name of the physical quantity
```
dc2917, line 12: I think this might (though only might) be slightly (and only slightly) confusing as the table from which the EM27 properties are read contains a "Name" field and a "Description" field.

finesse/sensor_reading.py
dc2917, line 1: A few EM27-specific things left lying around here

dc2917 (changes requested):
Looks good overall! Functionality is great, I've just made a few comments about some namings and pointed out a few places where EM27-specific things have been left in.

finesse/gui/main_window.py
```diff
@@ -69,7 +69,7 @@ def __init__(self) -> None:
         spectrometer: QGroupBox = SpectrometerControl()
 
         # Setup for interferometer monitor
-        em27_monitor = EM27Monitor()
+        em27_monitor = SensorsPanel()
         em27_monitor.setSizePolicy(
```
dalonsoa, line 73: Shouldn't this be renamed something different than `em27_monitor`?

dalonsoa (approved):
Keeping aside Dan's comments, this looks good to me!

dalonsoa:
But checking links is failing :)

finesse/sensor_reading.py
alexdewar: Hmm... I don't mind whether we go with `name` or `description`. The reason I went with `description` is because elsewhere I've used that to mean a *human-readable* name (though seemingly I forgot to mention this in the docstring :upside_down_face:)

finesse/gui/sensors_panel.py
```diff
@@ -58,17 +58,17 @@ def _create_layouts(self) -> None:
         self._layout.addWidget(top)
         self._layout.addWidget(bottom)
 
-    def _get_prop_lineedit(self, prop: EM27Property) -> QLineEdit:
+    def _get_prop_lineedit(self, prop: SensorReading) -> QLineEdit:
         """Create and populate the widgets for displaying a given property.
 
         Args:
-            prop: the EM27 property to display
+            prop: the property to display
```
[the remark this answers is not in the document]
alexdewar: True!

finesse/gui/main_window.py
alexdewar: Indeed!

finesse/sensor_reading.py
alexdewar: I reverted to calling it `name`, but added a comment saying it should be human readable

finesse/sensor_reading.py
dc2917: Just need to update the module docstring and the docstrings of `__str__` and `val_str`.

dc2917 (approved):
One tiny change to sensor_reading.py and we're good to go!

PR #575, opened 2024-04-29
Fix OPUS request URLs

### LINKED ISSUE #574: OPUS command URLs are broken
Since 72b6e851, OPUS commands have been broken. The problem is that the query part of the URL is being URL encoded itself, so e.g. http://10.10.0.2:80/opusrs/cmd.htm?opusrsstart is becoming http://10.10.0.2:80/opusrs/cmd.htm%3Fopusrsstart.

alexdewar:
# Description

Since commit 72b6e851, the request URLs used by `OPUSInterface` have been broken (#574), which causes commands to fail (e.g. starting recording). Fix by creating the URL from a string rather than a `QUrl`. We had unit tests for this but they didn't catch it.

Fixes #574.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/http_requester.py
+++ b/finesse/hardware/http_requester.py
@@ -4,7 +4,6 @@ from collections.abc import Callable
 from functools import partial
 from typing import Any
 
-from PySide6.QtCore import QUrl
 from PySide6.QtNetwork import QNetworkAccessManager, QNetworkReply, QNetworkRequest
 
 from finesse.config import DEFAULT_HTTP_TIMEOUT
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
These changes look ok, but it would be good to know why it was not working and why the tests were not picking it. In principle, the existing code looked ok as well.

alexdewar:
Sorry, it was a slightly rushed description. The problem was that we were using the `setPath()` method of the `QUrl` to set the path with a query string, e.g. `setPath("cmd.htm?opusrsstart")`, but the question mark gets escaped rather than passed through. You need to use the `setQuery()` method instead to set this part, but I decided it was easier to just go back to using simple strings. The reason the test didn't pick it up was because the test didn't look at the case where there was a query string. Actually, I'll remedy that :smile:

PR #581, opened 2024-04-30
Port DECADES backend to common sensors API

### LINKED ISSUE #567: Port DECADES module to use generic sensor interface

alexdewar:
# Description

This PR ports the `Decades` sensor class to use the common sensors API, by making it a subclass of `SensorsBase`, as we do for other sensor devices. I also made a few changes to the frontend so that the generic `SensorsPanel` displays the DECADES parameters correctly and removed the DECADES-specific panel that was there before.

I also changed the behaviour of `Decades` slightly along the way. Previously, we only requested three specific parameters from the API as a demo. As we don't know which ones will actually be useful to Jon (and, at this stage, probably he doesn't either), I changed it to display all parameters. This does also mean that there is no mechanism for only displaying a subset of the parameters at the moment, but we can add this feature when needed. Additionally, I changed the default polling time from 60s to 5s, because many of the sensor values can change quite rapidly (e.g. aircraft velocity).

I had to modify the `SensorsPanel` code to remove old displayed parameters when connecting to a new device so that the old values don't hang around, which isn't a problem we had before. I also made the sensor readings part scrollable, which it needs to be in order to display all of the various DECADES parameters. I also just tidied it up a bit (e.g. by using `QFormLayout`s instead of manually adding `QLabel`s next to various widgets).

Fixes #567.

## Type of change

## Key checklist

## Further checks


finesse/config.py
```diff
@@ -57,19 +57,9 @@
...
 """The URL of the DECADES sensor data endpoint."""
 
-DECADES_POLL_INTERVAL = 60.0
+DECADES_POLL_INTERVAL = 5.0
 """Poll rate for DECADES sensors."""
 
-DECADES_TOPIC = "decades"
-"""The topic name to use for DECADES sensor-related messages."""
-
-DECADES_QUERY_LIST = [
```
dalonsoa, line 66: Actually, I seem to remember @dc2917 asked Jon about this and he mentioned this specific parameters. Most definitely, he was not interested in all of them (100+ parameters).

dalonsoa (approved):
Keeping aside my comment below, it looks good to me.

finesse/config.py
alexdewar: Ohhh I see. I guess I should have asked :grimacing:.

@jonemurray Is that right? Which parameters do you want to read from the DECADES system?

alexdewar: Hmm... maybe what I'll do then is go back to showing just a subset of the readings, but make it user-configurable so that we don't have to redeploy if @jonemurray wants to change this.

PR #596, opened 2024-05-28
Fix mypy error in test

alexdewar:
It seems the latest version of `mypy` has noticed the trickery I was doing with this method and is complaining about it.

The body of the method is never actually executed -- we just need it because the parent class is abstract -- but let's replace it with something that `mypy` is happier.

### DIFF
--- a/tests/gui/test_path_widget.py
+++ b/tests/gui/test_path_widget.py
@@ -11,7 +11,7 @@ from finesse.gui.path_widget import PathWidget
 class DummyWidget(PathWidget):
     """Override abstract member functions so we can create an instance."""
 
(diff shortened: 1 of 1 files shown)


tests/gui/test_path_widget.py
```diff
@@ -13,7 +13,7 @@ class DummyWidget(PathWidget):
 
     def try_get_path_from_dialog(self) -> Path | None:
```
dc2917, line 14:
```suggestion
    def try_get_path_from_dialog(self) -> None:
```

alexdewar: Good spot!

alexdewar:
Thanks!

PR #604, opened 2024-05-30
Release v2 of FINESSE

### LINKED ISSUE #582: Connection to OPUS sometimes times out
@jonemurray has reported that he's seen connections to the OPUS software timing out when
using FINESSE. It seems that the current default timeout value we're using for HTTP
connections
is 2s:
    https://github.com/ImperialCollegeLondon/FINESSE/blob/2abdeec1c03c4cdb691ed42ad7bcc8e13d1ed209/finesse/config.py#L73

This is clearly not enough, though it is surprising that the HTTP connection is timing
out given that it is to the same machine! It could be that, as with HTTP connections to
the EM27 itself, the connections aren't handled on a separate thread meaning there can
be a long delay for a response (~2s for the EM27).

Either way, we should increase the default timeout and see if that fixes things.

alexdewar:
This PR contains various small tweaks and bumps the version number to 2.0.0. I thought these small changes probably didn't deserve their own PRs.

Commit 4ff93ff is an actual bug fix (fixes #582). Jon has tested it thoroughly last week and confirmed that it works.

Besides that, there are some config file changes and a fix to the readme.


dalonsoa (approved):
LGTM!

PR #630, opened 2024-07-31
Change from `poetry` to `rye`

alexdewar:
This is a proof of concept showing how the project will change if we move from `poetry` to `rye`.


alexdewar:
The tests will all fail :upside_down_face:

PR #643, opened 2024-10-11
Fix some comparisons to appease ruff

alexdewar:
It seems like I just broke the CI by merging some bot PRs. Now things are failing because of a (presumably new) ruff check. Not sure why the CI didn't catch it, but in any case it's easy to fix.


dc2917 (approved):
> Not sure why the CI didn't catch it, but in any case it's easy to fix.

Maybe different versions of ruff being used in CI/pre-commit/local?

alexdewar:
Yeah, that would make sense

PR #653, opened 2024-10-16
Upgrade `matplotlib` and fix `mypy` errors

### LINKED ISSUE #644: `mypy` fails with `matplotlib` v3.9
See broken dependabot PR: #635

The problem seems to be that the deduced types for `matplotlib` functions are no longer compatible with our code in `temperature_plot.py`.

To fix this, I think we should check to see whether we are doing something we shouldn't (e.g. should we be including `None`s in our input arguments?) and try to tidy things on our side. If that fails, we can just suppress `mypy` in the places where it's falling over.

alexdewar:
# Description

This PR upgrades `matplotlib` to 3.9.2 and fixes the somewhat odd type errors that came with it. The code looked fine to me, but apparently we were falling foul of the type checker by freely converting between `np.ArrayLike` objects and `list`s, even though the code itself worked fine. I considered just suppressing the errors here, but it seemed a bit cleaner to just yield to `mypy` and use `np.array`s instead of `list`s.

Fixes #644.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/temperature_plot.py
+++ b/finesse/gui/temperature_plot.py
@@ -6,6 +6,7 @@ from functools import partial
 
 import matplotlib.pyplot as plt
 import matplotlib.ticker as ticker
+import numpy as np
 from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg
 from pubsub import pub
 from PySide6.QtCore import QSize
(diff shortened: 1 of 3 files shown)


alexdewar:
@dc2917 It would be handy to have your review on this as the original author of the matplotlib bits.

dc2917 (approved):
Seems sensible to me, and works as expected

PR #664, opened 2024-10-22
Updates and fixes for tests

### LINKED ISSUE #437: `test_tc4820.py::test_read` fails in VS Code only
It complains that it can't find the test result. I suspect it may be that the output of this test is too long (it's a parameterised test), but in any case I think it would benefit from being split out into two or more separate tests, which will hopefully also fix this problem.

alexdewar:
# Description

The tests currently can't be discovered by VS Code's test adaptor, for a couple of reasons. Firstly, it's not compatible with the `pytest-mypy` extension. Secondly, the ordering of a parameterised test differed between runs and that also seems to break things (I was using a `set` to supply parameters so its order was different on every run). I fixed the first problem by replacing the `pytest-mypy` extension with a CI workflow to run `mypy` directly (as @AdrianDAlessandro has done elsewhere). Moving to a newer version of `mypy` caused it to fail (for the same reason that #658 is failing), so I also had to fix up some type hints. I fixed the second problem by putting the offending parameters in a sorted `list`. Fixing this problem also means we can run our tests in parallel with `pytest-xdist`, so I added that too -- while the tests were finishing in <30s anyway, parallelising it halves the time it takes to run on my machine, which seems like a worthwhile improvement.

Fixes #437.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,13 +22,13 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Run tests
-        run: poetry run pytest --cov-report=xml
-
       - name: Run mypy
-        # run even if tests fail
-        if: always()
         run: poetry run mypy .
+
+      - name: Run tests
+        # run even if mypy fails
+        if: always()
+        run: poetry run pytest --cov-report=xml
 
       - name: Upload coverage to Codecov
         if: runner.os == 'Linux' && success()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Run mypy
         # run even if tests fail
         if: always()
-        run: poetry run mypy finesse
+        run: poetry run mypy .
 
       - name: Upload coverage to Codecov
         if: runner.os == 'Linux' && success()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,11 @@ jobs:
       - name: Run tests
         run: poetry run pytest --cov-report=xml
 
+      - name: Run mypy
+        # run even if tests fail
+        if: always()
+        run: poetry run mypy finesse
+
       - name: Upload coverage to Codecov
         if: runner.os == 'Linux' && success()
         uses: codecov/codecov-action@v4
--- a/tests/hardware/plugins/temperature/test_tc4820.py
+++ b/tests/hardware/plugins/temperature/test_tc4820.py
@@ -66,11 +66,17 @@ def _get_message(len: int) -> tuple[int, bytes, Any]:
     )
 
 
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -25,6 +25,11 @@ jobs:
       - name: Run tests
         run: poetry run pytest --cov-report=xml
 
+      - name: Run mypy
+        # run even if tests fail
+        if: always()
```
dalonsoa, line 30: This is here to ensure that both, tests and mypy are run, regardless of the result of each of them, right?

tests/hardware/plugins/temperature/test_tc4820.py
```diff
@@ -66,11 +66,17 @@ def _get_message(len: int) -> tuple[int, bytes, Any]:
...
+
+    for c in range(0, 128, 10):
+        chars.add(chr(c))
+
+    return sorted(chars)
 
-for c in range(0, 128, 10):
-    _TERM_CHARS.add(chr(c))
+
+_TERM_CHARS = _get_term_chars()
```
dalonsoa, line 78: Not sure why this complicated. Why not just:
```python
_TERM_CHARS = ["^", "*"] + [chr(c) for c in range(0, 128, 10)]
```
And it is not random, but rather deterministic, always the same characters.

dalonsoa (approved):
Just a minor change, but looks good, otherwise.

tests/hardware/plugins/temperature/test_tc4820.py
alexdewar: It's so that we don't end up with duplicates for the first two. It doesn't matter much, but we don't really want to run the exact same test twice.

dalonsoa: But you are. There's no randomness in here. Regardless of using the function or the list, you get exactly the same collection of characters, right?

alexdewar: We won't get duplicates now because I use a set, then convert it to a list.

I'm not sure whether the `range(0, 128, 10)` bit includes `^` and `*`, so I add them separately. I could check, but then if we change the range (e.g. to include more chars) then we might accidentally remove them and we definitely should test those two, because they have a special meaning.

I suppose another way of doing it would be to just include every char between 0 and 127 inclusive (i.e. the whole ASCII range). That's not so many tests, really.

alexdewar: I just realised I describe it as "random ASCII characters" but I meant they were arbitrary, not actually random :stuck_out_tongue:. I should probably change this comment

.github/workflows/ci.yml
alexdewar: Though I've realised we should probably do it the other way round, because now the code coverage won't be uploaded unless `mypy` succeeds

PR #669, opened 2024-10-24
Add versions to hardware sets and measurement scripts

### LINKED ISSUE #443: Add `version` field to hardware set config files
This would be useful if we ever want to make non-backwards-compatible changes to the file format.

### LINKED ISSUE #444: Add `version` field to measure script files
As with #443, it would be useful if measure script files had a version field in case we ever want to make non-backwards-compatible changes to the format.

dalonsoa:
# Description

Adds a version number to the hardware sets and the measure scripts. For now, this is not really used for anything, but set the basis in case we need to make backwards incompatible changes.

As the version field is mandatory, existing scripts and hardware sets will fail to work. If this is not the desired behaviour, we can easily make the field optional and give it a default value of 1. 

Fixes #443 
Fixes #444

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/hardware_set/hardware_set.py
+++ b/finesse/gui/hardware_set/hardware_set.py
@@ -82,6 +82,7 @@ def _non_empty(x: Any) -> bool:
 
 _hw_set_schema = Schema(
     {
+        "version": int,
         "name": str,
         "devices": And(
             _non_empty,
@@ -101,6 +102,7 @@ _hw_set_schema = Schema(
 class HardwareSet:
     """Represents a collection of devices for a particular hardware configuration."""
 
+    version: int
     name: str
     devices: frozenset[OpenDeviceArgs]
     file_path: Path
@@ -128,7 +130,7 @@ class HardwareSet:
         """Save this hardware set as a YAML file."""
         with file_path.open("w") as file:
             devices = dict(map(_device_to_plain_data, self.devices))
-            data = dict(name=self.name, devices=devices)
+            data = dict(version=self.version, name=self.name, devices=devices)
             yaml.dump(data, file, sort_keys=False)
 
     @classmethod
@@ -146,7 +148,9 @@ class HardwareSet:
             OpenDeviceArgs.create(k, **v)
             for k, v in plain_data.get("devices", {}).items()
         )
-        return cls(plain_data["name"], devices, file_path, built_in)
+        return cls(
+            plain_data["version"], plain_data["name"], devices, file_path, built_in
+        )
 
 
 def _get_new_hardware_set_path(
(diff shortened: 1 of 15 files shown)


dalonsoa:
@alexdewar , this is a bit odd. The CI stops just after the mypy step, but no error is reported and all looks good! Do you have any idea of what's going on?

image

finesse/gui/hardware_set/hardware_set.py
```diff
@@ -82,6 +82,7 @@ def _non_empty(x: Any) -> bool:
 
 _hw_set_schema = Schema(
     {
+        "version": int,
```
alexdewar, line 85: Let's enforce the schema version. Users may attempt to load a file made with a newer version of FINESSE, with unexpected results.

I *think* you can just write this:

```suggestion
        "version": 1,
```

(Though it may be better to use a constant for the magic number.)

See: https://github.com/keleshev/schema?tab=readme-ov-file

finesse/gui/measure_script/script.py
```diff
@@ -106,6 +108,7 @@ def parse_script(script: str | TextIOBase) -> dict[str, Any]:
 
     schema = Schema(
         {
+            "version": int,
```
alexdewar, line 111: See comment above.

alexdewar (changes requested):
Thanks for this!

I think that we should be enforcing the schema version though: we don't want users to attempt to load files in a *newer* format, as well as an older one. See my comments.

The other thing is that I don't think we need to made the file format version an attribute of the `HardwareSet` or `Script` classes. The file format will just be checked once when a file is loaded, then it won't be used after that, so it can be discarded.

The CI runners are failing because the `pytest` jobs are timing out, which is really odd. I was able to replicate it locally: it seems like random tests are just hanging indefinitely. It only occurs with commit 9697aab9ae980da99fc6c36f8b143f1d129f55c1 and not earlier. I was concerned that it was some race condition caused by me parallelising the tests, but it still happens if you run `pytest -n 1` and tests run fine on `main`. Optimistically though, if you make the changes I suggest then maybe this problem will go away by itself! If not, we can revisit.

dalonsoa:
I thought that the point of the version was to do things one way or another depending on the version number - at least, raise a deprecation warning or something along those lines. If everything boils down to having a different schema incompatible with the previous ones, validating the configuration will fail, anyway, version or not version.

alexdewar:
@dalonsoa That's true, but I still think it's better to be explicit. (I suppose we could make a change that affects the way we use a field which doesn't actually change the schema too.) We *might* want to support old file formats in future too though and having an explicit `version` property makes this a bit cleaner. 

In any case, we probably wouldn't want a `HardwareSet` object to behave differently depending on the file format... right? The file format should only affect the file loading stage, so we don't need this information afterwards.

dalonsoa:
OK, so I'm going to close this PR and create another one just with a change in the Schema - enforcing it to be 1, and modifying the yaml files, so those work. Some tests might need updating, but they will be way less.

alexdewar:
It also means we can be clearer with error messages, e.g. "The file you are trying to load was made with a newer version of FINESSE" rather than just "config file malformed" or whatever

alexdewar:
> OK, so I'm going to close this PR and create another one just with a change in the Schema - enforcing it to be 1, and modifying the yaml files, so those work. Some tests might need updating, but they will be way less.

Cool, thanks :+1:

PR #670, opened 2024-10-25
Add functional test that all hardware set files are valid

alexdewar:
While we test the `_load_builtin_hardware_sets()` does what it's supposed to by using mocking, we don't test that the actual YAML files bundled with FINESSE are valid, which is a potential error source (e.g. if a device class name changes and a config file isn't updated, then we won't notice until a user tries to actually use the file).

Let's just ensure that all the built-in hardware set config files load properly.

I'm not sure why we don't do this already tbh.

@dalonsoa I thought this might be useful for your PR #669.


dalonsoa (approved):
LGTM! Such a simple test, and such a big reassurance of the validity of some parts of the software!

Issue #638, opened 2024-10-11
Allow for opening devices asynchronously


alexdewar:
This is a watered down version of #324.

Some devices take a long time to open. In particular, the `ST10Controller` moves the mirror to the home position as part of its initialisation routine, which can take several seconds. It's stupid to have the main (GUI) thread blocking for this long, so let's refactor things so that this happens on a background thread. This will mean that devices can have a new possible state -- "connecting" -- which the software will have to deal with, but this shouldn't be too difficult.

PR #678, opened 2024-10-30
Reorganise hwset code

alexdewar:
This PR purely involves moving some code around. I started doing it as part of my work on #638, but figured it probably makes sense to do this as a separate PR so all this churn doesn't appear in the diff for the functional changes I want to make.

Currently we have a `finesse.gui.hardware_set.hardware_set` module and I thought it just made more sense to move everything to the parent module (i.e. `finesse.gui.hardware_set`), so that's what I've done. There's also a module `finesse.gui.hardware_set.device_connection` with just a couple of functions in it, so I've moved these to the parent module too.

No functional changes intended.


dalonsoa (commented):
I'm not sure I like this. I'm not against having code in `__init__.py` files, but I feel this is a lot of functionality that its better encapsulated in its own module and imported only when needed in those places that need them. What is to gain with this change?

alexdewar:
There's not a lot that we gain really tbh... I started out by moving a couple of enums out of `hardware_set.py` into `__init__.py` then possibly got a little carried away after that :stuck_out_tongue:. You're probably right that it makes more sense to keep most of this in its own submodule.

alexdewar:
Maybe I'll just shelve this. I'm not sure it's worth it after all.

PR #680, opened 2024-10-31
Display mirror angle

### LINKED ISSUE #583: Display current mirror angle (esp. when running script)
@jonemurray has requested that we indicate the current mirror angle somewhere on the
GUI. In particular, it seems that this feature would be useful when measure scripts are
running.

dc2917:
# Description

This PR adds the ability to display the current mirror angle on the GUI, as determined by the stepper motor controller.

The GUI should display the angle moved to (broadcast by the device once it has finished moving), or the name of the corresponding preset.

We'd like to also have the GUI indicate that the mirror is moving, however I don't think this is possible to test with the dummy device.

Fixes #583 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -40,7 +40,7 @@ class StepperMotorControl(DevicePanel):
 
         # Create widgets to show the current mirror position
         layout.addWidget(QLabel("Current position"), 0, 4)
-        self.mirror_position_display = QLabel("")
+        self.mirror_position_display = QLabel()
         self.mirror_position_display.setAlignment(Qt.AlignmentFlag.AlignCenter)
         layout.addWidget(self.mirror_position_display, 1, 4)
 
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -1,7 +1,8 @@
 """Code for controlling the stepper motor which moves the mirror."""
 
 from pubsub import pub
-from PySide6.QtWidgets import QButtonGroup, QGridLayout, QPushButton, QSpinBox
+from PySide6.QtCore import Qt
+from PySide6.QtWidgets import QButtonGroup, QGridLayout, QLabel, QPushButton, QSpinBox
 
 from finesse.config import ANGLE_PRESETS, STEPPER_MOTOR_TOPIC
 from finesse.gui.device_panel import DevicePanel
@@ -36,6 +37,12 @@ class StepperMotorControl(DevicePanel):
 
         layout.addWidget(self.angle, 1, 2)
         layout.addWidget(self.goto, 1, 3)
+
+        # Create widgets to show the current mirror position
+        layout.addWidget(QLabel("Current position"), 0, 4)
+        self.mirror_position_display = QLabel("")
+        self.mirror_position_display.setAlignment(Qt.AlignmentFlag.AlignCenter)
+        layout.addWidget(self.mirror_position_display, 1, 4)
 
         self.setLayout(layout)
 
--- a/finesse/gui/stepper_motor_view.py
+++ b/finesse/gui/stepper_motor_view.py
@@ -47,6 +47,10 @@ class StepperMotorControl(DevicePanel):
         self.setLayout(layout)
 
         pub.subscribe(
+            self._indicate_moving,
+            f"device.{STEPPER_MOTOR_TOPIC}.move.begin",
+        )
+        pub.subscribe(
             self._update_mirror_position_display,
             f"device.{STEPPER_MOTOR_TOPIC}.move.end",
         )
@@ -69,17 +73,18 @@ class StepperMotorControl(DevicePanel):
         target = float(self.angle.value()) if btn is self.goto else btn.text().lower()
         pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.move.begin", target=target)
 
-    def _update_mirror_position_display(self, moved_to: float | None) -> None:
+    def _indicate_moving(self, target) -> None:
+        """Update the display the indicate that the mirror is moving."""
+        self.mirror_position_display.setText("Moving...")
+
+    def _update_mirror_position_display(self, moved_to: float) -> None:
         """Display the angle the mirror has moved to.
 
-        moved_to will be None if mirror is moving, in which case display "Moving...",
-        otherwise, display angle or its associated name.
+        If angle corresponds to a preset, show the associated name, otherwise just show
+        the value.
         """
-        if moved_to is None:
-            self.mirror_position_display.setText("Moving...")
+        preset = next((k for k, v in ANGLE_PRESETS.items() if v == moved_to), None)
+        if preset:
+            self.mirror_position_display.setText(preset.upper())
         else:
-            preset = next((k for k, v in ANGLE_PRESETS.items() if v == moved_to), None)
-            if preset:
-                self.mirror_position_display.setText(preset.upper())
-            else:
-                self.mirror_position_display.setText(f"{moved_to}°")
+            self.mirror_position_display.setText(f"{moved_to}°")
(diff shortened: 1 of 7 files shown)


finesse/gui/stepper_motor_view.py
```diff
@@ -37,8 +38,19 @@ def __init__(self) -> None:
         layout.addWidget(self.angle, 1, 2)
         layout.addWidget(self.goto, 1, 3)
 
+        # Create widgets to show the current mirror position
+        layout.addWidget(QLabel("Current position"), 0, 4)
+        self.mirror_position_display = QLabel("")
```
alexdewar, line 43:
```suggestion
        self.mirror_position_display = QLabel()
```

finesse/gui/stepper_motor_view.py
```diff
@@ -51,7 +63,25 @@ def _add_checkable_button(self, name: str) -> QPushButton:
...
+        if moved_to is None:
+            self.mirror_position_display.setText("Moving...")
+        else:
+            if moved_to in ANGLE_PRESETS.values():
+                preset = list(ANGLE_PRESETS.keys())[
+                    list(ANGLE_PRESETS.values()).index(moved_to)
+                ]
+                self.mirror_position_display.setText(preset.upper())
+            else:
+                self.mirror_position_display.setText(f"{moved_to}°")
```
alexdewar, line 87: I'd be inclined to use a generator expression here, e.g.:

```suggestion
            preset = next((k for k, v in ANGLE_PRESETS.items() if v == moved_to), None)
            if preset:
                self.mirror_position_display.setText(preset.upper())
            else:
                self.mirror_position_display.setText(f"{moved_to}°")
```

finesse/gui/stepper_motor_view.py
```diff
@@ -51,7 +63,25 @@ def _add_checkable_button(self, name: str) -> QPushButton:
     def _preset_clicked(self, btn: QPushButton) -> None:
         """Move the stepper motor to preset position."""
         # If the motor is already moving, stop it now
+        pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.notify_on_stopped")
         pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.stop")
 
         target = float(self.angle.value()) if btn is self.goto else btn.text().lower()
         pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.move.begin", target=target)
```
alexdewar, line 70: The problem here is that the `move.end` message will be sent in response to the `stop` message, not the `move.begin` one below, which means it'll send the *last* angle as an argument. So:

```suggestion
        pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.stop")

        pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.notify_on_stopped")
        target = float(self.angle.value()) if btn is self.goto else btn.text().lower()
        pub.sendMessage(f"device.{STEPPER_MOTOR_TOPIC}.move.begin", target=target)
```

finesse/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -219,7 +219,7 @@ def close(self) -> None:
 
     @Slot()
     def _send_move_end_message(self) -> None:
-        self.send_message("move.end")
+        self.send_message("move.end", moved_to=self.angle)
```
alexdewar, line 222: How about adding a `send_move_end_message` method to the `StepperMotorBase` class? I know it's only a one-liner, but it seems a bit tidier. Then we could get rid of `ST10Controller`'s `_send_move_end_message` method altogether too.

alexdewar (changes requested):
I've made some small suggestions. Besides that there are a couple of bugs:

1. The display shows the previous angle, not the current one (see comment)
2. The display will never show a "moving" state, because `_update_mirror_position_display` is only ever called when the motor has finished moving. To figure out when it has started moving you need to listen for the `move.begin` message instead. My PR might help with working on this: #681

finesse/gui/stepper_motor_view.py
dc2917: Thanks! I knew there had to be a more elegant way haha

finesse/hardware/plugins/stepper_motor/st10_controller.py
dc2917: Is [this](2ef80f771b94889527647829c85ddbe0d1c5f9c4) what you had in mind? Wasn't sure what to do with the `Slot` decorator, but if so I'll try and get a bit more test coverage and then re-request review.

alexdewar: Yes, exactly! I'm still not entirely clear on what `Slot`s are for, but it seems like you can omit the decorator without problems, so I think it's fine.

alexdewar:
FYI: I'm just merging #694 and I've updated the required status checks for the Python 3.12 changes, so this branch will need updating before the checks pass now. You should just be able to do a simple merge/rebase.

finesse/gui/stepper_motor_view.py
```diff
@@ -53,5 +65,21 @@ def _preset_clicked(self, btn: QPushButton) -> None:
...
+
+    def _update_mirror_position_display(self, moved_to: float | None) -> None:
+        """Display the angle the mirror has moved to.
+
+        moved_to will be None if mirror is moving, in which case display "Moving...",
+        otherwise, display angle or its associated name.
+        """
+        if moved_to is None:
+            self.mirror_position_display.setText("Moving...")
```
alexdewar, line 79: This is still something that needs tweaking if we want it to display "moving" when the mirror is moving. You need to have a separate handler which sets the text to moving when you get a `move.begin` message.

finesse/gui/stepper_motor_view.py
```diff
@@ -53,5 +65,21 @@ def _preset_clicked(self, btn: QPushButton) -> None:
...
+        else:
+            preset = next((k for k, v in ANGLE_PRESETS.items() if v == moved_to), None)
+            if preset:
+                self.mirror_position_display.setText(preset.upper())
+            else:
+                self.mirror_position_display.setText(f"{moved_to}°")
```
alexdewar, line 85: On reflection, I'm thinking it might be useful to see the angle for presets too. How about:

```suggestion
            text = f"{moved_to}°"
            if preset := next((k for k, v in ANGLE_PRESETS.items() if v == moved_to), None):
                text += f" ({preset})"
            self.mirror_position_display.setText(text)
```

alexdewar (changes requested):
Almost there I think! But it still won't display "moving" when the mirror is moving. See comment.

finesse/gui/stepper_motor_view.py
dc2917: This does display the names as, e.g. "cold_bb", rather than the text on the labels ("COLD BB"), but it's still obvious which each corresponds to haha

tests/gui/test_stepper_motor_view.py
```diff
@@ -62,3 +68,22 @@ def test_goto_clicked(sendmsg_mock: MagicMock, qtbot: QtBot) -> None:
...
+
+def test_update_mirror_position_display(qtbot: QtBot) -> None:
+    """Test the mirror position display updates correctly."""
+    control = StepperMotorControl()
+
+    control._update_mirror_position_display(moved_to=ANGLE_PRESETS["zenith"])
+    assert control.mirror_position_display.text() == "180.0\u00b0 (zenith)"
+
+    control._update_mirror_position_display(moved_to=12.34)
+    assert control.mirror_position_display.text() == "12.34" + "\u00b0"
```
alexdewar, line 89:
```suggestion
    assert control.mirror_position_display.text() == "12.34\u00b0"
```

alexdewar (approved):
LGTM! Just need to resolve that merge conflict then this is ready to go :smile:

PR #681, opened 2024-11-01
Add dummy stepper motor delay in default configuration

alexdewar:
# Description

It turns out that while the `DummyStepperMotor` has the facility to add a delay between starting and stopping a movement, we're not actually using this facility anywhere, for some reason, which is a bit of an oversight. The parameter wasn't changeable via the frontend either. Let's fix this, as it's useful for testing things like @dc2917's work on #680.

### DIFF
--- a/finesse/gui/hardware_set/dummy_ftsw500.yaml
+++ b/finesse/gui/hardware_set/dummy_ftsw500.yaml
@@ -3,6 +3,8 @@ name: "Dummy devices + FTSW500 spectrometer"
 devices:
   stepper_motor:
     class_name: stepper_motor.dummy.DummyStepperMotor
+    params:
+      move_duration: 1.5
   temperature_controller.hot_bb:
     class_name: temperature.dummy_temperature_controller.DummyTemperatureController
   temperature_controller.cold_bb:
--- a/finesse/hardware/plugins/stepper_motor/dummy.py
+++ b/finesse/hardware/plugins/stepper_motor/dummy.py
@@ -7,7 +7,14 @@ from PySide6.QtCore import QTimer
 from finesse.hardware.plugins.stepper_motor.stepper_motor_base import StepperMotorBase
 
 
-class DummyStepperMotor(StepperMotorBase, description="Dummy stepper motor"):
+class DummyStepperMotor(
+    StepperMotorBase,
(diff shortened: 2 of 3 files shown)


dc2917 (approved):
Ahh, this makes way more sense. I did see the dummy stepper motor `device` has a timer that makes use of `move_duration`, and the init docstring says "This class uses a simple timer to notify when the move is complete after a fixed amount of time." - I just assumed it was faster than I could click 😅

finesse/hardware/plugins/stepper_motor/dummy.py
```diff
@@ -26,6 +33,8 @@ def __init__(
         """
         if steps_per_rotation < 1:
             raise ValueError("steps_per_rotation must be at least one")
+        if move_duration < 0.0:
+            raise ValueError("move_duration cannot be negative")
```
dalonsoa, line 37: I'd restrict to a positive number, as it should not be possible to have this duration as zero.
```suggestion
        if move_duration <= 0.0:
            raise ValueError("move_duration needs to be strictly positive")
```

finesse/gui/hardware_set/dummy_ftsw500.yaml
```diff
@@ -3,6 +3,8 @@ name: "Dummy devices + FTSW500 spectrometer"
 devices:
   stepper_motor:
     class_name: stepper_motor.dummy.DummyStepperMotor
+    params:
```
dalonsoa, line 6: Do we need steps per rotation, here?

dalonsoa (approved):
LGTM!

finesse/hardware/plugins/stepper_motor/dummy.py
alexdewar: It is a *bit* weird, but it does sort of make sense to have it as an option for e.g. unit tests. We don't want to be waiting unnecessarily there. Instantaneous movement shouldn't actually break anything.

finesse/gui/hardware_set/dummy_ftsw500.yaml
alexdewar: If we miss it out, the default will be used instead, which should be fine.

Issue #668, opened 2024-10-24
`ST10Controller` hangs GUI while moving mirror to start position


alexdewar:
I was surprised to find there wasn't an issue for this already as it's been on my to-do list for ages now!

Currently the `ST10Controller` code hangs the GUI thread (meaning the window gets marked as "not responding") while moving to its initial position, which takes about 5s or so, which is a bit horrible. We should delay signalling that the device is open until the motor has homed.

Depends on #638.

Issue #415, opened 2023-11-14
Check EM27Sensors is working after opening


alexdewar:
Currently, if the server is not available then the device still appears to open successfully, before the HTTP request times out. It would be better to wait until at least one response has been received before designating the device as open.

This change should probably be done as part of #324, as that will also involve running devices' constructors on separate threads.

alexdewar:
Relates to #638.

Issue #247, opened 2023-06-27
Connecting to wrong serial port causes GUI to hang


alexdewar:

alexdewar:
Could be fixed by #638

alexdewar:
Can't reproduce. It was probably fixed by #638.

PR #682, opened 2024-11-01
Allow device plugins to offer asynchronous opening

### LINKED ISSUE #638: Allow for opening devices asynchronously

alexdewar:
# Description

I think I've done enough tinkering with this now and should probably stop :stuck_out_tongue:.

# Background

Currently, the process of opening devices (which can include making a USB connection, moving a motor, etc.) all takes place on the main thread. The advantage of this model is that it's simple to work with -- the frontend doesn't have to worry about any intermediate stage where a device is still initialising, but can't actually be used for anything. The disadvantage is that it means that either we have to "finish" the device opening process early (e.g. before we've received a response from a server) or block the main (GUI) thread until it completes, both of which have drawbacks (e.g. #668, #247, #415).

# Implementation

The backend changes were not especially complicated. The frontend is informed that we've started connecting to a device with a new `device.before_opening` message. I've added another class parameter, `async_open`, which device classes can use to indicate that they open asynchronously. It defaults to `False`, so classes have to opt in to this new behaviour. (Currently I haven't implemented this for any actual devices, though I have tested it.) Device classes with `async_open==True` just need to explicitly call `signal_is_opened` when they've finished opening (in practice, this will be from another thread).

Making the frontend work with these changes was a bit more fiddly. You should be able to disconnect from devices that are still connecting (though this might require some small changes to actual device code). I've added a new widget to show the connection status of individual devices to the device management dialog. Everything else should (hopefully) just work as before.

I'm afraid this PR is a little on the long side (sorry!) but a lot of the changes are to the tests, so I hope it's not too painful to review.

Closes #638.


finesse/hardware/device.py
```diff
@@ -76,19 +76,26 @@ class AbstractDevice(ABC):
...
     The key represents the parameter name and the value is a list of possible values.
     """
+    _device_async_open: ClassVar[bool | None] = None
+    """Whether the device opens asynchronously (i.e. completes after __init__)."""
 
     def __init_subclass__(
-        cls, parameters: Mapping[str, str | tuple[str, Sequence]] = {}
+        cls,
+        parameters: Mapping[str, str | tuple[str, Sequence]] = {},
+        async_open: bool | None = None,
```
dalonsoa, line 85: What's the reason of being `None`? Should be either async or not?

dalonsoa (approved):
I haven't check the tests, but the actual code looks ok to me.

It is a bit difficult to follow who does what as there are a lot of classes knot together to handle the status of the devices, and the connection of the devices, etc. so I think that at some point you will need to write a thorough developers guide explaining all of these relationships and why they are the way they are. 

I don't see where a new threads are created to open a device. Could you point me in the right place. Also, my only initial concern was updating the GUI in the main thread from messages sent from another thread, but I guess this is not a problem if you have tested it and it works.

finesse/hardware/device.py
alexdewar: If it's `None`, then it just leaves the `_device_async_open` property as whatever it was before, so if it was set by a parent class, it won't be overridden.

alexdewar:
> It is a bit difficult to follow who does what as there are a lot of classes knot together to handle the status of the devices, and the connection of the devices, etc. so I think that at some point you will need to write a thorough developers guide explaining all of these relationships and why they are the way they are.

Yeah, it's a bit of a tangle :disappointed:. If I have time, I'll add a bit more documentation before merging this PR. Otherwise I'll do it later.

> I don't see where a new threads are created to open a device. Could you point me in the right place. Also, my only initial concern was updating the GUI in the main thread from messages sent from another thread, but I guess this is not a problem if you have tested it and it works.

This part of the code doesn't actually create any threads. (I did consider an implementation where we left things the same as before, but just ran the `__init__` methods on separate threads, but decided against it in the end.) Any class with `async_open=True` is just responsible for calling `signal_is_opened()` at some later stage, which in practice will probably involve other threads. This doesn't mean they have to explicitly make Python threads though: e.g. all the devices which send HTTP requests use Qt for this, which makes its own background threads (and sends a signal when you get a response).

That's a good point about updating the GUI from a different thread. It's not a problem for the devices I've been working on, because they will use `QThread`s + Qt signals to pass messages back to the main thread, which then invokes `signal_is_opened()`, but it could be a problem for other devices. I'll make sure to document it properly.

dc2917 (approved):
Code looks good, and the functionality sensibly implemented to me. Let's hope it works with the real devices!

alexdewar:
Thanks for reviews :smile:

PR #694, opened 2024-11-05
Update required Python version to 3.12

### LINKED ISSUE #584: Port to Python 3.12
I don't think there is any particular reason why this should be difficult, but there may be some small issues that need sorting out.

We should update the versions in:

- [ ] Minimum version specified `pyproject.toml` (then run `poetry lock --no-update`)
- [ ] The version used for `ruff`'s `pyupgrade` checks (also in `pyproject.toml`)
- [ ] The versions in CI workflows

We should merge #642 and #635 first.

alexdewar:
Update pyproject.toml and regenerate poetry.lock file. Also update CI workflows to use Python 3.12.

I meant to test this in the lab, but didn't have enough time in the end. I'd be *really* surprised if it broke anything though (the tests pass and the GUI works fine with dummy devices), so I think we should just do it. I'll test it when I next go into the lab (hopefully next week) and we can always revert.

Closes #584.


dalonsoa (commented):
There're are some things failing with links - presumably because of the removal of Codacy - and the tests are not running. It looks good, but I'll wait to approve until that is sorted.

dalonsoa (approved):
Actually, tests are passing, but those are not the ones that are required. The settings need updating. As it turns out, I do not have Admin access. Have you, @alexdewar ?

PR #695, opened 2024-11-07
Updates and fixes for the stepper motor controller

### LINKED ISSUE #116: Pytest displays (harmless) error about `MockSerialReader`
Sometimes when I run `pytest` I get harmless error messages at the end of the output that look like this:

```
Exception ignored in: <function DeviceBase.__del__ at 0x7fb5dc296550>
  File "/home/alex/code/FINESSE/finesse/hardware/device_base.py", line 10, in __del__
    self.close()
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/st10_controller.py", line 226, in close
    self._reader.quit()
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/st10_controller.py", line 80, in quit
    super().quit()
RuntimeError: Internal C++ object (MockSerialReader) already deleted.
Exception ignored in: <function _SerialReader.__del__ at 0x7fb5dc1e6ee0>
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/st10_controller.py", line 74, in __del__
    self.quit()
  File "/home/alex/code/FINESSE/finesse/hardware/stepper_motor/st10_controller.py", line 80, in quit
    super().quit()
RuntimeError: Internal C++ object (MockSerialReader) already deleted.
```

We're probably doing something racy in the destructor. It's harmless, so doesn't need an urgent fix, but it does sometimes obscure `pytest`'s output which is annoying.

### LINKED ISSUE #668: `ST10Controller` hangs GUI while moving mirror to start position

alexdewar:
# Description

The primary motivation for this PR was to add async device opening for the `ST10Controller` device so it doesn't hang the GUI (#668), which has been bothering me for a while. The implementation for this is fairly simple: it's just a case of waiting till the motor has stopped its initial move before invoking `signal_is_opened()`. The only detail is that the motor may become stuck before this happens, so I've added a timer to check that it reaches the home position within 10s. By doing things in this way, it means we can also get rid of some other functionality, which is a bonus.

I also found some other things while in the lab the other day (with @jonemurray's help), which make the homing logic I copied over from the old code make a bit more sense. So I've tweaked some comments and fixed up the `_get_input_status` function. This should help for when we need to rejig the code for the new UNIRAS rig, which will likely be set up differently.

I also discovered the source of some annoying log spam messages (#116) and fixed that. I added tests for all the parts of the `ST10Controller` code which are currently untested.

All in all this was quite satisfying :smile:.

Closes #668. Closes #116.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/plugins/stepper_motor/st10_controller.py
+++ b/finesse/hardware/plugins/stepper_motor/st10_controller.py
@@ -133,17 +133,10 @@ class _SerialReader(QThread):
         while self._process_read():
             pass
 
-    def read_sync(self, timeout: float | None = None) -> str:
-        """Read synchronously from the serial device.
-
-        Args:
-            timeout: Amount of time to wait for a response (None==default)
-        """
-        if timeout is None:
-            timeout = self.sync_timeout
-
+    def read_sync(self) -> str:
+        """Read synchronously from the serial device."""
         try:
-            response = self.out_queue.get(timeout=timeout)
+            response = self.out_queue.get(timeout=self.sync_timeout)
         except Exception:
             raise SerialTimeoutException()
 
@@ -390,18 +383,14 @@ class ST10Controller(
         # "Send string"
         self._write_check(f"SS{string}")
 
-    def _read_sync(self, timeout: float | None = None) -> str:
+    def _read_sync(self) -> str:
         """Read the next message from the device synchronously.
-
-        Args:
-            timeout: Amount of time to wait for a response (None==default)
 
         Raises:
             SerialException: Error communicating with device
-            SerialTimeoutException: Timed out waiting for response from device
             ST10ControllerError: Malformed message received from device
         """
-        return self._reader.read_sync(timeout)
+        return self._reader.read_sync()
 
     def _write(self, message: str) -> None:
         """Send the specified message to the device.
--- a/finesse/hardware/plugins/stepper_motor/st10_controller.py
+++ b/finesse/hardware/plugins/stepper_motor/st10_controller.py
@@ -37,8 +37,8 @@ class _SerialReader(QThread):
     The one exception is the send string command ("SS"), which, in addition to eliciting
     an immediate ack (or nack) response, also leads to another response being sent when
     the motor has finished moving. Note that while the motor is moving, other commands
-    can be sent and even processed, e.g. you can request the motor's current position
-    while it is moving.
+    can be sent and even processed, e.g. you can request that another move happens after
+    the first is complete.
 
     All of the reading we do from the device goes via this class, as conceivably the
     magic send string response could be sent at any point after the send string command
--- a/finesse/hardware/plugins/stepper_motor/st10_controller.py
+++ b/finesse/hardware/plugins/stepper_motor/st10_controller.py
@@ -10,7 +10,7 @@ The specification is available online:
 import logging
 from queue import Queue
 
-from PySide6.QtCore import QThread, Signal, Slot
+from PySide6.QtCore import QThread, QTimer, Signal, Slot
 from serial import Serial, SerialException, SerialTimeoutException
 
 from finesse.hardware.plugins.stepper_motor.stepper_motor_base import StepperMotorBase
@@ -153,7 +153,9 @@ class _SerialReader(QThread):
         return response
 
 
-class ST10Controller(SerialDevice, StepperMotorBase, description="ST10 controller"):
+class ST10Controller(
+    SerialDevice, StepperMotorBase, description="ST10 controller", async_open=True
+):
     """An interface for the ST10-Q-NN stepper motor controller.
 
     This class allows for moving the mirror to arbitrary positions and retrieving its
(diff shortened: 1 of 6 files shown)


finesse/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -182,10 +172,20 @@ def __init__(self, port: str, baudrate: int = 9600, timeout: float = 5.0) -> Non
...
         self._reader.read_error.connect(self.send_error_message)
         self._reader.start()
 
+        self._init_error_timer = QTimer()
+        """A timer to raise an error if the motor takes too long to move."""
+        self._init_error_timer.setInterval(10000)  # 10 seconds
+        self._init_error_timer.setSingleShot(True)
+        self._init_error_timer.timeout.connect(
+            lambda: self.send_error_message(
+                RuntimeError("Timed out waiting for motor to move")
```
dalonsoa, line 185: Ok, so this is raised and then... what? How it is the current (stuck) connection handled?

finesse/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -182,10 +172,20 @@ def __init__(self, port: str, baudrate: int = 9600, timeout: float = 5.0) -> Non
...
 
         self._reader = _SerialReader(self.serial, timeout)
-        self._reader.async_read_completed.connect(self._send_move_end_message)
+        self._reader.async_read_completed.connect(self._on_initial_move_end)
         self._reader.read_error.connect(self.send_error_message)
         self._reader.start()
 
+        self._init_error_timer = QTimer()
+        """A timer to raise an error if the motor takes too long to move."""
+        self._init_error_timer.setInterval(10000)  # 10 seconds
```
dalonsoa, line 181: Why 10s? Is it a sensible amount of time to wait?

finesse/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -263,10 +278,8 @@ def _home_and_reset(self) -> None:
         # In case the motor is still moving, stop it now
         self.stop_moving()
 
-        # If the third (boolean) value of the input status array is set, then move the
-        # motor first. I don't know what the input status actually means, but this is
-        # how it was done in the old program, so I'm copying it here.
-        if self._get_input_status(3):
+        # If the homing switch is already active, move the motor a bit
+        if self._get_input_status(6):
```
dalonsoa, line 282: Why do we want to do this?

dalonsoa (approved):
The changes look sensible and the code is seem cleaner. However, I'm not sure about in what state is the whole thing if the motor gets stuck. What happens next? Can the user retry again to move things? Does they need to switch on and off the machine? I think it is good that this issue is identified and handled better than before, it is not entirely clear to me how to recover from this error.

finesse/hardware/plugins/stepper_motor/st10_controller.py
alexdewar: It is handled the same way all device errors are atm: the user gets an error message in a dialog and the device is closed. I've never seen this happen, but I think it makes sense to check it rather than wait indefinitely.

alexdewar: Note that this is just for the initial move, when we're homing the motor.

You can actually ask the motor controller if a drive error has occurred, which would be a useful thing to do the rest of the time when we're trying to move the motor, but that's a separate job.

finesse/hardware/plugins/stepper_motor/st10_controller.py
alexdewar: It could be made a bit shorter, but if it takes longer than 10s then something's definitely gone wrong. (This is the total time taken for the motor to home itself, which involves moving all the way to the optoswitch so it knows which way it's oriented, then rotating so it's facing downwards, which we take as 0°.)

@dc2917 has made the sensible suggestion to make timeout configurable though, so I'll do that.

finesse/hardware/plugins/stepper_motor/st10_controller.py
alexdewar: I suppose the logic is that if the motor is already at the homing switch, it's better to move it away a bit so it can home again to get a more accurate value for where the edge of the switch is (I don't know over how many degrees the switch is active). This is the way things were done in the old software.

alexdewar:
@dc2917 I've just realised this will create merge conflicts with your branch because I've added some tests for the ST10. Sorry!

jonemurray:
Hi All, I have been keeping track of the comments regarding this. Firstly, we can never be too sure how the environment might impact the motor once we're flying and at altitude. These motors are quite robust and are not driving anyway near their capacity but a failure may occur. As it's very difficult to access the instrument in flight having the controller relay a motor fault, if it can be done is helpful. This might help us differentiate between a timeout due to a motor failure and an opto-switch failure, although I'd hope implementing a limit switch (real or virtual) should help. 

I'm OK with a configurable timeout for homing, but we should be sure that a user is aware this can be set and will cause issues if set too low.

The move of the homing switch (if the system is initiated with the blade covering the homing switch) is crucial, as Alex notes we're looking to set our reference position to the edge of the blade from a set direction. The rise/fall time of the switch is 15 micro seconds, at the angular speed the motor turns we'll have good repoducibility of position. At 360 degrees/s the rise time equates to 0.005 degrees.

PR #696, opened 2024-11-07
Tests for the sensors panel

### LINKED ISSUE #106: Add tests for `SensorsPanel`
We need to test that the GUI correctly displays the data returned from the web server in a robust manner, i.e. in case of a given sensor absence, failure or dynamic changes to the order of sensors listed.

dc2917:
# Description

This PR adds units tests for the functionality of the sensors panel on the GUI.

Fixes #106 

## Type of change

## Key checklist

## Further checks


alexdewar:
Hmm... `mypy` seems to be complaining. I removed the `pytest-mypy` extension, so you'll have to run it manually over your code to find the source of the error.

dc2917:
> Hmm... `mypy` seems to be complaining. I removed the `pytest-mypy` extension, so you'll have to run it manually over your code to find the source of the error.

Yeah, had this on the mirror angle PR as well, for the exact same thing. It's related to the `QWidget` inheritance - it seems I need to do something like `assert isinstance(`widget_to_investigate`, `ConcreteWidget`), and then can query attributes of the concrete widget type.

In terms of running mypy, I have it running as a pre-commit hook but perhaps it's ignoring the tests/ directory. Pretty sure I ran `pre-commit run --all-files` before pushing 🤷.

tests/gui/test_sensors_panel.py
```diff
@@ -0,0 +1,80 @@
...
+from finesse.sensor_reading import SensorReading
+
+
+@patch("finesse.gui.sensors_panel.pub.subscribe")
+def test_init(subscribe_mock, qtbot) -> None:
+    """Test SensorsPanel's constructor."""
+    panel = SensorsPanel()
+    assert panel.title() == "Sensor readings"
+
+    # Check the panel's widgets are created correctly
```
dc2917, line 18: I wondered if it was worth making any assertions about the other widgets in the panel (`_reading_layout` etc), but given that they're all basically empty I wasn't sure how much value it would add.

alexdewar: I agree. It would be a pain to check everything, so I think it's fine to just check the main moving parts (e.g. labels that change etc.).

tests/gui/test_sensors_panel.py
```diff
@@ -0,0 +1,80 @@
...
+
+    readings = (
+        SensorReading("Quantity1", 1.0, "quantity 1 units"),
+        SensorReading("Quantity2", -8.1, "quantity 2 units"),
+    )
+
+    # Check LED flashes
+    with patch.object(panel, "_poll_light") as poll_light_mock:
+        poll_light_mock.flash = Mock()
+
```
alexdewar, line 73: I don't think this line is necessary. If you make something a mock, then all its attributes (and attributes' attributes etc.) are also mocks.

```suggestion
```

alexdewar (approved):
Looking good. Thanks for sorting this!

There's one line I think you can drop. but it's not esp important.

PR #720, opened 2024-11-27
Add async open support for various network-based devices

### LINKED ISSUE #415: Check EM27Sensors is working after opening

### LINKED ISSUE #666: Wait for response from DECADES server before signalling device is open
Similar to #415.

### LINKED ISSUE #667: Wait for response from OPUS before signalling device is open

alexdewar:
# Description

This PR changes the various network-based devices to use the `async_open` feature. While I haven't been able to test this on the real hardware, unfortunately, I have checked that HTTP requests are sent after device opening (of course, they time out). I'm pretty confident it will work with the real hardware as the changes are pretty minimal, though you can never be sure. Jon has volunteered to test it while he's away, so I think we should just merge it; if it breaks things we can always revert and we don't want to be hanging around waiting to merge this.

Closes #415. Closes #666. Closes #667.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/plugins/spectrometer/opus_interface.py
+++ b/finesse/hardware/plugins/spectrometer/opus_interface.py
@@ -96,7 +96,7 @@ class OPUSInterface(
         self._url = f"http://{host}:{port}/opusrs"
         """URL to make requests."""
 
-        self._status = SpectrometerStatus.UNDEFINED
+        self._status: SpectrometerStatus | None = None
         """The last known status of the spectrometer."""
         self._status_timer = QTimer()
         self._status_timer.timeout.connect(self._request_status)
@@ -124,7 +124,7 @@ class OPUSInterface(
         # If the status has changed, notify listeners
         if new_status != self._status:
             # On first update, we need to signal that the device is now open
-            if self._status == SpectrometerStatus.UNDEFINED:
+            if self._status is None:
                 self.signal_is_opened()
 
             self._status = new_status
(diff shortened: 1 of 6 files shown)


dc2917 (approved):
Seems sensible to me!

finesse/hardware/plugins/spectrometer/opus_interface.py
```diff
@@ -122,6 +123,10 @@ def _on_reply_received(self, reply: QNetworkReply) -> None:
 
         # If the status has changed, notify listeners
         if new_status != self._status:
+            # On first update, we need to signal that the device is now open
+            if self._status == SpectrometerStatus.UNDEFINED:
```
dalonsoa, line 127: Isn't it a bit odd that when it's undefined, it is set as open?

dalonsoa (approved):
The changes look good to me. Of course, only testing with the real hardware will reveal any issues.

finesse/hardware/plugins/spectrometer/opus_interface.py
alexdewar: Sorry, to clarify -- I set it to UNDEFINED in `__init__` before we actually have a status, so it's a sentinel value. But it's probably not a good idea, for the reason I mentioned

PR #721, opened 2024-11-27
:white_check_mark: Add test for roundtrip saving of scripts

### LINKED ISSUE #700: Add test that written measure scripts are readable
Issue #699 highlighted that we don't currently have a test which checks that saved measure script files are actually valid. We should add a test for this, to avoid future bugs.

dalonsoa:
# Description

Simply checks that a measure script saved with `dlg._try_save()` can be loaded without errors by `Script.try_load()`. It does not check if the loaded contents are the same, just if the saved script is a valid script to re-load.

NOTE: On MacOS, there's a bunch of errors everywhere saying: `ValueError: Port /dev/cu.BLTH does not end with a number`. Not a big deal because the chances of UNIRAS/FINESSE to be used in MacOS are slim, but just mentioning it as it makes development under MacOS a bit painful as it is unclear whats a genuine error and what is not. 

Fixes #700

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,12 +6,12 @@ from unittest.mock import MagicMock
 import pytest
 from pubsub import pub
 
-from finesse.hardware.serial_device import SerialDevice
-
 
 @pytest.fixture
 def serial_mock(monkeypatch) -> MagicMock:
     """Fixture for Serial's constructor."""
+    from finesse.hardware.serial_device import SerialDevice
+
     mock = MagicMock()
 
     def new_init(self, *args, **kwargs):
--- a/tests/gui/measure_script/test_script_edit_dialog.py
+++ b/tests/gui/measure_script/test_script_edit_dialog.py
@@ -98,7 +98,7 @@ def test_try_save(
 
 
 def test_save_load_roundtrip(dlg: ScriptEditDialog, tmp_path: Path) -> None:
-    """Test the _try_save() method."""
+    """Test the _try_save() and try_load() methods in tandem."""
     path = tmp_path / "test_script.yaml"
     with patch.object(dlg.script_path, "try_get_path") as path_mock:
         path_mock.return_value = path
--- a/tests/gui/measure_script/test_script_edit_dialog.py
+++ b/tests/gui/measure_script/test_script_edit_dialog.py
@@ -97,6 +97,18 @@ def test_try_save(
                     errmsg_mock.assert_called_once()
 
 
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -6,12 +6,12 @@
 import pytest
 from pubsub import pub
 
-from finesse.hardware.serial_device import SerialDevice
-
 
 @pytest.fixture
 def serial_mock(monkeypatch) -> MagicMock:
     """Fixture for Serial's constructor."""
+    from finesse.hardware.serial_device import SerialDevice
```
dalonsoa, line 13: This outside of the function were preventing the tests to be loaded because of the error mentioned in the PR description. Now they are loaded fine, just most of those related to hardware fail in MacOS.

tests/gui/measure_script/test_script_edit_dialog.py
```diff
@@ -97,6 +97,18 @@ def test_try_save(
                     errmsg_mock.assert_called_once()
 
 
+def test_save_load_roundtrip(dlg: ScriptEditDialog, tmp_path: Path) -> None:
+    """Test the _try_save() method."""
```
dalonsoa, line 101:
```suggestion
    """Test the _try_save() and try_load() methods in tandem."""
```

tests/gui/measure_script/test_script_edit_dialog.py
```diff
@@ -97,6 +97,18 @@ def test_try_save(
                     errmsg_mock.assert_called_once()
 
 
+def test_save_load_roundtrip(dlg: ScriptEditDialog, tmp_path: Path) -> None:
```
alexdewar, line 100: It doesn't look like this is causing issues currently, but I think there should be `qtbot` in the arguments as you are making a `QWidget`:

```suggestion
def test_save_load_roundtrip(dlg: ScriptEditDialog, tmp_path: Path, qtbot: QtBot) -> None:
```

(The qtbot just sets up a fake Qt application context for the test. Without it you sometimes get segfaults if you try to use Qt functionality.)

alexdewar (approved):
I've made one small suggestion, but otherwise LGTM! Thanks for this.

I think the issue you're having on your Mac counts as a bug... I'll open an issue for it. We don't want to make development on Macs unnecessarily painful (it's possible it could affect a Linux system too, I think).

alexdewar:
Is this ready to merge @dalonsoa?

PR #722, opened 2024-11-28
Fix sorting of port names without terminating numbers

alexdewar:
# Description

As reported by @dalonsoa, not all USB device port names end in a number on MacOS, in contrast to Windows and Linux (at least as far as I've seen).

This breaks our current code for sorting the ports, which will throw an error if a port name doesn't end in a number. Fix by making the number optional and also including the rest of the string in the sorting operation.

While we won't actually be deploying on MacOS, the fix is easy and we want to make life easy for developers.

@dalonsoa Would you mind verifying that this fixes things for you? If you're happy with this, feel free to just merge it.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/serial_device.py
+++ b/finesse/hardware/serial_device.py
@@ -28,13 +28,20 @@ def _port_info_to_str(vendor_id: int, product_id: int, count: int = 0) -> str:
     return out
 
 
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
This works and can run all tests without any error 👍

PR #731, opened 2024-12-09
Reduce precision of temperatures written to CSV file

alexdewar:
Currently temperatures will be written with the maximum precision possible for the `Decimal` type, which equates to a lot of decimal places. This wasn't such an issue when using the DP9800 as it uses a protocol with plain text numbers to a limited number of decimal places, but now that we are using the Seneca device with a binary encoding this can end up being a lot longer. Let's just cap the precision to 0.01 Kelvin.

This was requested by @jonemurray and given that it is only a small change I thought it would be easier to just do it now.


dc2917 (approved):

dalonsoa (approved):
LGTM!

PR #742, opened 2024-12-11
Allow for reading measure scripts generated with FINESSE v2.0.0

### LINKED ISSUE #741: Measure scripts generated with v2 of FINESSE not compatible anymore
This was introduced (on purpose!) when we added a version field to the measure script files. If you try running measure scripts generated with v2 then you'll get an error.

I was hoping that no one would be making measure scripts until we made a new release, but it seems that Jon is now testing things in earnest with v2 of the software, so I'd rather not subject him to this breakage.

It's an easy fix: we just need to make the version field optional for now. If we ever make a version 2 of the measure script file format, we can make it mandatory then.

alexdewar:
# Description

I was hoping that by adding a version field to measure scripts and hardware sets for this next release we could ensure that that field is always present, but it seems like Jon has been doing quite a lot of testing with v2.0.0 (quite reasonably!) so he already has some measure scripts that don't have this field. I don't want him to have to put up with error messages when he upgrades, so let's just assume that measure scripts without a version are version 1. At least going forward we will have this field.

(NB: This isn't a problem for hardware sets, because those are currently all bundled with FINESSE, though we will have user-generated ones in future: #393.)

Fixes #741.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/gui/measure_script/script.py
+++ b/finesse/gui/measure_script/script.py
@@ -127,7 +127,14 @@ def parse_script(script: str | TextIOBase) -> dict[str, Any]:
     )
 
     try:
-        output = schema.validate(yaml.safe_load(script))
+        output = yaml.safe_load(script)
+
+        # v2.0.0 and older didn't have a version field, but the file formats are
+        # otherwise identical
+        if "version" not in output:
+            output["version"] = 1
+
+        output = schema.validate(output)
         output.pop("version")
         return output
     except (yaml.YAMLError, SchemaError) as e:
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
Looks good but why not making the schema assign automatically a default version value when validating? In other words, in [here](https://github.com/ImperialCollegeLondon/FINESSE/blob/4016f562ef2608ab609930746af19490180c9c68/finesse/gui/measure_script/script.py#L110) we use something like:

```python
    schema = Schema(
        {
            Optional("version", default=1): int,
            "repeats": measurements_type,
            ...
         }
    )

alexdewar:
@dalonsoa I was kind of in two minds about whether to change the actual schema or not. In the end I decided that if we ever make a v2 of the file format it'll probably be useful to be able to check the version up front to know which schema to use, so it's probably best to make sure it's always there.

PR #743, opened 2024-12-11
Get mirror immediate position

### LINKED ISSUE #691: Allow for getting current position when stepper motor is moving
Currently we're using the `SP` command, which retrieves the set position. However, it won't work if the motor is moving, so we return `None` in this case. There's also a similar artificial limitation in the `DummyStepperMotor` device for testing purposes.

It would be useful to know the motor's position even while it's moving, though: if we could record this information in the CSV file then that's potentially extra data and it would be useful to be able to see the position in the GUI too (#583).

According to the manual, there is another command -- `IP` -- which retrieves the immediate (estimated) position, with this caveat: "This is the calculated trajectory position, which is not always equal to actual
position."

It would be worth playing with this to see if we can just replace `SP` with `IP`. Alternatively, we could just use `IP` only when the motor is moving.

dc2917:
# Description

This PR enables getting the immediate position of stepper motors while they are moving.
Previously, querying the position during motion would return `None`. Now, for the ST10, an estimated position is returned.

I've had a go at making the dummy stepper motor return a position based on its `QTimer` interval and remaining time, but it's not fully complete, so we can either address that separately or I can finish it up here.

Fixes #691 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/hardware/plugins/stepper_motor/test_st10_controller.py
+++ b/tests/hardware/plugins/stepper_motor/test_st10_controller.py
@@ -361,26 +361,13 @@ def test_is_moving(
         [(4, "IP=hello", pytest.raises(ST10ControllerError))],
     ),
 )
-@patch(
-    "finesse.hardware.plugins.stepper_motor.st10_controller.ST10Controller.is_moving",
-    new_callable=PropertyMock,
-)
 def test_get_step(
-    is_moving_mock: PropertyMock,
     step: int,
     response: str,
     raises: Any,
     dev: ST10Controller,
 ) -> None:
     """Test getting the step property."""
-    # When the motor is stationary:
-    is_moving_mock.return_value = False
-    with read_mock(dev, response):
-        with raises:
-            assert dev.step == step
-
-    # When the motor is moving:
-    is_moving_mock.return_value = True
     with read_mock(dev, response):
         with raises:
             assert dev.step == step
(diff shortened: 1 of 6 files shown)


tests/hardware/plugins/stepper_motor/test_st10_controller.py
```diff
@@ -357,39 +357,33 @@ def test_is_moving(
...
-    "finesse.hardware.plugins.stepper_motor.st10_controller.ST10Controller.is_moving",
-    new_callable=PropertyMock,
-)
-def test_get_step_moving(
-    is_moving_mock: PropertyMock,
-    dev: ST10Controller,
-) -> None:
-    """Test getting the step property when the motor is moving."""
+    # When the motor is moving:
     is_moving_mock.return_value = True
```
alexdewar, line 383: We don't actually need to mock `ST10Controller.is_moving` anymore because it's not called by `ST10Controller.step`

alexdewar (changes requested):
I've made one suggestion about the test, but it otherwise looks good. Thanks for doing this!

I think it's fine to leave `DummyStepperMotor` as it is for now -- let's not make things hard for ourselves.

I've just realised that we also need to update `data_file_writer.py`, because we record whether or not the motor is moving as a column in the CSV file and it currently does this by checking whether the returned angle is `None`. Would you mind fixing this up too?

dc2917:
> I've just realised that we also need to update data_file_writer.py, because we record whether or not the motor is moving as a column in the CSV file and it currently does this by checking whether the returned angle is None. Would you mind fixing this up too?

I've removed that column from the CSV file entirely. I presume that's what we want, since the only way we check for whether the mirror is moving is if it returns `None`, which will never happen.

alexdewar (changes requested):
The stepper motor classes do have an `is_moving` property that you can use instead. Unfortunately we do need this column -- sorry I wasn't clear :disappointed:

alexdewar (approved):
LGTM! Thanks for fixing up that extra file :smile:

PR #750, opened 2024-12-18
Fix: `DummyStepperMotor` returns `None` if moving

alexdewar:
# Description

As with the `ST10Controller` before #743, the `DummyStepperMotor` returns an angle of `None` if it is in motion. However, this isn't what we want anymore; if you try to record data while using the `DummyStepperMotor` you now get an error when the code tries to use this `None` value. Change it to return the old angle instead.

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
LGTM!

PR #751, opened 2024-12-18
Update to Python v3.13

alexdewar:
# Description

There seems like no better time to do this. I've done some manual testing with Python 3.13 and everything still seems to work as expected (except for the bug fixed in #750).

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
If it works, all good! I guess not tested in a real system, right? I'm just a bit worried because of the `serial` library...

alexdewar:
Nope, just with dummy devices :disappointed:.

I'd be happier if it were tested on the real hardware too, but as we won't have access to it for a while, I think it's probably just ok to merge it and fix things later if needed. Jon has offered to do some testing in the meantime -- I'll make an rc release to send him.

I have checked and no one's complaining about PySerial not working on Python 3.13, so that's something.

PR #766, opened 2025-01-14
Add tests for DP9800 temperature monitor

### LINKED ISSUE #128: Add tests for communications with DP9800 (temperature monitoring device)

dc2917:
# Description

This PR adds unit tests for the DP9800 temperature monitor. 

I also found I had misinterpreted the format of the DP9800 data, mixing up field widths due to `SP` and %-formatting, and so have fixed that here too. (Note that the only case in which that would have caused an error using the actual hardware is if it was reading temperatures with a five-digit mantissa, i.e. temperatures above 10000 C, so no cause for concern, but still embarrassing 🫣).

Fixes #128 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/finesse/hardware/plugins/temperature/dp9800.py
+++ b/finesse/hardware/plugins/temperature/dp9800.py
@@ -70,8 +70,8 @@ def parse_data(data: bytes) -> tuple[list[Decimal], str]:
     except UnicodeDecodeError as e:
         raise DP9800Error(e)
 
-    vals_begin = 3  # Following STX T SP
-    vals_end = 74  # Following STX T (SP %4.2f)*9, assuming 9 vals
+    vals_begin = 2  # Following STX T
+    vals_end = 74  # Following STX T (%8.2f)*9, assuming 9 vals
     etx_index = data.find(b"\x03")
 
     vals = [Decimal(val) for val in data_ascii[vals_begin:vals_end].split()]
@@ -137,14 +137,13 @@ class DP9800(SerialDevice, TemperatureMonitorBase, description="DP9800"):
         The DP9800 returns a sequence of bytes containing the
         temperatures measured on each channel, in the format
 
-          STX T SP t1 SP t2 SP t3 SP t4 SP t5 SP t6 SP t7 SP t8 SP t9 ff ETX BCC NUL
+          STX T t1 t2 t3 t4 t5 t6 t7 t8 t9 ff ETX BCC NUL
 
         where
-            t1, t2, ..., t9: temperature values in the format %4.2f
+            t1, t2, ..., t9: temperature values in the format %8.2f
             STX: Start of Text (ASCII 2)
             ETX: End of Text (ASCII 3)
             NUL: Null character (ASCII 0)
-            SP: Space (ASCII 32)
             BCC: Block Check Character
             ff: System flag in hexadecimal
 
--- a/tests/hardware/plugins/temperature/test_dp9800.py
+++ b/tests/hardware/plugins/temperature/test_dp9800.py
@@ -146,5 +146,5 @@ def test_dp9800_get_temperatures(
     read_mock.assert_called_once()
 
     temperatures = dev.get_temperatures()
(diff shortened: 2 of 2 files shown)


finesse/hardware/plugins/temperature/dp9800.py
```diff
@@ -70,8 +70,8 @@ def parse_data(data: bytes) -> tuple[list[Decimal], str]:
     except UnicodeDecodeError as e:
         raise DP9800Error(e)
 
-    vals_begin = 3  # Following STX T SP
-    vals_end = 74  # Following STX T (SP %4.2f)*9, assuming 9 vals
+    vals_begin = 2  # Following STX T
+    vals_end = 74  # Following STX T (%8.2f)*9, assuming 9 vals
```
alexdewar, line 74: So is there no space after the `STX` and `T` after all then? Surely if we were one char out the numbers we'd be seeing would be all wrong though?

tests/hardware/plugins/temperature/test_dp9800.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Stick in a non-ASCII char and check for error
+    data_string = data.decode("ascii")[:3] + "\x80" + data.decode("ascii")[4:]
+    with pytest.raises(DP9800Error):
+        parse_data(bytes(data_string, "utf-8"))
+
+    temperatures_2_to_9, ff = parse_data(data)
+    expected_temperatures = [
+        Decimal(val) for val in [3.4, 5.6, 7.8, 9.0, 2.3, 4.5, 6.7, 8.9]
+    ]
```
alexdewar, line 81:
```suggestion
    expected_temperatures = map(Decimal, [3.4, 5.6, 7.8, 9.0, 2.3, 4.5, 6.7, 8.9])
```

tests/hardware/plugins/temperature/test_dp9800.py
```diff
@@ -0,0 +1,152 @@
...
+    temperature_reader_mock: MagicMock, read_mock: MagicMock, dev: DP9800, data: bytes
+) -> None:
+    """Test DP9800's get_temperatures method."""
+    temperature_reader_mock.return_value = data
+
+    dev.get_temperatures()
+    read_mock.assert_called_once()
+
+    temperatures = dev.get_temperatures()
+    expected = [Decimal(val) for val in [3.4, 5.6, 7.8, 9.0, 2.3, 4.5, 6.7, 8.9]]
```
alexdewar, line 151:
```suggestion
    expected = map(Decimal, [3.4, 5.6, 7.8, 9.0, 2.3, 4.5, 6.7, 8.9])
```

alexdewar (approved):
This looks great! Good work.

I've got a question about the change in `vals_begin`, but other than that I think we're good to merge.

finesse/hardware/plugins/temperature/dp9800.py
dc2917: It's not that there isn't a space, it's that for temperatures lower than 10000, the first character will be a space because of the fixed width of each field.

dc2917: Sorry, let me rephrase that: there isn't an explicit space character. It's only if the temperature is less than 10000, in which case the fixed width of the temperature field will insert spaces.

alexdewar: Ohhhhh I see. That makes sense!

alexdewar:
Oh, my suggestions broke things :upside_down_face:. Feel free to just revert them and merge @dc2917

finesse/hardware/plugins/temperature/dp9800.py
dc2917: Yeah, sorry, basically I think I just took %4.2f to mean "4 characters before the decimal point, 2 after it", whereas it should be %8.2f which gives a total width of 8 characters, with 5 before the decimal point. So I guess I stuck in the `SP` to make the 5th character.

PR #767, opened 2025-01-17
Rename project to FROG

### LINKED ISSUE #352: Choose a new name for the project and replace uses of FINESSE everywhere.
As we are now entering a new project (UNIRAS), it probably makes sense to rename this project from FINESSE, which was the name of the previous project. When starting UNIRAS we considered producing two separate pieces of software for different hardware used by the FINESSE and UNIRAS rigs (albeit with shared code between the two). As we have now decided that the best way forward is to produce a single piece of software which can be configured at runtime to use whatever hardware is needed from the set provided by the FINESSE and UNIRAS rigs, it makes sense to have a name for this software that is independent of both of the projects. As this is open source, the hope would be that in future other features can be added for other use cases (esp. in terms of other hardware, but possibly in terms of visualisation capabilities etc.) both within Imperial and also possibly outside, which is another reason to give it a broader name.

The options are:

- Leave the name as FINESSE, which is the least effort, but doesn't advertise well that the software also works with other hardware and is being developed with UNIRAS funding
- Rename it to UNIRAS. It would be a good way of trumpeting that the software is an output of this new project
- Rename it to something else.

alexdewar:
# Description

We've finally got a new name for the project: FROG (Far infraRed Observation GUI). See discussion here: #352 

So I've renamed everything in the code base accordingly. I've already renamed the repo, so the new URLs should work (the old ones should automatically redirect).

There were a number of problems with the existing FINESSE name:

1. It was the same name used by the old C# software
2. It's also the name of the [FINESSE rig](https://www.imperial.ac.uk/space-and-atmospheric-physics/research/missions-and-projects/atmospheric-missions/finesse/), whereas the software is designed to handle other rigs (i.e. UNIRAS) so it was confusing and not great from a branding perspective

Now we have a separate name specifically for the software that can hopefully last the test of time (and possibly be used for other projects?).

Note that there are a couple of places in the code base where the term FINESSE is used to refer to the *rig* not the software; these have been left as is:

```
frog/gui/hardware_set/__init__.py:3:  This allows the user to choose between, e.g., the FINESSE and UNIRAS rigs.
frog/gui/hardware_set/finesse.yaml:2:  name: FINESSE
```

Closes #352.

## Type of change

## Key checklist

## Further checks


README.md
```diff
@@ -1,24 +1,25 @@
...
+FROG (**F**ar infra**R**ed **O**bservation **G**UI) is software for controlling a
+spectrometer system developed by Imperial College London's [Space and Atmospheric
+Physics group], written in Python and with a graphical interface using the [PySide6 Qt
+bindings].
 
 Emissivity of the Earth's different surface types helps determine the efficiency with
 which the planet radiatively cools to space and is a critical variable in climate
 models. However, to date, most measurements of surface emissivity have been made in the
-mid-infrared. The FINESSE project is novel in employing a ground-based system capable of
+mid-infrared. The FROG project is novel in employing a ground-based system capable of
```
dc2917, line 19: I think we should still refer to the project (as in the physics research project) as FINESSE?

docs/hardware_sets.md
```diff
@@ -1,13 +1,13 @@
 # Hardware sets
 
 Hardware sets are collections of hardware device types along with their associated
-parameters. They are used in FINESSE to represent a particular hardware rig (e.g.
-FINESSE or UNIRAS), so that users can easily swap between them. FINESSE comes with some
-built in configurations, but users can also define their own, either for convenience
-during testing or to build on FINESSE's functionality.
+parameters. They are used in FROG to represent a particular hardware rig (e.g. FROG or
```
dc2917, line 4:
```suggestion
parameters. They are used in FROG to represent a particular hardware rig (e.g. FINESSE or
```

README.md
alexdewar: D'oh! Good spot.

We don't actually have a link to the FINESSE project either and should probably do that.

docs/hardware_sets.md
alexdewar: Another good spot. Clearly I wasn't being as careful as I thought I was :stuck_out_tongue:

dc2917 (approved):

PR #816, opened 2025-04-02
Check stepper alarm status after move

### LINKED ISSUE #703: Check for drive faults
One way to fix #692 is to query the controller's alarm status after every move to check whether the move terminated prematurely due to the motor hitting a limit switch. While we're querying the alarm status, we will also obtain information about whether the motor has experienced a drive fault. This seems like useful information to share with the user and happily, we'll be getting it anyway, so it makes sense to do something with it. Probably this should just consist of broadcasting a device error and closing the device.

alexdewar:
# Description

The ST10 controller has an "alarm status" that can be queried to find out if an error has occurred. This includes drive faults, low voltage etc. as well as if the motor has hit a limit switch (which was the original motivation for looking at this -- currently we just hardcode the position of the limits in FROG so we don't hit them). Jon has said that he might not put limit switches on the new UNIRAS rig, in which case we could just turn them off, but in any case it seems useful to know whether something's gone wrong with the motor, especially if you're 10,000 ft up.

When an error occurs, we should get a notification because the motor will stop so it makes sense to query the motor at this point.

Closes #703.

## Type of change

## Key checklist

## Further checks

- [xgg] Tests have been added or an issue has been opened to tackle that in the future. (Indicate issue here: # (issue))

### DIFF
--- a/frog/hardware/plugins/stepper_motor/st10_controller.py
+++ b/frog/hardware/plugins/stepper_motor/st10_controller.py
@@ -254,10 +254,20 @@ class ST10Controller(
 
         # For future move end messages, use a different handler
         self._reader.async_read_completed.disconnect(self._on_initial_move_end)
-        self._reader.async_read_completed.connect(self.send_move_end_message)
+        self._reader.async_read_completed.connect(self._on_move_end)
 
         # Signal that this device is ready to be used
         self.signal_is_opened()
+
+    def _on_move_end(self) -> None:
+        """Signal whether move was successful or an error occurred."""
+        if alarm_code := self.alarm_code:
+            # A controller error occurred
+            self.send_error_message(ST10ControllerError(str(alarm_code)))
+            return
+
+        # Move was successful
+        self.send_move_end_message()
 
     def _check_device_id(self) -> None:
         """Check that the ID is the correct one for an ST10.
--- a/frog/hardware/plugins/stepper_motor/st10_controller.py
+++ b/frog/hardware/plugins/stepper_motor/st10_controller.py
@@ -526,6 +526,9 @@ class ST10Controller(
             SerialTimeoutException: Timed out waiting for response from device
             ST10ControllerError: Malformed message received from device
             UnicodeEncodeError: Message to be sent is malformed
+
+        Returns:
+            int: The integer corresponding to the named value
         """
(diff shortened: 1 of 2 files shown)


frog/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -225,11 +254,21 @@ def _on_initial_move_end(self) -> None:
...
 
+    def _on_move_end(self) -> None:
+        """Signal whether move was successful or an error occurred."""
+        if alarm_code := self.alarm_code:
+            # A controller error occurred
+            self.send_error_message(ST10ControllerError(str(alarm_code)))
+            return
+
+        # Move was successful
+        self.send_move_end_message()
```
dc2917, line 270: Just wondering whether an if-else would be more logical, e.g.

```py
if alarm_code := self.alarm_code:
    # A controller error occurred
    ...
else:
    # Move was successful
    ...
```

tests/hardware/plugins/stepper_motor/test_st10_controller.py
dc2917, line 1: No test for `ST10Controller._request_int`?

frog/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -474,6 +514,25 @@ def _request_value(self, name: str) -> str:
...
+        Args:
+            name: Variable name
+            base: Base of integer (e.g. 16 for hexadecimal)
+
+        Raises:
+            SerialException: Error communicating with device
+            SerialTimeoutException: Timed out waiting for response from device
+            ST10ControllerError: Malformed message received from device
+            UnicodeEncodeError: Message to be sent is malformed
+        """
```
dc2917, line 529:
```suggestion

        Returns:
            int: The integer corresponding to the named value
        """
```

dc2917 (approved):
Seems sensible, and looks good overall! Just made a couple of suggestions, and wondering whether we're missing some tests

tests/hardware/plugins/stepper_motor/test_st10_controller.py
alexdewar: I've added one now as well as one for `_on_move_end`

PR #818, opened 2025-04-07
`ST10Controller`: Disable limit switches

### LINKED ISSUE #817: Disable limit switches for `ST10Controller`
@jonemurray has reported that we don't actually need them and that he may not even install them in the new UNIRAS rig. As the code for handling the limit switches could be fiddly, we can just make our lives easier by disabling the limit switches at the outset; they don't actually perform a safety function anyway.

According to the manual, we can use the `DL` command for this.

alexdewar:
# Description

The limit switches aren't actually needed to protect anything on either the FINESSE or UNIRAS rigs. The positions of the limit switches are currently hardcoded which isn't ideal. Handling them properly (i.e. distinguishing error states caused by hitting the limit switches vs other errors) will be annoying. So let's just disable them.

Unrelated change: I noticed the mirror angle is sometimes displayed to a silly number of DPs in the GUI, so I've rounded it to the nearest degree.

Closes #817.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/frog/hardware/plugins/stepper_motor/st10_controller.py
+++ b/frog/hardware/plugins/stepper_motor/st10_controller.py
@@ -219,6 +219,8 @@ class ST10Controller(
         # Check that we are connecting to an ST10
         self._check_device_id()
 
+        self._disable_limit_switches()
+
         # Move mirror to home position
         self._home_and_reset()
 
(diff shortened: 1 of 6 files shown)


frog/hardware/plugins/stepper_motor/st10_controller.py
```diff
@@ -281,6 +283,18 @@ def _check_device_id(self) -> None:
...
+    def _disable_limit_switches(self) -> None:
+        """Disable the limit switches on the controller.
+
+        Without this, we get an error if the motor hits a limit switch. The limit
+        switches aren't actually needed to protect anything in either the FINESSE or
+        UNIRAS rigs (the worst that will happen is the motor will just spin around), so
+        we don't need them.
+        """
+        # "Define limit". This command allows for setting the behaviour of the limit
+        # switches. The "3" option means that they are treated as normal inputs.
```
dc2917, line 295: > The "3" option means that they are treated as normal inputs.

I can't quite work out what this means. Why does this effectively disable the limits?

dc2917: Ah, OK, fair enough!

dc2917 (approved):

Issue #815, opened 2025-04-01
`OPUSInterface` gives device error if user calibrates EM27 while FROG is connected


alexdewar:
When `OPUSInterface` is connected it polls the OPUS software in the background to check the device status. Seemingly, if the user begins a calibration routine using OPUS, the software fails to process any incoming requests for around 15s, which causes our requests to time out and the device to disconnect. This isn't the whole duration of the calibration, just the part where the software displays "starting calibration".

This is a problem as Jon wants to be able to pause measure scripts in the middle and perform calibrations. If the device disconnects, the measure script will be stopped.

I'm labelling this as a bug, even though the bug is really on Bruker's side (their handling of HTTP requests in general is horrible -- the response time is inexplicably several seconds, even when the connection is to localhost).

We can work around it either by increasing the timeout (currently we use 10s) or retrying on failure. Something related that has occurred to me is that we might want to check whether OPUS actually processes multiple requests at once, or if we can only send one every ~2s. If it is de facto rate limited, then we might want to tweak things to make sure status requests don't end up taking priority over requests to do something (e.g. start/stop a measurement).

PR #825, opened 2025-04-08
Refactor generic HTTP code

alexdewar:
# Description

I started off looking at the HTTP code for #815 and figured I'd probably need to make some changes to it (to allow for automatically retrying on failure) but thought it would be easier if I refactored things a bit first.

Currently we have an `HTTPRequester` class which is used to make requests by a few other device classes (`Decades`, `EM27Sensors` and `OPUSInterface`). None of the classes expose a device parameter to allow users to control the timeout for HTTP connections, but I think this is a useful value to be able to change, so I think this is worth adding. The way I've done it is to make a new `HTTPDevice` class, analagous to the `SerialDevice` class which provides this parameter.

I also made a couple of tweaks to `Device`/`AbstractDevice` to fix some small issues that were causing problems.

In retrospect, I'm not sure all this was entirely necessary, but I think the resulting code is a bit cleaner.


frog/hardware/http_device.py
```diff
@@ -0,0 +1,66 @@
...
+):
+    """A device which communicates via HTTP requests."""
+
+    def __init__(self, timeout: float = DEFAULT_HTTP_TIMEOUT) -> None:
+        """Create a new HTTPDevice.
+
+        Args:
+            timeout: The maximum time in seconds to wait for a response from the server
+        """
+        if timeout <= 0:
```
dc2917, line 28: It doesn't look like we enforced this before. I guess we don't want to allow for no timeout? Likely to only cause problems I suppose

dc2917 (approved):
Nice! I think this provides a much tidier interface.

frog/hardware/http_device.py
alexdewar: Yeah, I'm not sure if there'd be a legitimate reason for that. Users can always set a really long timeout if they want

PR #828, opened 2025-04-24
Fix: ST10Controller: Don't handle move end events when closing

alexdewar:
# Description

If you close FROG while the stepper motor is still moving, you can end up with `ST10Controller._on_move_end` being invoked after the serial device is closed, which will trigger an uncaught exception. Fix by disconnecting the slots from `_reader.async_read_completed` on close.

I think this was caused by my last fix to the `ST10Controller` code. It's like playing whack-a-mole :laughing: 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/frog/hardware/plugins/stepper_motor/st10_controller.py
+++ b/frog/hardware/plugins/stepper_motor/st10_controller.py
@@ -236,6 +236,10 @@ class ST10Controller(
         if not self.serial.is_open:
             return
 
+        # Don't handle move end events because they will likely happen after the device
+        # is disconnected
+        self._reader.async_read_completed.disconnect()
+
         try:
             self.stop_moving()
             self.move_to("nadir")


dc2917:
This seems strange. 

> you can end up with ST10Controller._on_move_end being invoked after the serial device is closed, which will trigger an uncaught exception. 

How does this happen if FROG has already been closed?

alexdewar:
It can happen while FROG is still in the process of shutting down. We close all devices explicitly when FROG is shut down so it happens cleanly (e.g. for the stepper motor we also want to point the mirror downwards when we disconnect).

The problem is that we now get a notification whenever the motor stops moving without having to also call `notify_on_send`, including for this last move command, which then breaks things because we now also request the controller's alarm status when the motor stops and this won't work if the serial device is closed :upside_down_face:.

Another way of implementing this would be to just not use the `move()` method here and manually send `FP0` (or whatever it is) instead.

dc2917:
Ahh OK, I see! Then I think this makes sense. 

I guess just waiting for the stepper motor to finish moving before triggering the shutdown sequence wouldn't be practical?

alexdewar:
> I guess just waiting for the stepper motor to finish moving before triggering the shutdown sequence wouldn't be practical?

We could totally do this. I just don't think it's needed. The current sequence is:

1. Send stop moving command
2. Send command to move to nadir
3. Disconnect

The controller will happily keep on moving the motor after you've disconnected from it (you could even pull the cable out). We *could* have another step that waits for the move to finish and FROG would keep running in the background while this happened, but it's not needed unless we want to check whether the move succeeded or something... that would end up making the code more complicated though.

dc2917 (approved):

PR #830, opened 2025-04-24
Fix: Confirm whether to close measure script dialog when user presses escape

### LINKED ISSUE #799: Confirmation dialog not shown if users exit measure script dialog with esc
I've confirmed that this is a bug, at least on my Linux machine -- presumably it's the same on Windows but it would be good to double-check this (and to double-check that any fix definitely works on Windows).

The `ScriptRunDialog` class overrides the `closeEvent` method so that we can show users a confirmation dialog if they try to close it while the measure script is still running, to make sure they don't accidentally terminate their experiment prematurely. Bizarrely though, it seeems that you can close the dialog by pressing esc and if you do so then the `closeEvent` method is never invoked and so the measure script will be stopped without asking for user confirmation.

We should ask for confirmation if the user presses esc too. It might also be worth seeing if there are any other key combinations that will close the dialog without confirmation.

**ALSO:** I've noticed that "yes" is the default option in the confirmation dialog, so if the user accidentally presses enter at this point, it will stop the measure script. No/cancel should be the default.

alexdewar:
# Description

By default, pressing escape triggers the reject event, which closes the window unconditionally. We want to confirm whether to close the measure script dialog, so intercept the keypress in this case.

Fixes #799.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/frog/gui/measure_script/script_run_dialog.py
+++ b/frog/gui/measure_script/script_run_dialog.py
@@ -1,7 +1,8 @@
 """Provides a dialog to display the progress of a running measure script."""
 
 from pubsub import pub
-from PySide6.QtGui import QCloseEvent, QHideEvent
+from PySide6.QtCore import Qt
+from PySide6.QtGui import QCloseEvent, QHideEvent, QKeyEvent
 from PySide6.QtWidgets import (
     QDialog,
     QDialogButtonBox,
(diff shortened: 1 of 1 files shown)


alexdewar:
I've confirmed that this fix also works on Windows.

dc2917 (approved):

PR #831, opened 2025-04-24
ST10Controller: Unconditionally stop motor before move

### LINKED ISSUE #715: `ST10Controller`: Unconditionally stop motor before starting new move
The stepper motor controller has this interesting feature whereby every time you send a move command it enqueues it in a buffer, so you can tell it to move to point A then point B. We never use this feature, however: the frontend just sends a stop command immediately before it sends a move command. As with #704 though, this is really a case of letting the hardware implementation bleed into the frontend. We should do this in the backend instead and thereby simplify the interface.

The `DummyStepperMotor` class won't need changing.

alexdewar:
# Description

If you send two consecutive move commands to the ST10 controller it'll move to the first position before moving to the second. This might be a useful feature in some cases, but it's not something we need. Let's simplify the API for `ST10Controller` by sending a stop command before every move command so that we don't have to worry about doing it on the frontend side.

Closes #715.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/frog/gui/measure_script/script.py
+++ b/frog/gui/measure_script/script.py
@@ -244,9 +244,6 @@ class ScriptRunner(StateMachine):
         self.current_measurement_count: int
         """How many times a measurement has been recorded at the current angle."""
 
-        # Send stop command in case motor is moving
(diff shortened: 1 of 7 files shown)


dc2917 (approved):
Makes sense!

PR #856, opened 2025-08-01
Use `uv_build` build backend cf. `hatchling`

alexdewar:
It seems that there is a `uv` build backend that we could be using. I suspect it doesn't make much odds and we're not doing many builds anyway, but let's use `uv_build` as it's the recommended default.

I've tried it locally and it seems to work (very quickly!).


dc2917 (approved):
Makes sense!

PR #969, opened 2026-04-17
Bump version to v3.1.0

alexdewar:
# Description

Bump the version to v3.1.0 in preparation for the next release.

Now that we have a `CITATION.cff` file, there is another place where the program version needs to be updated. I tried using [`bump-my-version`](https://pypi.org/project/bump-my-version/) to update the version strings together, but the configuration was annoying so I gave up. In the end, I added a test to check that the `version` field in `CITATION.cff` matches the one in `pyproject.toml`, which is similar to what I've done for MUSE2.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,8 +1,8 @@
 cff-version: 1.2.0
 message: If you use this software, please cite it as below.
 title: FROG
-version: 3.0.0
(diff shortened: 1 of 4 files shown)


dc2917 (approved):
Fair enough!

